PR #2, opened 2025-05-20
feat: initial interface design proposal

rvagg:


rvagg:
Using this as a base commit but things have progressed quite a bit since this

PR #33, opened 2025-05-26
feat: implementation of initial design with mock backend

rvagg:
See https://github.com/FilOzone/synapse-sdk/blob/rvagg/mock-impl/example-usage.js for the client experience that generates the output below. Currently everything is entirely mocked, no chain or SP chatter, just delays and pretence.

```
$ npm install
$ npm run example 

> synapse-sdk@0.0.0 example
> npm run build && node example-usage.js


> synapse-sdk@0.0.0 build
> tsc

[Mock] Synapse initialized with options: { withCDN: true, rpcAPI: undefined, serviceContract: undefined }
Checking balance...
[Mock] Checking balance...
Current balance: 100 USDFC

Creating storage service...
[Mock] Creating storage service...
[Mock] Storage service created with proofSetId: ps_wc4uyv1kdyk, SP: f01234
Using proof set ID: ps_wc4uyv1kdyk
Using storage provider: f01234

Uploading data...
[Mock] Starting upload...
[Mock] Generating CommP for 24 bytes...
Generated CommP: baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha
[Mock] Storing data with provider...
Stored data with provider: f01234
[Mock] Committing to chain...
Blob committed on chain: 0x66f2ecd3b0d56
Data is being proven in proof set: ps_wc4uyv1kdyk

Downloading data...
[Mock] Downloading baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha
[Mock] Options: undefined
[Mock] Verifying data integrity...
Downloaded: "Hello, Filecoin Synapse!"
Download successful: true

Downloading directly from SP (no CDN)...
[Mock] Downloading baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha
[Mock] Options: { withCDN: false }
[Mock] Verifying data integrity...
Direct download successful: true

Downloading without verification...
[Mock] Downloading baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha
[Mock] Options: { noVerify: true }
Unverified download successful: true

Settling payments...
[Mock] Settling payments...
Settled payment rail for epoch 58275121
Settlement cost: 0.05 USDFC

Deleting data...
[Mock] Deleting baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha
Deleted blob with CommP baga6ea4seaqjtovkwk4myyzj56eztkh5pzsk5upksan6f5outesy62bsvl4dsha from proof set ps_wc4uyv1kdyk

Withdrawing funds...
[Mock] Withdrawing 10 USDFC...
[Mock] Withdrawal successful, new balance: 90
Withdrawn 10 USDFC, new balance: 90 USDFC
```

### DIFF
--- a/src/storage-service.ts
+++ b/src/storage-service.ts
@@ -1,0 +1,84 @@
+/**
+ * Mock implementation of StorageService
+ */
+
+import type { StorageService, CommP, DownloadOptions, SettlementResult, ProofSetId, StorageProvider } from './types.js'
+import { MockUploadTask } from './upload-task.js'
+import { normalizeCommP } from './commp.js'
+
+export class MockStorageService implements StorageService {
+  readonly proofSetId: ProofSetId
+  readonly storageProvider: StorageProvider
+  private _storedData: Map<string, Uint8Array> = new Map()
+
+  constructor(proofSetId: ProofSetId, storageProvider: StorageProvider) {
+    this.proofSetId = proofSetId
+    this.storageProvider = storageProvider
+  }
+
+  upload(data: Uint8Array | ArrayBuffer): MockUploadTask {
+    console.log('[Mock] Starting upload...')
+    const uploadTask = new MockUploadTask(data)
+
+    // Store data for later retrieval (in mock)
+    const storeData = async () => {
+      const commp = await uploadTask.commp()
+      const bytes = data instanceof ArrayBuffer ? new Uint8Array(data) : data
+      this._storedData.set(commp.toString(), bytes)
+    }
+    storeData().catch(console.error)
+
+    return uploadTask
+  }
+
+  async download(commp: CommP | string, options?: DownloadOptions): Promise<Uint8Array> {
+    const normalizedCommP = normalizeCommP(commp)
+    const commpString = normalizedCommP.toString()
+
+    console.log('[Mock] Downloading', commpString)
+    console.log('[Mock] Options:', options)
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 300))
+
+    // Check if we have the data stored
+    const data = this._storedData.get(commpString)
+    if (!data) {
+      throw new Error(`Data not found for CommP: ${commpString}`)
+    }
+
+    // Mock verification
+    if (!options?.noVerify) {
+      console.log('[Mock] Verifying data integrity...')
+      await new Promise(resolve => setTimeout(resolve, 200))
+    }
+
+    // Return a copy of the data
+    return new Uint8Array(data)
+  }
+
+  async delete(commp: CommP | string): Promise<void> {
+    const normalizedCommP = normalizeCommP(commp)
+    const commpString = normalizedCommP.toString()
+
+    console.log('[Mock] Deleting', commpString)
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 500))
+
+    // Remove from mock storage
+    this._storedData.delete(commpString)
+  }
+
+  async settlePayments(): Promise<SettlementResult> {
+    console.log('[Mock] Settling payments...')
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 1000))
+
+    return {
+      settledAmount: '0.05', // Mock amount in USDFC
+      epoch: Math.floor(Date.now() / 30000), // Mock epoch
+    }
+  }
+}
--- a/src/synapse.ts
+++ b/src/synapse.ts
@@ -1,0 +1,79 @@
+/**
+ * Mock implementation of the main Synapse class
+ */
+
+import type { Synapse as ISynapse, SynapseOptions, StorageOptions, TokenAmount } from './types.js'
+import { MockStorageService } from './storage-service.js'
+
+export class MockSynapse implements ISynapse {
+  private _options: SynapseOptions
+  private _balance: number = 100 // Mock starting balance
+
+  constructor(options: SynapseOptions) {
+    this._options = options
+    console.log('[Mock] Synapse initialized with options:', {
+      withCDN: options.withCDN,
+      rpcAPI: options.rpcAPI,
+      serviceContract: options.serviceContract,
+    })
+  }
+
+  async balance(): Promise<TokenAmount> {
+    console.log('[Mock] Checking balance...')
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 200))
+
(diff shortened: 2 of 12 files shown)


rjan90:
Hi @juliangruber. Given your expertise in this area, we were wondering if you could give this PR a review if you have availability? Your feedback would be invaluable to get the SDK work off the ground.

src/storage-service.ts
```diff
@@ -0,0 +1,84 @@
...
+    console.log('[Mock] Starting upload...')
+    const uploadTask = new MockUploadTask(data)
+
+    // Store data for later retrieval (in mock)
+    const storeData = async () => {
+      const commp = await uploadTask.commp()
+      const bytes = data instanceof ArrayBuffer ? new Uint8Array(data) : data
+      this._storedData.set(commp.toString(), bytes)
+    }
+    storeData().catch(console.error)
```
juliangruber, line 29: I would expect this function to fail if the `storeData` call fails. What is the motivation for decoupling the two calls?

juliangruber: In case `storeData` fails, can we make the `uploadTask` Promise reject (if it's a Promise, if it's not, find another way to emit an error)?

src/storage-service.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  async settlePayments(): Promise<SettlementResult> {
+    console.log('[Mock] Settling payments...')
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 1000))
+
+    return {
+      settledAmount: '0.05', // Mock amount in USDFC
```
juliangruber, line 80: Discussion point: Do we want to use strings, or BigInts? BigInts can be divided by e.g. 1e18, to implement precise decimals (like attoFIL and FIL).

src/synapse.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Mock implementation of the main Synapse class
+ */
+
+import type { Synapse as ISynapse, SynapseOptions, StorageOptions, TokenAmount } from './types.js'
```
juliangruber, line 5: I find it easier to reason about types if they're defined in the same file as the class that uses them, vs throwing all types into one central type file.

src/synapse.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.log('[Mock] Checking balance...')
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 200))
+
+    return this._balance
+  }
+
+  async deposit(amount: TokenAmount): Promise<TokenAmount> {
+    const depositAmount = typeof amount === 'string' ? parseFloat(amount) : Number(amount)
```
juliangruber, line 31: parsing floats will not be great for dealing with amounts, you'd want them to be precise. I suggest (as above) to either use BigInts, or at least strings

src/synapse.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.log(`[Mock] Depositing ${depositAmount} USDFC...`)
+
+    // Simulate network delay
+    await new Promise(resolve => setTimeout(resolve, 1000))
+
+    this._balance += depositAmount
```
juliangruber, line 37: are you concerned about the balance getting out of sync with network state? Would it be cleaner to always sync it and not cache it?

juliangruber: Downside: Read after write consistency might be sacrificed, and that's usually more important to keep up, for DX

src/upload-task.ts
juliangruber, line 1: This class will probably need a way to throw errors

src/storage-service.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Mock implementation of StorageService
+ */
+
+import type { StorageService, CommP, DownloadOptions, SettlementResult, ProofSetId, StorageProvider } from './types.js'
+import { MockUploadTask } from './upload-task.js'
+import { normalizeCommP } from './commp.js'
+
+export class MockStorageService implements StorageService {
```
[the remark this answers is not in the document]
rvagg: 🤷 those are open questions for now, I imagined this thing to be flexible enough to point it at different contracts .. also the naming seems to be in flux, "Pandora" only appeared on my radar 2 days ago and I don't even know what it's supposed to cover! We can rename this later, I don't expect people to instantiate this class directly, but have it instantiated for them by the main service.

src/storage-service.ts
```diff
@@ -0,0 +1,84 @@
...
+  readonly proofSetId: ProofSetId
```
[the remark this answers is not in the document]
rvagg: Yeah, this is fair, I imagine this "storage service" to eventually evolve into a flexible thing that can interface with PDP for you and maybe that's with multiple providers and multiple proof sets; but initially, the intention is that you make one of these objects per provider & proofset tuple. We'll figure it out by default for you but if you want to interact with other providers or make more proofsets then you'd make a new one of these and interact with that, so holding the proofset here makes sense for that .. for now at least.

src/storage-service.ts
rvagg: just throwing a quick mock in here for now :jedi-hand-wave: ignore the details I think, we'll fill it in soon

src/synapse.ts
rvagg: I think you're suggesting that we only ever deal in full integer token amounts, not fractional amounts? I think you're right, I don't like making the user divide everything by 10^18 but I guess that's common practice in this world.

src/synapse.ts
rvagg: no, `this._balance` doesn't exist in the current form of this code, it's silly to have it locally

src/upload-task.ts
rvagg: yeah, I was just deferring that until we had errors to throw; the intention is that if any of these steps errors then the remaining steps in the sequence should error, which will mean holding on to the promises for each of them and resolving them at each layer I think

src/storage-service.ts
rvagg: I'm just going with `number|bigint` inputs and `bigint` outputs now, but I'll add a `decimals(token)` that always returns `18` for now which will serve both USDFC and FIL but may be something else for another ERC20 we add support for so it may need to be dynamic and do an eth_call for you

PR #42, opened 2025-05-31
feat: signing operations for PDP

rvagg:
I think this is something approximating what's going on here: https://github.com/FilOzone/filecoin-services/pull/12 and here: https://www.notion.so/filecoindev/Synapse-Authentication-Spec-201dc41950c1804488b5cd3515fa2f4f

Parking it here as a draft, I don't need it yet.


rvagg:
cc @ZenGround0

rvagg:
@ZenGround0:

I've fleshed this out a bit more and added helpers to sign the various different forms of auth blob. I've also put an example Node.js program in that will run it and it will push through some sample values and show you a hex blob form of it.

```sh
npm install
npm run build
node example-auth-simple.js
```

Edit example-auth-simple.js to change the inputs to each of these.

I've not spent a ton of time on this, so there may be some things wrong (or janky), but 🤞.

There also seems to be some discrepancies between the current contract @ https://github.com/FilOzone/filecoin-services/blob/5511d2eb63159263027e510b957fa1adba5d5017/service_contracts/src/SimplePDPServiceWithPayments.sol and the description @ https://www.notion.so/filecoindev/Synapse-Authentication-Spec-201dc41950c1804488b5cd3515fa2f4f mainly around this `clientDataSetId` vs `proofSetId` in the 3 non-create calls.

example-auth-simple.js
```diff
@@ -0,0 +1,183 @@
...
+    console.log(`   Client Dataset ID: ${ADD_ROOTS_PARAMS.clientDataSetId}`)
+    console.log(`   First Root ID: ${ADD_ROOTS_PARAMS.firstRootId}`)
+    console.log(`   Roots to Add: ${ADD_ROOTS_PARAMS.rootDataArray.length}`)
+    ADD_ROOTS_PARAMS.rootDataArray.forEach((root, i) => {
+      console.log(`     [${i}] CID: ${root.cid.toString().substring(0, 20)}... (size: ${root.rawSize})`)
+    })
+
+    const addRootsSig = await synapse.signAddRoots(
+      ADD_ROOTS_PARAMS.clientDataSetId,
+      ADD_ROOTS_PARAMS.firstRootId,
```
ZenGround0, line 121: The biggest thing we have to sync on for threading all the data together is that firstRootId is going to require going to PDPVerifier (proxy) getter

PR #53, opened 2025-06-03
feat: basic PDP service contract interactions

rvagg:
**Draft** for now, parking it here because I started tinkering with this but haven't made much progress. https://github.com/FilOzone/filecoin-services/pull/18 was a result of finding it was tricky to fetch the data with a single call.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -62,3 +62,80 @@ export const RPC_URLS = {
     websocket: 'wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v1'
   }
 } as const
+
+/**
+ * PDP service contract addresses (proxy addresses for upgradeable contracts)
+ */
+export const PDP_SERVICE_CONTRACT_ADDRESSES = {
+  mainnet: '', // TODO: Get actual mainnet address from deployment
+  calibration: '0x2B76E983d30553E7717547230670D4F4F4d813aC'
+} as const
+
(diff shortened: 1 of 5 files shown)


src/constants.ts
```diff
@@ -62,3 +62,80 @@ export const RPC_URLS = {
...
+export const PDP_SERVICE_CONTRACT_ADDRESSES = {
+  mainnet: '', // TODO: Get actual mainnet address from deployment
+  calibration: '0x2B76E983d30553E7717547230670D4F4F4d813aC'
+} as const
+
+/**
+ * PDP Verifier contract addresses
+ */
+export const PDP_VERIFIER_ADDRESSES = {
+  mainnet: '', // TODO: Get actual mainnet address from deployment
```
ZenGround0, line 78: `0x9C65E8E57C98cCc040A3d825556832EA1e9f4Df6` 
You can find this value in the filozone/pdp README.md

src/constants.ts
```diff
@@ -62,3 +62,80 @@ export const RPC_URLS = {
...
+  calibration: '0x5A23b7df87f59A291C26A2A1d684AD03Ce9B68DC'
+} as const
+
+/**
```
ZenGround0, line 82: you're also going to want the PDPVerifier ABI to get Next RootID for constructing signer data for adding roots

PR #54, opened 2025-06-03
Further integrate `withCDN=true`

juliangruber:
This integrates the CDN flag into all of the mock calls where I think the real implementation would also need it. Also uses the CDN for retrieval.

### DIFF
--- a/src/storage-service.ts
+++ b/src/storage-service.ts
@@ -13,16 +13,20 @@ export class MockStorageService implements StorageService {
   readonly proofSetId: ProofSetId
   readonly storageProvider: StorageProvider
   private readonly _storedData: Map<string, Uint8Array> = new Map()
+  private readonly _signerAddress: string
+  private readonly _withCDN: boolean
 
-  constructor (proofSetId: ProofSetId, storageProvider: StorageProvider) {
+  constructor (proofSetId: ProofSetId, storageProvider: StorageProvider, signerAddress: string, withCDN: boolean) {
     this.proofSetId = proofSetId
     this.storageProvider = storageProvider
+    this._signerAddress = signerAddress
+    this._withCDN = withCDN
   }
 
   upload (data: Uint8Array | ArrayBuffer): MockUploadTask {
     console.log('[MockSynapse] StorageService.upload() called')
     console.log('[MockSynapse] Data size:', data instanceof ArrayBuffer ? data.byteLength : data.length, 'bytes')
-    const uploadTask = new MockUploadTask(data)
+    const uploadTask = new MockUploadTask(data, this._withCDN)
 
     // Store data for later retrieval (in mock)
     const storeData = async (): Promise<void> => {
@@ -46,6 +50,12 @@ export class MockStorageService implements StorageService {
     console.log('[MockSynapse] StorageService.download() called')
     console.log('[MockSynapse] CommP:', commpString)
     console.log('[MockSynapse] Download options:', options)
+
+    if (options?.withCDN !== false && (this._withCDN || options?.withCDN)) {
+      console.log('[MockSynapse] Using CDN for download (withCDN=true)')
+      const res = await fetch(`https://${this._signerAddress}.calibration.filcdn.io/${commpString}`)
+      return new Uint8Array(await res.arrayBuffer())
+    }
 
     // Simulate network delay
     console.log('[MockSynapse] Simulating download network delay (300ms)...')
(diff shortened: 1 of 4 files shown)


src/storage-service.ts
```diff
@@ -47,6 +51,12 @@ export class MockStorageService implements StorageService {
     console.log('[MockSynapse] CommP:', commpString)
     console.log('[MockSynapse] Download options:', options)
 
+    if (options?.withCDN !== false && (this._withCDN || options?.withCDN)) {
```
juliangruber, line 54: It's a bit awkward to both support this option in the synapse constructor and the upload arguments. This makes it so if either is true, the CDN is used, unless a `false` was passed to this call.

juliangruber: Tests need to be added for this condition. I want to get approval on the direction first before adding more code

src/synapse.ts
```diff
@@ -396,7 +399,7 @@ export class Synapse implements ISynapse {
     )
     console.log('[MockSynapse] Storage service ready for operations')
 
-    return new MockStorageService(proofSetId, storageProvider)
+    return new MockStorageService(proofSetId, storageProvider, await this._signer.getAddress(), this._withCDN)
```
bajtos, line 402: I believe the storage service will eventually (for M1) need the full signer, so that it can sign the transaction creating the storage deal and sign the extraData parameter passed to Curio PDP API when working with ProofSets and Roots.

How about modifying the signature of the service constructor to accept the signer instead of the address?

```suggestion
    return new MockStorageService(proofSetId, storageProvider, this._signer, this._withCDN)
```

This is just a suggestion to consider. My understanding may be wrong.

src/storage-service.ts
```diff
@@ -13,16 +13,20 @@ export class MockStorageService implements StorageService {
   readonly proofSetId: ProofSetId
   readonly storageProvider: StorageProvider
   private readonly _storedData: Map<string, Uint8Array> = new Map()
+  private readonly _signerAddress: string
+  private readonly _withCDN: boolean
 
-  constructor (proofSetId: ProofSetId, storageProvider: StorageProvider) {
+  constructor (proofSetId: ProofSetId, storageProvider: StorageProvider, signerAddress: string, withCDN: boolean) {
```
bajtos, line 19: Boolean function arguments are difficult to understand when reading the code.

```js
new MockStorageService(proofSetId, storageProvider, address, true)
//                 It's not clear what does "true" mean here? ^^^
```

I propose using an "options" object.

```suggestion
  constructor (proofSetId: ProofSetId, storageProvider: StorageProvider, signerAddress: string, options: { withCDN?: boolean } = {}) {
```

src/synapse.ts
bajtos: Also, shouldn't the signer or signer address be provided in the storage constructor?

src/storage-service.ts
juliangruber: https://github.com/FilOzone/synapse-sdk/blob/0a42daac1205d13ca22a86f7e5d1a9fba1618f2e/src/synapse.ts#L137-L142

rvagg: This is a private constructor so I like being explicit here and let the factory method do the validating. With enough arguments this is going to get unwieldily, but it's still private so that's not a user concern.

But I'm easy on this tbh, I like the pattern of validating flexible args in a factory and be explicit in a constructor, and the factory only exists at this point to deal with async awkwardness. As long as properties get set and we have some consistency.

src/storage-service.ts
rvagg: We don't have to have the option in both places, we could just do it one or the other? It could even show upas an option to `upload()` and `download()`, and that might end up being too many places and too confusing. I'm fine with keeping the number of these down. What do you think?

src/synapse.ts
rvagg: yeah, I think you're right, signer all the way down so we can get that auth blob signed.

rvagg (approved):
I'm fine however you want to play this for now, we can merge when you're happy with it and refine it as we get to use it and figure out what's ergonomic and what's not

src/storage-service.ts
juliangruber: 🤷‍♂️ I think from the user's perspective, what you have intuitively make sense: You can define options on the instance, but overwrite on demand in the specialized function.

PR #61, opened 2025-06-05
chore(docs,ai): compress and organise LLM file

rvagg:
The file was getting a bit out of hand, had some inaccuracies in it, a lot of duplication and disorganisaion. I've done a bit of manual pruning here.

Sorry @rjan90, this probably got a lot less useful for you.


rvagg:
Just noting that this is a little experimental. I'm finding that the context blows up too quickly, but I'm concerned about dropping important context so we'll see how it goes.

PR #68, opened 2025-06-05
fix: pass the right cid bytes and size for auth for addroots

aarshkshah1992:
Uses https://github.com/FilOzone/synapse-sdk/pull/62/files

### DIFF
--- a/src/test/pdp-auth.test.ts
+++ b/src/test/pdp-auth.test.ts
@@ -41,15 +41,15 @@ const FIXTURES = {
       withCDN: true
     },
     addRoots: {
-      signature: '0x4d873bfc7e03792bf262434673d3f58cebaa5ecd1d81ff4d1974e3ef382e7ac40192c2973f7ec7895cfb0963b330dfa236c4c15093d3c136e321587d77c926f31b',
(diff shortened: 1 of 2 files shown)


src/test/pdp-auth.test.ts
```diff
@@ -41,15 +41,15 @@ const FIXTURES = {
...
+      signature: '0x0d4df872fd94796a2883f5c307d4d8402ded8ad47b9800b91a781735461a629562962255f9e14927cab71c60224fdf3f79f24799efac0aa4692d06362af5f87a1c',
+      digest: '0xf17b431d935267844219010cf340decd8d7e631fdc7f7a6d1d4b6a7f1c1c6430',
       clientDataSetId: 12345,
       firstAdded: 1,
       rootDigests: [
         '0xfc7e928296e516faade986b28f92d44a4f24b935485223376a799027bc18f833',
         '0xa9eb89e9825d609ab500be99bf0770bd4e01eeaba92b8dad23c08f1f59bfe10f'
       ],
-      rootSizes: [1024, 2048]
+      rootSizes: [2032, 4064]
```
rvagg, line 52: note these are not used in here, we have input values of 1024 and 2048 and these are the unpadded padded versions of those but they are included here to match the fixtures in https://github.com/FilOzone/filecoin-services/pull/36 although not strictly necessary

PR #70, opened 2025-06-05
feat(pdptool): AddRoots API call

rvagg:
Draft because I'm not 100% confident that we're passing the right types of CID around in here to where they need to be.


rvagg:
A win!

This is an actual piece on the curio server, and this is the actual raw byte size supplied here. Unfortunately you still need to figure out the inputs. But the curio database helps a little bit in figuring this out. We need to be fetching these things from the chain of course to make it easier.

Screenshot 2025-06-06 at 3 59 50 am

rvagg:
And I put pdp-tool-test.html into utils in here. This should make the branch on https://github.com/FilOzone/synapse-sdk/pull/58 redundant now.

PR #79, opened 2025-06-07
feat(utils): add Pandora storage provider admin portal

rvagg:
Add a single-page web application for administering the Pandora storage provider contract on Filecoin Calibration network.

Features:
- Public view of all approved storage providers
- Storage provider registration and status dashboard
- Contract owner management functions (approve/reject/remove providers)
- Support for custom Pandora contract addresses via clickable badge
- Real-time transaction monitoring and UI updates
- MetaMask and private key authentication options
- Dark theme UI inspired by plgenesis.com styling
- Responsive design for mobile and desktop

The portal serves as both a functional admin tool and example code demonstrating Synapse SDK integration with PandoraService methods.

---

I pretty much did none of this, the pieces are all in place in Synapse already, I just told Claude what I wanted, was a bit specific about it, and he did this while I was occupied with other things.

---

Screenshot 2025-06-07 at 11 44 49 pm

Screenshot 2025-06-07 at 11 44 58 pm

Screenshot 2025-06-07 at 11 45 29 pm

### DIFF
--- a/ADMIN_SAFE_INTEGRATION_PLAN.md
+++ b/ADMIN_SAFE_INTEGRATION_PLAN.md
@@ -1,0 +1,341 @@
+# Pandora Admin Portal - Safe Multisig Integration Plan
+
+## Overview
+
+This document outlines the plan to add Safe (formerly Gnosis Safe) multisig wallet support to the Pandora Admin Portal. The integration will enable contract owners using Safe multisig wallets to manage storage providers on the Filecoin network through a secure, decentralized interface.
+
+## Background
+
+### Current State
+The Pandora Admin Portal currently supports:
+- MetaMask wallet connection (browser extension)
+- Private key authentication (manual input)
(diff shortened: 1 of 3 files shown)


ZenGround0 (approved):
Looks good from a quick skim

rvagg:
fwiw I worked with Claude to propose a path to implementing Safe wallet into this. It's a little bit involved, I don't fully understand it, but I got it to come up with a plan and documentation about how. That would be helpful for admin at least. I can share that if needed.

rvagg:
Actually, I've just checked it in, it's a lot of information and could be helpful. @ZenGround0 maybe you can consider whether it's worth it. The main issue is that it needs to be hosted on its own domain, but that should even work on gh-pages I think 🤷 .

rvagg:
Going to land this, it's useful as a reference and we can ditch later if we don't care for it

Issue #88, opened 2025-06-10
Broken proof set & SP selection algorithm


rvagg:
I thought I'd implemented this properly but it turns out it was a mirage and I found out when I went back to upload another piece from a fresh SDK with Pandora having >1 providers in it and I was allocated a fresh provider.

When calling `createStorage()` without specifying a provider, the SDK randomly selects a storage provider first, then only checks that specific provider for existing proof sets. It _should_ automatically figure out who you've talked to before and prefer that.

Current behaviour:
1. Randomly select provider from all approved providers
2. Check only that provider for existing proof sets
3. Create new proof set if none found

Expected behaviour:
1. Check ALL existing proof sets across ALL providers
2. Prefer providers with existing proof sets (especially those with roots)
3. Only randomly select if no suitable existing proof sets found

The goal here is that repeat uses of the SDK will land you back at the same proof set, with the same provider, unless you specify otherwise.

While I'm in here, I'm going to add the ability to select a `proofSetId` as well as provider, I'll also consider allowing `providerAddress` because `providerId` is an int that's on chain and it's a bit awkward (as I discovered with use).

PR #92, opened 2025-06-10
feat(storage): implement smart provider selection to prefer existing relationships

rvagg:
- Add proofSetId, providerAddress parameters for explicit selection
- Implement progressive provider resolution with lazy loading
- Smart selection now prioritises existing proof sets over random provider selection
- Add parameter validation and conflict detection for better UX
- Reduce RPC calls through targeted data fetching with Promise.all
- Ensure repeat SDK usage returns to same provider/proof set when possible

Fixes: https://github.com/FilOzone/synapse-sdk/issues/88

### DIFF
--- a/src/pandora/service.ts
+++ b/src/pandora/service.ts
@@ -249,15 +249,6 @@ export class PandoraService {
   }
 
   /**
-   * Get only the proof sets managed by this Pandora contract
-   * @param clientAddress - The client's wallet address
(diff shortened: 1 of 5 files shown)


rvagg:
Seems to work well according to my manual integration testing, which is mostly about selecting the same provider I've been dealing with. That's the main thing we're solving for here!

Issue #82, opened 2025-06-09
Minimum size: what is it, can we reduce it?


rvagg:
CommP should have a minimum of 65 bytes (I'm fairly sure). But the minimum seems to be much higher, I haven't precisely measured it yet but the highest failure I have is at 1190 bytes but lowest success is 4760. So somewhere in there.

The error is this:

```
500 Internal Server Error - Failed to send transaction: failed to estimate gas: message execution failed (exit=[33], revert reason=[message failed with backtrace:
  00: f0158257 (method 3844450837) -- contract reverted at 108 (33)
  01: f0158257 (method 6) -- contract reverted at 14013 (33)
   (RetCode=33)], vm error=[0xc7b67cf300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000001d53697a65206d7573742062652061206d756c7469706c65206f66203332000000])
```

Which is this: https://github.com/FilOzone/pdp/blob/92b50453c27976fd46c31ecf0b57fffd74f43c1c/src/PDPVerifier.sol#L361-L363

My guess is that this may be something to do with Curio and the way it's doing its awkward raw->padded->unpadded dance. Maybe it's not reporting it correctly and the contract isn't getting its the version of the size it wants for smaller raw sizes.

Needs investigation:

1. We need to put the actual minimum into the SDK so users aren't confused by this.
2. We need to fix the minimum and make it much smaller if possible.

ZenGround0:
Quick checking this looks like a curio issue but rather than a min size thing it appears to be an alignment thing:

Investigating our nice error string from the contract gas estimation side
```
echo "0xc7b67cf300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000001d53697a65206d7573742062652061206d756c7469706c65206f66203332000000" | xxd -r -p | strings
Ƕ|?
Size must be a multiple of 32
```
This is the sort of thing curio should be taking care of on its own.  I also recall this showing up in pdp internal hackathon and never got around to investigating.  I am starting to be suspicious that this is related to the curio unpadded bug we found during the addRoots signing debugging.  I've got some theories to validate.

ZenGround0:
FYI this is useful for debugging: https://github.com/filecoin-project/curio/pull/528

rvagg:
[...]

PR #95, opened 2025-06-11
fix(addroots): report full padded piece size to contract to match Curio

rvagg:
Fixes: https://github.com/FilOzone/synapse-sdk/issues/82
Ref: https://github.com/filecoin-project/curio/pull/529

We can now support down to 65 bytes in size, but this change, and the corresponding change in Curio will mean breakage for everyone unless they have matching SDK & PDP server working together. So this will be a mandatory upgrade to speak PDP once we roll out the Curio update (ASAP).

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ const storage = await synapse.createStorage()
 
 // Upload data
 const uploadResult = await storage.upload(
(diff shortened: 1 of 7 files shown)


rvagg:
Matching contract fixtures here: https://github.com/FilOzone/filecoin-services/pull/48

Issue #85, opened 2025-06-10
Decouple piece downloads from storage provider instances


rvagg:
## Problem

`StorageService#download(commp)` is bound to the SP instance that the `StorageService` was created with. This fails when pieces are stored across multiple SPs.

**Current:** Download only works with the SP you're connected to
**Reality:** Pieces may exist on different SPs than the one you're using

## Example

```typescript
// Upload to SP-A
const storage1 = await synapse.createStorage({ providerId: 1 })
const { commp } = await storage1.upload(data)

// Later, connect to SP-B
const storage2 = await synapse.createStorage({ providerId: 2 })
await storage2.download(commp)  // ❌ Fails - wrong SP
```

## When This Breaks

1. Proof sets created outside SDK with different SPs
2. Switching SPs (avoiding problematic providers)
3. Same piece stored on multiple SPs
4. Different SP selected when `withCDN` changes

## Proposed Solution

Given that `SynapseStorage` is tied to a specific SP, we might need to decouple `download` and pull it up to the `Synapse` instance level so it's properly agnostic of the SP.

Add piece discovery to find and download from any SP:

```typescript
// Find which SPs have a piece
const providers = await synapse.findPiece(commp)

await synapse.download(commp)                    // Auto-select SP with findPiece internally
await synapse.download(commp, { providerId: 1 }) // Specific SP
```

## Technical Approach / Open Questions

**Piece Discovery**: On-chain vs off-chain? Can we get this information from `PandoraService` and/or `PDPVerifier`? Or do we need a subgraph call?

*Note: CDN support will be handled separately by the Spark team but the `withCDN` bool from the proofset should be available after lookup*

**Dynamic SP Connection**: We can use provider registry from `PandoraService` to get retrieval URLs.

rvagg:
@juliangruber I think it would be good for you to be aware of this as you consider CDN integration. Not sure when this will happen but it should happen ~soon because we'll be dealing with error reports from users with novel configurations.

rvagg:
Yeah, that's a good point, being able to be specific about where to download from. I like that. My only concern is usability and making it clear what's going on. `synapse.createStorage().download()` is very close to `synapse.download()`, maybe it needs a different name, or some kind of modifier. `synapse.createStorage().directDownload()` perhaps?

rvagg:
Here's my proposal.

### Core Design: Pluggable Discovery

A pluggable discovery interface that separates "finding pieces" from "downloading pieces":

```typescript
[...]

PR #97, opened 2025-06-11
feat(download): composable download location discovery mechanism (WIP)

rvagg:
Closes: https://github.com/FilOzone/synapse-sdk/issues/85

Very rough first go at this. I need to go back over this and rethink a few things. I'm considering using streams as the return type from this new API (or maybe we add `downloadStreaming()`?). Also a few things are wonky in here and smell a little bit. But the `PieceDiscovery` is the important bit here. I started off with `providerId` but I think address is more universally helpful. I just need to trace back through the code to work out how expensive that is for the common case.

### DIFF
--- a/src/synapse.ts
+++ b/src/synapse.ts
@@ -6,11 +6,13 @@ import { ethers } from 'ethers'
 import {
   type SynapseOptions,
   type StorageServiceOptions,
-  type FilecoinNetworkType
+  type FilecoinNetworkType,
+  type PieceDiscovery
 } from './types.js'
 import { StorageService } from './storage/index.js'
 import { PaymentsService } from './payments/index.js'
 import { PandoraService } from './pandora/index.js'
+import { ChainDiscovery } from './discovery/chain.js'
 import { CHAIN_IDS, CONTRACT_ADDRESSES, createError } from './utils/index.js'
 
 export class Synapse {
@@ -20,13 +22,14 @@ export class Synapse {
   private readonly _payments: PaymentsService
   private readonly _provider: ethers.Provider
   private readonly _pandoraAddress: string
+  private readonly _pieceDiscovery: PieceDiscovery
 
   /**
    * Create a new Synapse instance with async initialization.
    * @param options - Configuration options for Synapse
    * @returns A fully initialized Synapse instance
    */
-  static async create (options: SynapseOptions): Promise<Synapse> {
+  static async create (options: SynapseOptions & { pieceDiscovery?: PieceDiscovery }): Promise<Synapse> {
     // Validate options
     const providedOptions = [options.privateKey, options.provider, options.signer].filter(Boolean).length
     if (providedOptions !== 1) {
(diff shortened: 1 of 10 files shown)


src/synapse.ts
```diff
@@ -120,13 +123,23 @@ export class Synapse {
...
     }
 
+    // Determine Pandora address early for discovery initialization
+    const pandoraAddress = options.pandoraAddress ?? CONTRACT_ADDRESSES.PANDORA_SERVICE[network]
+    if (pandoraAddress === '' || pandoraAddress === undefined) {
+      throw new Error(`No Pandora service address configured for network: ${network}`)
+    }
+
+    // Create default discovery if not provided
+    const pieceDiscovery = options.pieceDiscovery ?? new ChainDiscovery(provider, pandoraAddress)
```
bajtos, line 133: @rvagg is this the place where to plug in the CDN-specific downloads?

```suggestion
    // Create default discovery if not provided
    const pieceDiscovery = options.pieceDiscovery ?? 
      (options.withCDN ? new FilCdnDiscovery(network) : new ChainDiscovery(provider, pandoraAddress))
```

src/synapse.ts
```diff
@@ -120,13 +123,23 @@ export class Synapse {
...
+
     return new Synapse(
       provider,
       signer,
       network,
       options.disableNonceManager === true,
       options.withCDN === true,
-      options.pandoraAddress
+      pandoraAddress,
+      pieceDiscovery
```
bajtos, line 142: I feel `withCDN` and `pieceDiscovery` are connected together. By exposing them as two independent parameters, we are creating edge cases where the combination of may not make sense & can be confusing - e.g. `withCDN = true` and `pieceDiscovery = new ChainDiscovery(/*...*/)`.

Also: post M1, I think we should support more than one CDN provider. It would be great if each CDN provider can provide a plugin for Synapse SDK that will provide all CDN-specific bits in one place.

Example usage as I am envisioning it:

```js
import { Synapse } from 'synapse'
import { FilCdnPlugin } from '@filcdn/synapse'

const synapse = Synapse.create({
  // ...
  withCDN: new FilCdnPlugin(network),
  // ...
})
```

I am flagging this as an aspect to think about more.

I think the proposal in this pull request is good enough for M1.

src/synapse.ts
```diff
@@ -231,6 +241,58 @@ export class Synapse {
...
+  ): Promise<ReadableStream<Uint8Array>> {
+    const client = await this._signer.getAddress()
+    const errors: Error[] = []
+    let attemptedUrls = 0
+
+    // Try URLs as they're discovered
+    for await (const url of this._pieceDiscovery.findPiece(commp, client, options)) {
+      attemptedUrls++
+      try {
+        const response = await fetch(url)
```
bajtos, line 264: Can we modify the discovery API so that `findPiece` returns either `url` or a `Request` object, i.e. the type of the first parameter of `fetch()` API ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#resource)).

This way, the discovery mechanism can provide additional metadata, e.g. request headers.

Think of a CDN that implements authorisation and requires clients to provide `Authorization: Bearer some-token`.

bajtos (approved):
Looks great to me!

What are the next steps to add support for retrievals using our CDN? (How can we contribute the changes? Should I open a PR targetting your feature branch, or wait until this PR is landed?)

src/synapse.ts
rvagg: yeah, this is a great idea, let's do that

bajtos:
_Cross-posting from a chat we had in Slack:_

We want to allow fallback in case the CDN retrieval fails, e.g. because the SDK user misconfigured `withCDN` flag and attempts to retrieve a file that's not linked to a proofset paying for CDN.

We agreed to implement this as "an optimistic GET and then fallback". 

We discussed other options, e.g. making a HEAD request first to check if the CDN retrieval is allowed, but we rejected it due to extra latency on the happy path.

PR #99, opened 2025-06-13
feat(retriever): implement PieceRetriever pattern for flexible piece downloads

rvagg:
- Add PieceRetriever interface and ChainRetriever/FilCdnRetriever implementations
- Add Synapse.download() method with provider selection and CDN support
- Add StorageService.providerDownload() to replace deprecated download()
- Extract piece URL construction utilities to utils/piece.ts
- Implement efficient parallel fetching with Promise.race() and proper abort handling
- Add signal propagation support for cancellation across retriever chain
- Update documentation to focus on storage operations over payment setup
- Add comprehensive test coverage for retriever components


rvagg:
Good advice from Copilot, adding value there. This is starting to feel very familiar, like building a miniature Lassie.

Pushed

```
fix(retriever): use Promise.any instead of Promise.race for provider selection

- Replace Promise.race with Promise.any to wait for first successful response
- Add ProviderAttemptResult interface for improved type safety
- Handle AggregateError when all providers fail
- Add test to verify fast-failing providers don't prevent success from slower ones

This ensures piece retrieval succeeds even when some providers fail quickly,
improving reliability of the download system.
```

PR #106, opened 2025-06-17
feat(cdn): add complete FilCDN retriever implementation

juliangruber:
Closes https://github.com/filcdn/roadmap/issues/5

### DIFF
--- a/src/retriever/filcdn.ts
+++ b/src/retriever/filcdn.ts
@@ -26,14 +26,12 @@ export class FilCdnRetriever implements PieceRetriever {
       providerAddress?: string
       withCDN?: boolean
       signal?: AbortSignal
-      fetch?: typeof fetch
     }
   ): Promise<Response> {
     if (options?.withCDN === true) {
       const cdnUrl = `https://${client}.${this.hostname()}/${commp.toString()}`
-      const fetchFn = options.fetch ?? globalThis.fetch
       try {
-        const cdnResponse = await fetchFn(cdnUrl, { signal: options?.signal })
+        const cdnResponse = await fetch(cdnUrl, { signal: options?.signal })
         if (cdnResponse.ok) {
           return cdnResponse
         } else if (cdnResponse.status === 402) {
--- a/src/test/retriever-filcdn.test.ts
+++ b/src/test/retriever-filcdn.test.ts
@@ -22,22 +22,27 @@ describe('FilCdnRetriever', () => {
           return baseResponse
         }
       }
-      const mockFetch: typeof fetch = async () => {
+
+      const originalFetch = global.fetch
+      global.fetch = async () => {
         throw new Error('Should not call fetch when withCDN is false')
       }
 
-      const cdnRetriever = new FilCdnRetriever(mockBaseRetriever, 'calibration')
-      const response = await cdnRetriever.fetchPiece(
-        mockCommP,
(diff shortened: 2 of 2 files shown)


src/retriever/filcdn.ts
```diff
@@ -22,32 +20,27 @@ export class FilCdnRetriever implements PieceRetriever {
...
-    //     if (cdnResponse.status === 402) {
-    //       // Payment required - fall back to chain retrieval
-    //       console.log('CDN requires payment, falling back to direct retrieval')
-    //     }
-    //   } catch (error) {
-    //     console.warn('CDN fetch failed:', error)
-    //   }
-    // }
+    if (options?.withCDN === true) {
+      const cdnUrl = `https://${client}.${this.network}.filcdn.io/${commp.toString()}`
```
bajtos, line 27: FWIW, the mainnet URL will be `https://${client}.filcdn.io/${commp.toString()}`.

It's not a big deal at this stage.

src/retriever/filcdn.ts
```diff
@@ -22,32 +20,27 @@ export class FilCdnRetriever implements PieceRetriever {
...
+      const fetchFn = options.fetch ?? globalThis.fetch
+      try {
+        const cdnResponse = await fetchFn(cdnUrl, { signal: options?.signal })
+        if (cdnResponse.ok) {
+          return cdnResponse
+        } else if (cdnResponse.status === 402) {
+          console.log('CDN requires payment')
```
bajtos, line 34: Let's use a more descriptive error message that tells the SDK user how to fix the problem.

```suggestion
          console.log('CDN requires payment. Please initialise Synapse SDK with the option `withCDN: true` and re-upload your files.')
```

rvagg: also `console.warn` I think?

rvagg:
@juliangruber @bajtos I'm fine if you both want to work on this, agree and merge; just please make sure your commits (or your final squashed commit) adhere to conventional commits style so the auto-release process properly versions the release. Probably should be something like: `feat(cdn): add complete FilCDN retriever implementation`. That'll give us a semver-minor bump. Just please avoid `!` in it so we don't get a semver-major.

src/retriever/filcdn.ts
```diff
@@ -22,32 +20,27 @@ export class FilCdnRetriever implements PieceRetriever {
...
-    //       // Payment required - fall back to chain retrieval
-    //       console.log('CDN requires payment, falling back to direct retrieval')
-    //     }
-    //   } catch (error) {
-    //     console.warn('CDN fetch failed:', error)
-    //   }
-    // }
+    if (options?.withCDN === true) {
+      const cdnUrl = `https://${client}.${this.network}.filcdn.io/${commp.toString()}`
+      const fetchFn = options.fetch ?? globalThis.fetch
```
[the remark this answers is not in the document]
bajtos: FWIW, if we are okay with modifying globals in our tests, we can also use Undici's built-in mocking feature (see https://nodejs.org/en/learn/test-runner/mocking#apis) or Node.js test runner's mocks (see https://nodejs.org/docs/latest/api/test.html#mocking).

src/retriever/filcdn.ts
juliangruber: a87f14d

PR #108, opened 2025-06-19
feat(payments): EIP-2612 support for USDFC->Payments (DRAFT)

rvagg:
This _should_ implement `depositWithPermit()` to slightly simplify (and speed up) the allowance + deposit flow for USDFC. Unfortunately I just can't make this work. I've tried deploying the latest `main` version of `Payments.sol` and depositing into that but I just can't get the USDFC DebtToken to accept it, it always fails with a signature validation and I can't work out what the delta is.

Here's the UX for the initial permit signing though with metamask, I just can't get this to land:

Screenshot 2025-06-19 at 5 04 13 pm

### DIFF
--- a/src/payments/service.ts
+++ b/src/payments/service.ts
@@ -26,6 +26,7 @@ export class PaymentsService {
   private readonly _signer: ethers.Signer
   private readonly _network: FilecoinNetworkType
   private readonly _disableNonceManager: boolean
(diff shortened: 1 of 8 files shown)


rvagg:
Works when I deploy payments contract with https://github.com/FilOzone/filecoin-services-payments/pull/142; but I'm not sure why

Issue #86, opened 2025-06-10
Metadata Field Design Discussion: Structured vs Unstructured Data


rjan90:
## Summary

Discussion about the design choice for the metadata field in the extraData structure - whether it should remain unstructured (current implementation) or become structured to enable better tooling support.

## Background

From PR discussion in [FilOzone/filecoin-services#19](https://github.com/FilOzone/filecoin-services/pull/19/files#r2125151537) and [this Slack Thread](https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1749039857407669?thread_ts=1748957215.157949&cid=C07CGTXHHT4) in the #fil-services channel:

- The metadata field was added during HotVault development for users to add string labels/identifiers for payment rails
- Currently designed as unstructured data for "opaque identifiers" (similar to a "note" field in payments)
- Example use case: Adding "deal-for-cat-picture" to be displayed in Explorer UI
- Never fully utilized in current implementation

### Current Design (Unstructured)
- **Pros**: Flexible, no need for upfront standardization, simple implementation
- **Use case**: User-defined labels/notes for deals/proofsets/roots that appear in Explorer
- **Implementation**: Part of extraData, emitted in events for indexer consumption

### Proposed Concern (Need for Structure)
- **Issue**: Unstructured metadata creates friction for tooling development
- **Problem**: Difficult for tools like explorers to parse and display service information consistently
- **Example**: Product descriptions, CDN pricing info not standardized
- **Risk**: Fragmented ecosystem where each service uses different metadata formats

rvagg:
Also not part of the content that's being addressed. The nice thing about the metadata _within_ the data is that it's part of the addressing and consistent across storage instances. If I replicate my piece 5 times but give it 5 different names, what does that mean? What if the metadata is a filename, can the same hashed piece of data be foo.mp4, bar.pdf, notes.txt, moo.wav and exploit.sh? Maybe this is ultimately a problem for the user, but my take would be that if you want metadata related to your content and it's to do with files then let's just support UnixFS packing, or some form of it, that bundles in a metadata block the UnixFS way that's also ~compatible with Kubo etc.

rvagg:
[...]

PR #111, opened 2025-06-21
feat: add SDK methods to fetch proofset roots and metadata

### LINKED ISSUE #102: 🔍 Add ProofSet Roots Retrieval
## Feature Request

Provide SDK methods to fetch proofset roots and their metadata, so developers no longer need to call the Curio storage API directly.

## Current Problem

There’s no SDK support for proofset roots, forcing developers to write manual API calls:

```typescript
// Manual Curio API call
try {
  const response = await fetch(`${providerUrl}pdp/proof-sets/${proofsetId}`, {
    method: "GET",
    headers: { Accept: "application/json" },
  });
  if (!response.ok) {
    if (response.status === 404) return null;
    throw new Error(`HTTP error! status: ${response.status}`);
  }
  return await response.json();
} catch (error) {
  console.error(`Error fetching proofset ${proofsetId}:`, error);
  return null;
}
```

## Proposed Solution

1. **`getProofsetRoots`**

   ```typescript
   async getProofsetRoots(proofsetId: string): Promise<string[]>
   ```

2. **`getProofsetRootsWithMetadata`**
   _(after resolving [#86](https://github.com/FilOzone/synapse-sdk/issues/86))_

   ```typescript
   async getProofsetRootsWithMetadata(proofsetId: string): Promise<ProofsetRoot[]>

   interface ProofsetRoot {
     rootCid: string;
     metadata?: string;
   }
   ```

3. **`calculateProofsetRoots`**

If possible given proofsetData

```typescript
async calculateProofsetRoots(proofsetData: ProofsetData): Promise<ProofsetRoot[]>
```

### Benefit

Enables fast dashboard creation by providing immediate access to uploaded data IDs, which can be retrieved via the SDK using `download(commP)`.

virajbhartiya:
- Add getProofSet() method to PDPServer for API calls
- Add getProofsetRoots() and getProofsetRootsWithMetadata() to StorageService
- Expose proofset methods in main Synapse class
- Add ProofsetData and ProofsetRoot type interfaces
- Eliminates need for manual Curio API calls

Resolves feature request for proofset root SDK methods

Closes #102

### DIFF
--- a/src/pdp/server.ts
+++ b/src/pdp/server.ts
@@ -28,7 +28,7 @@
 
 import { ethers } from 'ethers'
 import type { PDPAuthHelper } from './auth.js'
-import type { RootData, CommP } from '../types.js'
+import type { RootData, CommP, ProofsetData } from '../types.js'
 import { asCommP, calculate as calculateCommP, downloadAndValidateCommP } from '../commp/index.js'
 import { constructPieceUrl, constructFindPieceUrl } from '../utils/piece.js'
 import { MULTIHASH_CODES } from '../utils/index.js'
@@ -513,6 +513,31 @@ export class PDPServer {
 
     // Use the shared download and validation function
     return await downloadAndValidateCommP(response, parsedCommP)
+  }
+
+  /**
+   * Get proof set details from the PDP server
+   * @param proofSetId - The ID of the proof set to fetch
+   * @returns Promise that resolves with proof set data
+   */
+  async getProofSet (proofSetId: string): Promise<ProofsetData> {
+    const response = await fetch(`${this._apiEndpoint}/pdp/proof-sets/${proofSetId}`, {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json'
+      }
+    })
+
+    if (response.status === 404) {
+      throw new Error(`Proof set not found: ${proofSetId}`)
+    }
+
+    if (!response.ok) {
+      const errorText = await response.text()
+      throw new Error(`Failed to fetch proof set: ${response.status} ${response.statusText} - ${errorText}`)
+    }
+
+    return await response.json() as ProofsetData
   }
 
   /**
--- a/src/storage/service.ts
+++ b/src/storage/service.ts
@@ -19,7 +19,9 @@ import type {
   UploadCallbacks,
   UploadResult,
   RootData,
-  CommP
+  CommP,
+  ProofsetData,
+  ProofsetRoot
 } from '../types.js'
 import type { Synapse } from '../synapse.js'
 import type { PandoraService } from '../pandora/service.js'
@@ -923,5 +925,30 @@ export class StorageService {
    */
   async download (commp: string | CommP, options?: DownloadOptions): Promise<Uint8Array> {
     return await this.providerDownload(commp, options)
+  }
+
+  /**
+   * Get proofset roots for a given proofset ID
+   * @param proofsetId - The ID of the proof set to fetch roots for
+   * @returns Array of root CIDs
+   */
+  async getProofsetRoots (proofsetId: string): Promise<string[]> {
+    const proofsetData = await this._pdpServer.getProofSet(proofsetId)
+    return proofsetData.roots || []
+  }
+
+  /**
+   * Get proofset roots with metadata for a given proofset ID
+   * @param proofsetId - The ID of the proof set to fetch roots for
+   * @returns Array of roots with metadata
+   */
+  async getProofsetRootsWithMetadata (proofsetId: string): Promise<ProofsetRoot[]> {
+    const proofsetData = await this._pdpServer.getProofSet(proofsetId)
+    const roots = proofsetData.roots || []
+    
+    return roots.map((rootCid, index) => ({
+      rootCid,
+      metadata: proofsetData.metadata?.[index]
+    }))
   }
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -292,3 +292,25 @@ export interface UploadResult {
   /** Root ID in the proof set */
   rootId?: number
 }
+
+/**
+ * Proofset data returned from the API
+ */
+export interface ProofsetData {
+  /** The proof set ID */
+  proofSetId: number
+  /** Array of root CIDs in the proof set */
+  roots: string[]
+  /** Additional metadata (structure may vary) */
+  metadata?: any
+}
+
+/**
+ * Individual proofset root with metadata
+ */
+export interface ProofsetRoot {
+  /** The root CID */
+  rootCid: string
+  /** Optional metadata for this root */
+  metadata?: string
+}
--- a/src/synapse.ts
+++ b/src/synapse.ts
@@ -8,7 +8,8 @@ import {
   type StorageServiceOptions,
   type FilecoinNetworkType,
   type PieceRetriever,
-  type CommP
+  type CommP,
+  type ProofsetRoot
 } from './types.js'
 import { StorageService } from './storage/index.js'
 import { PaymentsService } from './payments/index.js'
@@ -284,6 +285,26 @@ export class Synapse {
     )
 
     return await downloadAndValidateCommP(response, parsedCommP)
+  }
+
+  /**
+   * Get proofset roots for a given proofset ID
+   * @param proofsetId - The ID of the proof set to fetch roots for
+   * @returns Array of root CIDs
+   */
+  async getProofsetRoots (proofsetId: string): Promise<string[]> {
+    const storage = await this.createStorage()
+    return await storage.getProofsetRoots(proofsetId)
+  }
+
+  /**
+   * Get proofset roots with metadata for a given proofset ID
+   * @param proofsetId - The ID of the proof set to fetch roots for
+   * @returns Array of roots with metadata
+   */
+  async getProofsetRootsWithMetadata (proofsetId: string): Promise<ProofsetRoot[]> {
+    const storage = await this.createStorage()
+    return await storage.getProofsetRootsWithMetadata(proofsetId)
   }
 }
 


src/types.ts
```diff
@@ -292,3 +292,25 @@ export interface UploadResult {
...
+  /** Array of root CIDs in the proof set */
+  roots: string[]
+  /** Additional metadata (structure may vary) */
+  metadata?: any
+}
+
+/**
+ * Individual proofset root with metadata
+ */
+export interface ProofsetRoot {
```
rvagg, line 311:
```suggestion
 * Proof set data returned from the API
 */
export interface ProofSetData {
  /** The proof set ID */
  proofSetId: number
  /** Array of root CIDs in the proof set */
  roots: string[]
  /** Additional metadata (structure may vary) */
  metadata?: any
}

/**
 * Individual proof set root with metadata
 */
export interface ProofSetRoot {
```

src/pdp/server.ts
```diff
@@ -515,6 +515,31 @@ export class PDPServer {
...
+    if (response.status === 404) {
+      throw new Error(`Proof set not found: ${proofSetId}`)
+    }
+
+    if (!response.ok) {
+      const errorText = await response.text()
+      throw new Error(`Failed to fetch proof set: ${response.status} ${response.statusText} - ${errorText}`)
+    }
+
+    return await response.json() as ProofsetData
```
rvagg, line 540: Noticing this made me remember that we're not doing any validation. So have a look at https://github.com/FilOzone/synapse-sdk/pull/113 for validation for the other responses so far. I'd like you to add the same for this PR. You could either wait for that PR to merge, or `git rebase upstream rvagg/validate-server` this branch and it should fit fairly cleanly on it and then you can add your new set of functions for this.

src/types.ts
```diff
@@ -292,3 +292,25 @@ export interface UploadResult {
...
+/**
+ * Proofset data returned from the API
+ */
+export interface ProofsetData {
+  /** The proof set ID */
+  proofSetId: number
+  /** Array of root CIDs in the proof set */
+  roots: string[]
+  /** Additional metadata (structure may vary) */
+  metadata?: any
```
rvagg, line 305: this doesn't match the response we're dealing with.

See: https://github.com/filecoin-project/curio/blob/d359767f4beacbbba6ae90c86e5b4067d03da946/pdp/handlers.go#L492-L496

And also https://github.com/filecoin-project/curio/blob/2ea28495a0409c088835143ddddcb540a317925a/pdp/handlers.go#L521-L527

So your type here is going to need to be more structured.

If you want some test data to get this right, have a look at the proof set that I've been adding to here: `curl https://polynomial.computer/pdp/proof-sets/292 | jq`.

When building a validator for this data, you should validate the CIDs are actually CIDs. `isValidCommP` is in src/commp.ts that you can use for that.

rvagg:
You're going to need to wind back the scope of this a bit @virajbhartiya. As noted in the original issue, https://github.com/FilOzone/synapse-sdk/issues/102, the metadata is a bit more complicated than just being able to get it from the server. If you follow the link to https://github.com/FilOzone/synapse-sdk/issues/86 you'll see that we're dealing with metadata that goes into the Pandora contract. Currently we don't store this on the PDP server so we can't ask for it, we also don't encode it in a nice way to retrieve it. We could just make one up with `withCDN` because that's all we care about right now, but I think we should just defer that. So for now, let's just limit it to getting the proofset data from the server and limit it there.

Also, one of the things I'd like to try and keep consistent is that whenever we return a CID, we do it as a `CID` type, not a `string`; although we accept `string|CID` from the user. You can use `asCommP` to do both validation and conversion when receiving the data from the server and returning it to the user.

src/storage/service.ts
```diff
@@ -924,4 +925,29 @@ export class StorageService {
   async download (commp: string | CommP, options?: DownloadOptions): Promise<Uint8Array> {
     return await this.providerDownload(commp, options)
   }
+
+  /**
+   * Get proofset roots for a given proofset ID
+   * @param proofsetId - The ID of the proof set to fetch roots for
+   * @returns Array of root CIDs
+   */
+  async getProofsetRoots (proofsetId: string): Promise<string[]> {
```
rvagg, line 934: I don't really see the utility of this since it's not a complex operation. I'm asking @nijoe1 here what he's wanting from this: https://github.com/FilOzone/synapse-sdk/issues/102#issuecomment-2994830438

_Maybe_ it makes sense given that the schema is more complicated than this PR suggests, but it would be something more like `proofSetData.roots.map((r) => r.rootCid)`. Leave it for now (remove the Metadata call below as per my other comment) and implement it properly and we'll have a look at it.

rvagg:
Also will need tests to cover this. See the other tests for pdp server interactions for a model.

virajbhartiya:
Hey @rvagg addressed the changes mentioned above. added tests as well

src/storage/service.ts
```diff
@@ -924,4 +925,24 @@ export class StorageService {
   async download (commp: string | CommP, options?: DownloadOptions): Promise<Uint8Array> {
     return await this.providerDownload(commp, options)
   }
+
+  /**
+   * Get proof set roots for a given proof set ID
+   * @param proofSetId - The ID of the proof set to fetch roots for
+   * @returns Array of root CIDs as CommP objects
+   */
+  async getProofSetRoots (proofSetId: string): Promise<CommP[]> {
```
rvagg, line 934: This shouldn't take an argument, we have `this._proofSetId`. A `StorageService` is tied to a proofset. If we take an arbitrary one but we're only querying the one server then we may not be able to get the info about it. A proper "getProofSetRoots" taking a param might have to look up the chain to see the PDP server to query.

rvagg:
@virajbhartiya I did a manual merge here since master has changed but I think I messed it up. Feel free to rebase locally and force push a new version if you'd prefer to clean it up yourself. Otherwise hopefully dealing with whatever I messed up isn't hard.

src/pdp/validation.ts
```diff
@@ -191,3 +192,91 @@ export function validateFindPieceResponse (value: unknown): FindPieceResponse {
...
+
+  // Required field - roots (array of ProofSetRootData)
+  if (!Array.isArray(obj.roots)) {
+    return false
+  }
+  for (const root of obj.roots) {
+    if (!isProofSetRootData(root)) {
+      return false
+    }
+  }
```
rvagg, line 262:
```suggestion
  if (!Array.isArray(obj.roots) || !obj.roots.every(isProofSetRootData)) {
    return false
  }
```

src/types.ts
```diff
@@ -292,3 +292,29 @@ export interface UploadResult {
...
+/**
+ * Individual proof set root data from API
+ */
+export interface ProofSetRootData {
+  /** Root ID within the proof set */
+  rootId: number
+  /** The root CID */
+  rootCid: string
+  /** Sub-root CID (usually same as rootCid) */
+  subrootCid: string
```
rvagg, line 317:
```suggestion
  rootCid: string
  /** Sub-root CID (usually same as rootCid) */
  subrootCid: string
```

I wouldn't mind these converted to `CommP`

I think the best way to achieve this is to replace `validateProofSetData` _and_ `isProofSetData` with an `asProofSetData` and use the same pattern as the `asCommP` - that is, don't throw, just return `null` if it doesn't validate, but as you are validating, you are also building the new form, basically copying everything over except for these CIDs which you'll run `asCommP` on and get non-`null`.

So far (as far as I can remember), the validators don't do any conversion, so this is a new pattern, which is OK, just name it differently because you're not doing an `is` boolean, you're doing an `as` conversion which may or may not work.

rjan90:
Aside from the conflict that needs to be resolved, do the latest changes from @virajbhartiya address your request to return the commP type instead of a string, @rvagg?

src/pdp/validation.ts
```diff
@@ -191,3 +192,104 @@ export function validateFindPieceResponse (value: unknown): FindPieceResponse {
...
+    nextChallengeEpoch: obj.nextChallengeEpoch
+  }
+}
+
+/**
+ * Validates and returns a ProofSetData
+ * @param value - The value to validate and convert
+ * @throws Error if validation fails
+ */
+export function validateProofSetData (value: unknown): ProofSetData {
```
rvagg, line 289: ditch this function and just do what you're doing inside this where you call it, so we only have an `as`, not a `validate` for the cases where we always want to transform

src/storage/service.ts
```diff
@@ -924,4 +925,23 @@ export class StorageService {
...
+    const proofSetData = await this._pdpServer.getProofSet(this._proofSetId)
+    return proofSetData.roots.map(root => {
+      const commP = asCommP(root.rootCid)
+      if (commP == null) {
+        throw createError(
+          'StorageService',
+          'getProofSetRoots',
+          `Invalid CommP received from server: ${root.rootCid.toString()}`
+        )
+      }
```
rvagg, line 943: don't we know that it's a CommP already from our `asProofSet` conversion? double-check that, maybe all you need to do is `map(root => root.rootCid)` and you're good to go?

src/storage/service.ts
```diff
@@ -1001,4 +1001,13 @@ export class StorageService {
...
   }
+
+  /**
+   * Get proof set roots for this storage service's proof set
+   * @returns Array of root CIDs as CommP objects
+   */
+  async getProofSetRoots (): Promise<CommP[]> {
+    const proofSetData = await this._pdpServer.getProofSet(this._proofSetId)
+    return proofSetData.roots.map(root => root.rootCid)
+  }
```
rvagg, line 1012:
```suggestion
  /**
   * Get proof set info for this storage service's proof set
   * @returns ProofSetData containing metadata for this proof set and its current roots
   */
  async getProofSetInfo (): Promise<ProofSetData> {
    return await this._pdpServer.getProofSet(this._proofSetId)
  }

  /**
   * Get proof set roots for this storage service's proof set
   * @returns Array of root CIDs as CommP objects
   */
  async getProofSetRoots (): Promise<CommP[]> {
    const proofSetData = await this.getProofSetInfo()
    return proofSetData.roots.map(root => root.rootCid)
  }
```

rvagg:
nmind, I see them:
* https://github.com/FilOzone/synapse-sdk/commit/ac125d79f1e004c15cfdd580bb54c9e8bd8fe6a8
* https://github.com/FilOzone/synapse-sdk/commit/efac08a4fedfa19911c61788175a24e352e205d5
* https://github.com/FilOzone/synapse-sdk/commit/ee8c3e62a3f363a7f530907a539b411b3e7c38d9

taken care of, I'll get the rest of it done

PR #113, opened 2025-06-23
chore(pdp): validate server responses

rvagg:
Matches Curio responses in https://github.com/filecoin-project/curio/blob/main/pdp/handlers.go

Also handle the piece_cid -> pieceCid transition in https://github.com/filecoin-project/curio/pull/547 - we're not dependent on that PR landing but if it doesn't land and this does then we'll need to wind back the `pieceCid` thing in here.


rvagg:
also added a commit that will validate and convert CommP strings to `CommP` objects (a variant of `CID`), since we have to validate them anyway we may as well convert so we're consistent

PR #115, opened 2025-06-23
feat(retriever): add SubgraphRetriever

silent-cipher:
This PR introduces the `SubgraphRetriever`, a new component that allows fetching piece information from a user-configured subgraph.

It can be chained with other retrievers, such as `ChainRetriever` and `FilCDNRetriever`, to create a flexible and powerful data discovery pipeline. This provides an alternative for users who rely on subgraphs for `commp` to provider mapping.

This also contains an `SubgraphService` that can be used independently - 
```ts
import { SubgraphService } from '@filoz/synapse-sdk'

const subgraphUrl = 'https://api.thegraph.com/subgraphs/name/some/subgraph'
const subgraphService = new SubgraphService(subgraphUrl)

// Example: Query for active providers with custom filtering
const activeProviders = await subgraphService.queryProviders({
  where: {
    status: 'Approved',
  },
  orderBy: 'totalProofSets',
  orderDirection: 'desc',
  first: 10
})

console.log('Top 10 active providers by proof sets:', activeProviders)
```


src/retriever/chain.ts
```diff
@@ -68,123 +69,52 @@ export class ChainRetriever implements PieceRetriever {
...
-          'ChainRetriever',
-          'fetchFromProviders',
-          `All providers failed to serve piece ${commp.toString()}. Details: ${failureDetails}`
-        )
-      }
-      // Re-throw unexpected errors
-      throw error
+      // Provider discovery failed - this is a critical error
+      return await tryChildOrThrow(
+        'Provider discovery failed and no child retriever was configured'
```
rvagg, line 96:
```suggestion
        'Provider discovery failed and no additional retriever method was configured'
```

"child" is an implementation detail, imagine an end-user seeing this

src/retriever/chain.ts
```diff
@@ -68,123 +69,52 @@ export class ChainRetriever implements PieceRetriever {
...
-  async fetchPiece (
-    commp: CommP,
-    client: string,
-    options?: {
-      providerAddress?: string
-      withCDN?: boolean
-      signal?: AbortSignal
+    // Step 2: If no providers found, try child retriever
+    if (providersToTry.length === 0) {
+      return await tryChildOrThrow('No providers found and no child retriever was configured')
```
rvagg, line 102:
```suggestion
      return await tryChildOrThrow('No providers found and no additional retriever method was configured')
```

src/retriever/chain.ts
```diff
@@ -68,123 +69,52 @@ export class ChainRetriever implements PieceRetriever {
...
+      return await fetchPiecesFromProviders(
+        providersToTry,
+        commp,
+        'ChainRetriever',
+        options?.signal
+      )
+    } catch (fetchError) {
+      // All provider attempts failed
+      return await tryChildOrThrow(
+        'All provider retrieval attempts failed and no child retriever was configured'
```
rvagg, line 116:
```suggestion
        'All provider retrieval attempts failed and no additional retriever method was configured'
```

src/retriever/subgraph.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    // Step 1: Find providers
+    let providersToTry: ApprovedProviderInfo[] = []
+    try {
+      providersToTry = await this.findProviders(commp, options?.providerAddress)
+    } catch (error) {
+      // Provider discovery failed - this is a critical error
+      return await tryChildOrThrow(
+        'Provider discovery failed and no child retriever was configured'
```
rvagg, line 58:
```suggestion
        'Provider discovery failed and no additional retriever method was configured'
```

src/retriever/subgraph.ts
```diff
@@ -0,0 +1,82 @@
...
+      )
+    }
+
+    // Step 2: If no providers found, try child retriever
+    if (providersToTry.length === 0) {
+      return await tryChildOrThrow('No providers found and no child retriever was configured')
```
rvagg, line 64:
```suggestion
      return await tryChildOrThrow('No providers found and no additional retriever method was configured')
```

src/utils/providerFetch.ts
```diff
@@ -0,0 +1,129 @@
+/**
```
rvagg, line 1: Move this file into src/retriever, and don't use camelCase for filenames, maybe just make it utils.ts in there

src/types.ts
```diff
@@ -108,6 +110,45 @@ export interface PieceRetriever {
...
+  }
+  /** Optional API key for authenticated subgraph access */
+  apiKey?: string
+}
+
+/**
+ * Defines the contract for a service that can retrieve provider information from a data source,
+ * typically a Synapse-compatible subgraph.
+ */
+export interface SubgraphRetrievalService {
```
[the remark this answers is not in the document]
rvagg: Or, maybe you're suggesting that this service could be replaced with another subgraph service (other than goldsky?) and this is all someone would need to insert to make it work? That might be OK, but you should document that somewhere, either in the jsdoc here or in the README or both.

src/subgraph/service.ts
```diff
@@ -0,0 +1,822 @@
...
+    }
+  }
+
+  /**
+   * Converts CommP to hex string
+   */
+  private commPToHex (commP: CommP): string {
+    return toHex(CID.parse(commP.toString()).bytes)
+  }
+
```
rvagg, line 463: you have a `CommP` here, which is a type of `CID` so you're doing a cid->string->parse operation here that's most likely going to be useless because it's already done; but you can do a quick/cheap validation/parse with `asCommP` from commp.ts that ends up being a noop if it's already what you need:

```suggestion
  private commPToHex (commP: CommP): string {
    const parsedCommP = asCommP(commP)
    if (parsedCommP == null) {
      throw new Error(`Invalid CommP: ${String(expectedCommP)}`)
    }
    return toHex(parsedCommP.bytes)
  }
```

src/subgraph/service.ts
```diff
@@ -0,0 +1,822 @@
...
+   *
+   * It sends a GraphQL query to the configured endpoint and parses the response to extract
+   * a list of providers, including their addresses and retrieval URLs.
+   *
+   * @param commp - The piece commitment (CommP) to search for.
+   * @returns A promise that resolves to an array of `ApprovedProviderInfo` objects.
+   *          Returns an empty array if no providers are found or if an error occurs during the fetch.
+   */
+  async getProvidersForCommP (commP: CommP): Promise<ApprovedProviderInfo[]> {
+    const hexCommP = this.commPToHex(commP)
```
rvagg, line 511: it might be better if you do the `asCommP` (and `=== null` check) here that I had above in `commPToHex`, then you can return an error from here if you don't get one, and `commPToHex` can just assume a `CommP` (in fact, you probably don't even need that function since it's just a `toHex(commP.bytes)`

src/subgraph/service.ts
```diff
@@ -0,0 +1,822 @@
...
+      .map((provider) => this.transformProviderData(provider))
+  }
+
+  /**
+   * Queries the subgraph to find a specific approved storage provider by their address.
+   *
+   * @param address - The wallet address of the provider to search for.
+   * @returns A promise that resolves to an `ApprovedProviderInfo` object if the provider is found, or `null` otherwise.
+   */
+  async getProviderByAddress (address: string): Promise<ApprovedProviderInfo | null> {
```
rvagg, line 542: is there any requirement on the address format here? does it need to be `0x` prefixed? does it matter? should you add some simple validation and/or docs about that?

src/subgraph/service.ts
```diff
@@ -0,0 +1,822 @@
...
+      return []
+    }
+
+    return data.roots.map((root) => ({
+      id: root.id,
+      setId: this.parseTimestamp(root.setId),
+      rootId: this.parseTimestamp(root.rootId),
+      rawSize: this.parseTimestamp(root.rawSize),
+      leafCount: this.parseTimestamp(root.leafCount),
+      cid: root.cid,
```
rvagg, line 745: as long as you're doing this, you should probably return CIDs as `CID`s, ideally `CommP`s - use `asCommP` to validate and parse, it doesn't throw, just returns `null`, but since you're not validating the rest then maybe that's OK for now. The only concern will be when we switch to CommPv2 but we might just make that backward compatible (ish).

src/types.ts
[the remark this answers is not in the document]
silent-cipher: Yes, this was for the users wanted to replace with there own subgraph service in subgraph retriever.
I'll make sure to add this to docs

src/synapse.ts
```diff
@@ -135,8 +136,23 @@ export class Synapse {
     if (options.pieceRetriever != null) {
       pieceRetriever = options.pieceRetriever
     } else {
-      const chainRetriever = new ChainRetriever(pandoraService)
-      pieceRetriever = new FilCdnRetriever(chainRetriever, network)
+      const chainRetriever = new ChainRetriever(pandoraService /*, no child here */)
+      let underlyingRetriever: PieceRetriever = chainRetriever
+
+      if (options.subgraphConfig != null) {
```
rvagg, line 142: OK, so this isn't quite what I imagined, but this might be OK. One minor flaw in this design is that we're locking users in to using the `SubgraphService` you've defined already - if you intended it to be replaceable by an alternative service (e.g. `TheGraphService`) then you'd have to set up your retriever chain externally. But maybe that's OK. _If_ you didn't intend for `SubgraphService` to be replaceable and imagine it to be extensible in itself then it just reinforces my point about not exporting an interface for it from types.ts.

Not asking for action here, just noting this for further consideration before merging. This might be the easiest path forward for usability so you don't have to mess around with setting up an override retriever like I originally imagined.

silent-cipher: I was expecting user to pass `pieceRetriever` chain if he want to replace with his own service instead of using our define `subgraphService`
or he can just pass the `subgraphConfig` to use our defined service with `subgraphRetriever` 

And even this was the flow before `subgraphRetriever` so didn't think much about it- If you've any other easier way for users to use it please let me know

src/subgraph/service.ts
silent-cipher: I can get rid of `commPToHex` function to simply replace it with `toHex(commP.bytes)`

src/subgraph/service.ts
silent-cipher: I think I would need your help for this - to convert hexCommp back to `commP` I can use this - `CID.decode(cidBytes).toString()` 
I'm not sure of it's compatibility

src/synapse.ts
rvagg: I don't have a better suggestion, but there is a mild smell here. Let me lay out the optionality we're opening up:

* `pieceRetriever` lets you do your own custom thing to completely override our construction here
* `subgraphConfig` is specifically for our opinionated `SubgraphRetriever`
* We export the `SubgraphRetrievalService` for you to override your the "service" argument on `SubgraphRetriever` (but it's not part of the Synapse options)

Further:

* In the concrete `SubgraphService` we have very strong opinions about what the queries will look like and how to achieve the desired results
* It requires a `goldsky` config option to even work, _and_ we push this option up to be required by the `Synapse` constructor
  * i.e. it looks like maybe it's actually a `GoldskySubgraphService` but we're pretending that it's configurable; even mentioning `api.thegraph.com` in the README as if it could be portable.
  * Is `endpoint` variable from `goldsky`? Is there potentially more ways to call them than a single URL? (maybe you can pay for your own hosted version? or run your own?)

Thinking out loud here I think maybe being more honest about `Goldsky` would help us here - if we make a `GoldskySubgraphService` and expose an easy way to replace that with something else, e.g. `TheGraphSubgraphService` then it might _feel_ a bit better. Perhaps the `SubgraphOptions` could be something like: `{ subgraphService: SubgraphRetrievelService } | { goldsky: GoldskySubgraphServiceOptions }`, i.e. you either provide your own service, or configure the one that we ship by default for you.

src/subgraph/service.ts
```diff
@@ -0,0 +1,645 @@
...
+    const parsed = Number(value)
+    return isNaN(parsed) ? 0 : parsed
+  }
+
+  /**
+   * Safely converts a hex format CID to CommP format
+   * @param hexCid - The CID in hex format
+   * @returns The CID in CommP format or null if conversion fails
+   */
+  private safeConvertHexToCid (hexCid: string): CommP | null {
```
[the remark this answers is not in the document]
rvagg: I suppose a lot of the code in here could be static and changing them all would be a bigger churn than necessary, so we'll leave it be

src/synapse.ts
```diff
@@ -135,8 +137,34 @@ export class Synapse {
...
+          if (options.subgraphService != null) {
+            subgraphService = options.subgraphService
+          } else if (options.subgraphConfig != null) {
+            subgraphService = new SubgraphService(options.subgraphConfig)
+          } else {
+            // This shouldn't happen due to the if condition above, but TypeScript doesn't know that
+            throw new Error('Invalid subgraph configuration: neither service nor config provided')
+          }
+
+          underlyingRetriever = new SubgraphRetriever(subgraphService, chainRetriever)
```
rvagg, line 157: How about this, does TS complain at this?

```suggestion
          const subgraphService: SubgraphRetrievalService =
            options.subgraphService ??
            (options.subgraphConfig ? new SubgraphService(options.subgraphConfig) : undefined)!;
      
          underlyingRetriever = new SubgraphRetriever(subgraphService, chainRetriever);
```

silent-cipher: Yes, ts-eslint will complain for the use of `non-null-assertion` - `(!)`

src/retriever/utils.ts
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Attempt to fetch a piece from multiple providers in parallel
+ * @param providers - List of providers to try
+ * @param commp - The piece to fetch
+ * @param retrieverName - Name of the calling retriever for error reporting
+ * @param signal - Optional abort signal
+ * @returns The first successful response
+ */
+export async function fetchPiecesFromProviders (
+  providers: ApprovedProviderInfo[],
```
rvagg, line 24: let's add a dedupe for `providers` inside this function, as I mentioned in Slack, it's possible to get the same provider more than once in here if oyu've stored the same piece with them more than once

rvagg: actually, maybe we should do it up in the subgraph retriever, if you look in chain.ts you'll see that we already do it there

if we can't do a uniqueness check in the graphql query then let's do it when we get the response

PR #118, opened 2025-06-23
feat: add ping method to check connectivity with storage provider

virajbhartiya:
## Add provider ping functionality to PDPServer

Implements basic provider connectivity validation by adding a `ping()` method to the `PDPServer` class. The method:

- Makes a GET request to the provider's `/ping` endpoint
- Validates 200 response status 
- Throws errors for non-200 responses or connection failures

This enables checking provider reachability before proceeding with storage operations. Part of #116 - provider selection improvements.

**Related:** Additional PR will follow with provider selection retry logic that uses this ping functionality.

Issue #116, opened 2025-06-23
Deal with flaky storage providers


rvagg:
This is primarily an M1/M2 question, and we need a more sophisticated selection mechanism at some point, but for now select your storage provider by:

* If you have an existing proofset that matches your parameters (`withCDN` is the only parameter we care about atm), then we select that provider and proofset
* If you don't have one then we randomly assign you one and that's all you get

But, in either case, if we can't connect to your storage provider then we have problems. We currently have one of our providers in the Pandora contract having connectivity issues.

I think maybe in here we should do a ping to the provider after we select them before we proceed after a `createStorage`. If we can't connect to the provider, then we should try again and exclude that provider from selection. Which might mean that it's the provider we were dealing with previously but now we have to set up with a new provider.

We also need to think of the UX/DX beyond M1/M2, how should you go about selecting a provider? What tools should we give you? I think at least being able to ping them would be good. So, I think a good outcome here is a `/ping` interface in pdp/server.ts that can be used when selecting a provider. So we should at least build that. The handler is [here](https://github.com/filecoin-project/curio/blob/8bf69f611fd4127b61826e7db11712eb376498ca/pdp/handlers.go#L118-L129), we're just looking for a 200.

* [ ] Create a method that does a /ping for a given provider in our PDP server tool
* [ ] Use the /ping method after selecting a provider to ensure we can talk to them, if not, we should select again but explicitly exclude them, and continue in a loop until we exhaust the list and have no more providers to try

rjan90:
> Use the /ping method after selecting a provider to ensure we can talk to them, if not, we should select again but explicitly exclude them, and continue in a loop until we exhaust the list and have no more providers to try

https://github.com/FilOzone/synapse-sdk/pull/119 implements this feature.

rjan90:
Now that #119 has been merged, should we close this issue and instead open a design discussion, focused on UX/DX beyond M2, about how we should approach selecting a provider?

PR #119, opened 2025-06-23
Feature/provider ping validation

virajbhartiya:
## Implement provider ping validation with retry logic

Enhances provider selection to validate connectivity before proceeding with storage operations. Builds on the ping functionality from the previous PR to implement robust provider selection with fallback logic.

**Key Features:**
- **Existing provider validation**: Pings existing providers before reusing them, falls back to new selection if unreachable
- **New provider selection with retry**: Randomly selects providers but excludes failed ones, retries until finding a reachable provider
- **Comprehensive error handling**: Graceful fallbacks with detailed logging when providers fail connectivity tests
- **Full test coverage**: Tests for ping validation, retry logic, exclusion handling, and fallback scenarios

**Behavior:**
1. For existing proofsets: Tests each provider in preference order, selects first that responds to ping
2. For new proofsets: Randomly selects from available providers, excludes failed ones, continues until exhausting all options
3. Throws errors only when no providers are reachable

**Part of #116** - Addresses connectivity issues with storage providers by ensuring only reachable providers are selected.

**Depends on:** #118

### DIFF
--- a/src/pdp/server.ts
+++ b/src/pdp/server.ts
@@ -133,11 +133,14 @@ export class PDPServer {
     retrievalEndpoint: string,
     serviceName: string = 'public'
   ) {
-    if (apiEndpoint.trim() === '') {
-      throw new Error('PDP API endpoint is required')
-    }
-    if (retrievalEndpoint.trim() === '') {
-      throw new Error('PDP retrieval endpoint is required')
+    // Allow empty endpoints only for null auth (ping validation scenarios)
+    if (authHelper !== null) {
+      if (apiEndpoint.trim() === '') {
+        throw new Error('PDP API endpoint is required')
+      }
+      if (retrievalEndpoint.trim() === '') {
+        throw new Error('PDP retrieval endpoint is required')
+      }
     }
     // Remove trailing slash from endpoints
     this._apiEndpoint = apiEndpoint.replace(/\/$/, '')
--- a/src/pdp/server.ts
+++ b/src/pdp/server.ts
@@ -117,7 +117,7 @@ export interface RootAdditionStatusResponse {
 export class PDPServer {
   private readonly _apiEndpoint: string
   private readonly _retrievalEndpoint: string
-  private readonly _authHelper: PDPAuthHelper
+  private readonly _authHelper: PDPAuthHelper | null
   private readonly _serviceName: string
 
   /**
@@ -128,7 +128,7 @@ export class PDPServer {
    * @param serviceName - Service name for uploads (defaults to 'public')
    */
   constructor (
-    authHelper: PDPAuthHelper,
+    authHelper: PDPAuthHelper | null,
     apiEndpoint: string,
     retrievalEndpoint: string,
     serviceName: string = 'public'
@@ -161,13 +161,13 @@ export class PDPServer {
     recordKeeper: string
   ): Promise<CreateProofSetResponse> {
     // Generate the EIP-712 signature for proof set creation
-    const authData = await this._authHelper.signCreateProofSet(clientDataSetId, payee, withCDN)
+    const authData = await this.getAuthHelper().signCreateProofSet(clientDataSetId, payee, withCDN)
 
     // Prepare the extra data for the contract call
     // This needs to match the ProofSetCreateData struct in Pandora contract
     const extraData = this._encodeProofSetCreateData({
       metadata: '', // Empty metadata for now
-      payer: await this._authHelper.getSignerAddress(),
+      payer: await this.getAuthHelper().getSignerAddress(),
       withCDN,
       signature: authData.signature
     })
@@ -251,7 +251,7 @@ export class PDPServer {
     }
 
     // Generate the EIP-712 signature for adding roots
-    const authData = await this._authHelper.signAddRoots(
+    const authData = await this.getAuthHelper().signAddRoots(
       clientDataSetId,
       nextRootId,
       rootDataArray // Pass RootData[] directly to auth helper
@@ -594,6 +594,9 @@ export class PDPServer {
   }
 
   getAuthHelper (): PDPAuthHelper {
+    if (this._authHelper == null) {
+      throw new Error('Auth helper is not initialized')
+    }
     return this._authHelper
   }
 }
--- a/src/pdp/server.ts
+++ b/src/pdp/server.ts
@@ -578,7 +578,7 @@ export class PDPServer {
    * @throws Error if provider is not reachable or returns non-200 status
    */
   async ping (): Promise<void> {
-    const response = await fetch(`${this._apiEndpoint}/ping`, {
+    const response = await fetch(`${this._apiEndpoint}/pdp/ping`, {
       method: 'GET',
       headers: {}
     })
--- a/src/pdp/server.ts
+++ b/src/pdp/server.ts
@@ -133,14 +133,11 @@ export class PDPServer {
     retrievalEndpoint: string,
     serviceName: string = 'public'
   ) {
-    // Allow empty endpoints only for null auth (ping validation scenarios)
-    if (authHelper !== null) {
-      if (apiEndpoint.trim() === '') {
-        throw new Error('PDP API endpoint is required')
-      }
-      if (retrievalEndpoint.trim() === '') {
-        throw new Error('PDP retrieval endpoint is required')
-      }
+    if (apiEndpoint.trim() === '') {
+      throw new Error('PDP API endpoint is required')
+    }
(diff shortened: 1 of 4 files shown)


src/storage/service.ts
```diff
@@ -627,33 +636,110 @@ export class StorageService {
...
+    provider: ApprovedProviderInfo,
+    signer: ethers.Signer
+  ): Promise<void> {
+    // Create a temporary PDPAuthHelper and PDPServer just for pinging
+    // We need these to create a PDPServer instance, but for ping we don't need valid auth
+    const tempPandoraAddress = '0x0000000000000000000000000000000000000000' // Dummy address for ping
+    const tempChainId = BigInt(1) // Dummy chain ID for ping
+
+    const authHelper = new PDPAuthHelper(tempPandoraAddress, signer, tempChainId)
+    const pdpServer = new PDPServer(authHelper, provider.pdpUrl, provider.pieceRetrievalUrl)
```
rvagg, line 739: OK, this is kind of gross.

~We don't need an authHelper so let's go into pdp/server.ts and make the constructor take `authHelper: PDPAuthHelper | null`; then in `getAuthHelper` at the bottom of that file, make it `throw` if `this._authHelper == null`. Then replace the 3 instances of `this._authHelper` with `this.getAuthHelper`. After all that, you can `new PDPServer(null, provider.pdpUrl, provider.pieceRetrievalUrl)` without all the rest of the stuff in here.~

Actually, we have `this._pdpServer` in `StorageService` which we should be using instead of making a new one.

Two things to try: first, chase these `static` methods back up and see if it's just going to be easier to make them instance methods rather than statics. If not, then pass down `pdpServer` as an argument so we can use it here. Once you have a `pdpServer` then you shouldn't even need this method, you can just `pdpServer.ping()` where you're currently calling `pingProvider`

rvagg:
Higher level critique here: you now have duplicated logic, which is simple, but takes up a bunch of lines of code so is worth simplifying. I think what you should do is have a new function that can choose amongst a sorted list of providers and call that in both places you need to try each one sequentially. That'll be easy to call from smartSelect but for the randomSelect you'll need to randomly sort the list of providers using the same randomness technique (it could be something like: create new empty list, insert random element from old list, repeat until old list is exhausted).

src/storage/service.ts
```diff
@@ -627,9 +657,82 @@ export class StorageService {
   }
 
   /**
-   * Select a random provider from the given list
+   * Select a random provider from the given list with ping validation
+   * @param providers - List of available providers
+   * @param signer - Signer for entropy generation
+   * @param excludedProviderAddresses - List of provider addresses to exclude from selection
+   * @param pdpServer - PDPServer instance for ping validation
```
rvagg, line 664: this jsdoc isn't accurate, you have `enablePingValidation` instead

src/storage/service.ts
```diff
@@ -121,12 +121,16 @@ export class StorageService {
     const signer = synapse.getSigner()
     const signerAddress = await signer.getAddress()
 
+    // Create a PDPServer instance for ping validation (with null auth since ping doesn't need it)
+    const pingPdpServer = new PDPServer(null, '', '')
+
```
rvagg, line 126: oh, of course .. sorry I lead you astray a little in my last comment @virajbhartiya, we don't have a `PDPServer` until we select an SP to talk to! So we need to wind back some of these changes, but the `| null` change is good for this. Let's move this `new PDPServer` back down to the ping operation so we do it on demand and not pass it down through the intermediate calls and have unnecessary arguments.

So remove from here, `resolveProviderAndProofSet`, `smartSelectProvider`, `selectRandomProvider`, `selectProviderWithPing`. Right at the bottom, in `selectProviderWithPing` you're not using this `pingPdpServer` anyway - you pass it down half way, make a new one in `selectRandomProvider` to pass down further, and then make a new one again in `selectProviderWithPing`!

So when you're done you should only see 2 instances of `new PDPServer` in this file - one for the per-SP one just doing the ping in `selectProviderWithPing`, and the one in the constructor just above this `create`.

src/storage/service.ts
```diff
@@ -584,25 +591,48 @@ export class StorageService {
...
-      const provider = await pandoraService.getApprovedProvider(providerId)
+      if (existingProviders.length > 0) {
+        try {
+          const selectedProvider = await StorageService.selectProviderWithPing(existingProviders, [], pdpServer)
 
-      return {
-        provider,
-        proofSetId: selected.pdpVerifierProofSetId,
-        isExisting: true
+          // Find the proof set ID for this provider
```
rvagg, line 611:
```suggestion
          // Find the first matching proof set ID for this provider
```

src/storage/service.ts
```diff
@@ -627,9 +657,82 @@ export class StorageService {
...
+   * @param excludedProviderAddresses - List of provider addresses to exclude from selection
+   * @param pdpServer - PDPServer instance for ping validation
+   * @returns A provider that responds to ping
+   * @throws Error if no providers are reachable
    */
   private static async selectRandomProvider (
+    providers: ApprovedProviderInfo[],
+    signer: ethers.Signer,
+    excludedProviderAddresses: string[] = [],
+    enablePingValidation: boolean = false
```
rvagg, line 672: both of these arguments are not being provided by the caller, I think you have a logic mixup here, maybe you don't need the `excludedProviderAddresses` here _or_ in `selectProviderWithPing`, just pass down your list to `selectProviderWithPing` and let it return the first in your sorted (or random) list and that's the one to go with; I think that works in both cases?

src/pdp/server.ts
```diff
@@ -128,16 +128,19 @@ export class PDPServer {
...
-      throw new Error('PDP retrieval endpoint is required')
+    // Allow empty endpoints only for null auth (ping validation scenarios)
+    if (authHelper !== null) {
+      if (apiEndpoint.trim() === '') {
+        throw new Error('PDP API endpoint is required')
+      }
+      if (retrievalEndpoint.trim() === '') {
+        throw new Error('PDP retrieval endpoint is required')
+      }
     }
```
rvagg, line 144:
```suggestion
    if (apiEndpoint.trim() === '') {
      throw new Error('PDP API endpoint is required')
    }
    if (retrievalEndpoint.trim() === '') {
      throw new Error('PDP retrieval endpoint is required')
    }
```

I think we can undo this now, no? We should always expect these parameters, it's not really useful without them.

src/pdp/server.ts
```diff
@@ -594,6 +597,9 @@ export class PDPServer {
   }
 
   getAuthHelper (): PDPAuthHelper {
+    if (this._authHelper == null) {
+      throw new Error('Auth helper is not initialized')
```
rvagg, line 601:
```suggestion
      throw new Error('AuthHelper is not available for an operation that requires signing')
```

src/storage/service.ts
```diff
@@ -627,33 +651,97 @@ export class StorageService {
...
+   * @param excludedProviderAddresses - List of provider addresses to exclude from selection
+   * @returns A provider that responds to ping
+   * @throws Error if no providers are reachable
+   */
+  private static async selectProviderWithPing (
+    sortedProviders: ApprovedProviderInfo[],
+    excludedProviderAddresses: string[]
+  ): Promise<ApprovedProviderInfo> {
+    // Filter out excluded providers
+    const availableProviders = sortedProviders.filter(p => !excludedProviderAddresses.includes(p.owner))
```
rvagg, line 716:
```suggestion
  /**
   * Select a provider from a sorted list with ping validation.
   * This is shared logic used by both smart selection and random selection.
   * @param sortedProviders - List of providers to try in order
   * @returns A provider that responds to ping
   * @throws Error if no providers are reachable
   */
  private static async selectProviderWithPing (sortedProviders: ApprovedProviderInfo[]): Promise<ApprovedProviderInfo> {
```

We're just not using the exclusion feature here are we? Let's ditch it, and from the tests too. Unless I'm missing something?

virajbhartiya:
@rvagg dosen't `${this._apiEndpoint}` include `/pdp`?

rvagg:
it's because in src/test/pdp-server.test.ts we do this: `pdpServer = new PDPServer(authHelper, serverUrl + '/pdp', serverUrl)`, which is dumb, because we append /pdp to everything anyway, but we want those URLs to be different for the purpose of the tests in there, but avoid the /pdp addition to make sure the pdp/server.ts adds it itself, same with /piece for retrievals.

I've pushed a small refactor to that test to clarify things, keeping the urls different but avoiding those suffixes so we can watch for the proper ones to be appended.

rvagg (approved):

PR #126, opened 2025-07-07
feat: benchmark timing analysis

### LINKED ISSUE #125: Profile PDP interaction times
Need to quantify the times it takes to do the various operations. Currently just create proof set and add roots, but we'll also have delete operations too.
We want to know how long they take and why. We should document this in the readme and also look for opportunities to shorten where possible.
Most of the time is waiting for transactions to land on chain and be confirmed, but the client and Curio have different timings for this, we should know exactly why. I think we currently have them both set for 1 confirmation but Curio still always seems slower.

virajbhartiya:
detailed timing instrumentation and a benchmark script for PDP operations
Closes #125


rjan90:
@virajbhartiya I think it would be useful to also collect the timings here, and not just adding scripts for it. Rod wrote down how many runs should be done, etc, in his comment here: https://github.com/FilOzone/synapse-sdk/issues/125#issuecomment-3043437881. Is this something you also plan to add in this PR?

rvagg:
I will say though, having a utility for anyone to run is quite useful, so don't undo all the work here. Aim for a start to make a no-op by default, and use Symbols for global access where we use globals. Then if you can get us some numbers and after that we'll look at how to retain the code changes in some form.

virajbhartiya:
> If we need something checked in for PLDG or some other way to signify amount of effort then we could somehow reflect that; even if it means merging into a separate branch, or just checking in a timing report.

@rvagg as for PLDG, the rating is given by the tech partner only and not number of PRs. So if you can communicate my work while filling up my rating form it'll be great. Thanks!

rvagg:
@claude review this PR

Issue #105, opened 2025-06-16
Make methods names more user-friendly


rjan90:
When users sign transactions via MetaMask in the demo-app: https://fs-upload-dapp.netlify.app, they see technical contract method names like "AddRoots" which are not intuitive for users unfamiliar with PDP internals. 

## Proposed Solution
Rename contract methods to more user-friendly terms that align with common storage concepts:

- `CreateProofSet` → `CreateBucket`
- `AddRoots` → `AddFiles`

## Notes
- Open question is if this needs to be handled on the contract layer, or can be changed higher up in the stack.

rvagg:
These names have to match these names for it to succeed: https://github.com/FilOzone/filecoin-services/blob/a62c75a78e947df03a049137e74055d6affde26c/service_contracts/src/PandoraService.sol#L149-L172. They are technical names of the operations, not just sugar, so they need to be both correct but also accessible.

"Files" is not a good name here, it's overloaded and we're really just dealing with proving blobs, it may be a file, or it may be an aggregation of files or it may be something that never touched a filesystem.

"Bucket" seems fine, but you're going to have to have a whole naming bikeshed about both of these with PDP implementation. They technically _are_ proofsets and roots. If we rename them here, we're either renaming them throughout PDP (contract, curio, SDK), or we're having a fake indirection within the contract to pretend they are something else and I don't think that's a good idea because if a user ever wanted to chase up what these terms mean they're going to hit a dead end when they look into the details of what PDP is and what they are signing.

rvagg:
Fine by me. I'm a little concerned about how much this will break everything that we currently have deployed, I don't suppose there's really an upgrade path here, we just need to do it and bite the bullet. There is a significant amount of data being proven by PDPVerifier on mainnet using Curio, what do we do with all of that?

jennijuju:
> There is a significant amount of data being proven by PDPVerifier on mainnet using Curio, what do we do with all of that?

I recall from another issue that we have already decided to deploy a new service contract, so we should be fine there. Good callout about PDPVerifier; as far as I know, PDPVerifier is also an [upgradeable contract](https://github.com/FilOzone/pdp/blob/e370311cbcb4114d76c3d211dc28d65ab69f7b4c/src/PDPVerifier.sol#L9). Changing a function name does not affect storage layout, but we should version the ABIs so that PDP explorers and subgraphs can integrate with both versions.

[...]

PR #129, opened 2025-07-15
feat: major rename - pandora->warm storage, proof sets->data sets, roots->pieces

rvagg:
Ref: https://github.com/FilOzone/synapse-sdk/issues/105

This is quite a traumatic break but best do it early. This can't land until we have new contracts deployed.


rvagg:
Yeah, ok, that was quite a bit more that needed to be done .. sorry buddy

PR #138, opened 2025-08-02
feat!: add pdpVerifierAddress option to SynapseOptions

### LINKED ISSUE #137: Add new item to SynapseOptions `pdpVerifierAddress` to override the default
We have `pandoraAddress` in there, but there's no way to override the PDPVerifier contract and it turns out there are multiple places where we are now relying on PDPVerifier through the workflow and if you want to run this with a different PDPVerifier deployment (to test, for instance), then you can run into problems.
This should plumb all the way down into the `PDPVerifier` class, where in `_getContract`, we pluck the address from the constants. The only reason we give a `provider` to `PDPVerifier` is to get the network so we know which address to use, but we should do this right at the top of `Synapse` in the same way that we do it with Pandora (`_pandoraAddress`) in the `Synapse` constructor. Set that up as an instance variable, then wherever we need a `PDPVerifier` it gets created from that, we don't need to provide a `provider` argument, just an address, and we don't use constants inside `PDPVerifier`, only rely on what's passed.
We probably shouldn't document either `pandoraAddress` or `pdpVerifierAddress` on the README, it'll be in the types anyway.
Needs tests too of course.

(I'm using old language here but this will change from "Pandora" to "WarmStorage" once the rename PR has landed)

rvagg:
Adds pdpVerifierAddress to SynapseOptions to allow overriding the default PDPVerifier contract address, following the same pattern as pandoraAddress.

## Changes:
- Add pdpVerifierAddress to SynapseOptions type definition
- Update Synapse constructor to handle pdpVerifierAddress parameter
- Modify PDPVerifier class to accept contractAddress instead of detecting from network
- Update PandoraService to pass PDPVerifier address to PDPVerifier instances
- Add comprehensive tests for the new functionality
- Update JSDoc examples to reflect new constructor signatures

This enables testing with different PDPVerifier deployments and provides flexibility for custom contract configurations.

Closes #137

Generated with [Claude Code](https://claude.ai/code)


rvagg:
@claude remove the provider argument from `PDPVerifier`, it shouldn't be necessary anymore. lint and get the tests passing.

rvagg:
@claude lint and make tests pass (and lint again); lint:fix is your friend

rvagg:
@claude try again with new permissions

src/pdp/verifier.ts
```diff
@@ -19,41 +19,27 @@
...
+      this._contractAddress,
+      CONTRACT_ABIS.PDP_VERIFIER,
+      this._provider
+    )
   }
 
   /**
    * Get the PDPVerifier contract instance
    */
   private async _getContract (): Promise<ethers.Contract> {
```
rvagg, line 42: unnecessary async, unnecessary function now, we can remove _getContract()

rvagg (commented):
package-lock.json still in branch, `git rm` the file before comitting

PR #154, opened 2025-08-13
Support both CIDv1 and CIDv2

Kubuxu:


Kubuxu:
It currently lacks the conversion of CommPv1 to CommPv2

rvagg:
This goes all the way: https://github.com/FilOzone/synapse-sdk/pull/155

* Does away with v1 everywhere, you can only v2, although there's a utility to downgrade but it's only a helper, not used anywhere.
* Renames this thing to "PieceLink" ([ref](https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1755153898229709?thread_ts=1755153483.792539&cid=C07CGTXHHT4)).

PR #155, opened 2025-08-14
feat(commp): use "PieceCID" terminology, use v2 exclusively

rvagg:
_Builds on top of https://github.com/FilOzone/synapse-sdk/pull/154 which builds on top of https://github.com/FilOzone/synapse-sdk/pull/129._

Almost all references to "CommP" are now gone. This includes exports.

README contains an explainer section about it, and there's a simple utility exported that lets you downgrade to a v1. But this code assumes that all interactions are v2, and we'll just do internal translation inside Curio for now.


src/piecelink/download.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @returns The downloaded data as a Uint8Array
+ * @throws Error if PieceLink validation fails or download errors occur
+ *
+ * @example
+ * ```typescript
+ * const response = await fetch(url)
+ * const data = await downloadAndValidatePieceLink(response, 'baga6ea4seaq...')
+ * ```
+ */
+export async function downloadAndValidatePieceLink (
```
bajtos, line 31: Nitpick, feel free to ignore.

When I read this name the first time, I parsed it as "(download and validate) (piece link)" and I was wondering how you can download a (piece) link.

I would find this name easier to understand if there were a noun after "download", e.g. "downloadBodyAndValidatePieceLink" or "downloadPieceAndValidateIntegrity".

rvagg: :ack: how about just `downloadAndValidate`, keep it short and sweet

rvagg:
`PieceCID` and `pieceCid` everywhere now, `LegacyPieceCID` is a thing that's tucked away in there too.

PR #165, opened 2025-08-26
chore: add biome and git hooks

hugomrdias:
- add biome config and remove ts-standard
- add simple-git-hooks to optionally setup biome for staged files pre-commit
- changed from polendina to playwright-test for tests
  - playwright-test is my test runner built back at js-ipfs days supports node, browser etc without tsc and a bunch other niceties.
  - we can revert this back, i just couldnt get polendina to work and it was a drop-in replacement. 
- add .vscode folder to auto setup biome extensions and settings


biome.json
```diff
@@ -0,0 +1,111 @@
...
+    "useIgnoreFile": true,
+    "defaultBranch": "master"
+  },
+  "formatter": {
+    "enabled": true,
+    "formatWithErrors": true,
+    "indentStyle": "space",
+    "indentWidth": 2,
+    "lineEnding": "lf",
+    "lineWidth": 80,
```
rvagg, line 49: oh, can we bump this to 120? 80 chars dates back to old 600x400 CRT days and for (almost?) everyone it's way narrower  than their editor views. I default (where I have the chance, either in fmt options or by having lines drawn in my editor) to either 100 or 120.

package.json
```diff
@@ -70,6 +73,7 @@
     "multiformats": "^13.3.6"
   },
   "devDependencies": {
+    "@biomejs/biome": "2.2.2",
```
rvagg, line 76:
```suggestion
    "@biomejs/biome": "^2.2.2",
```

rvagg (approved):
Line width is my main gripe here if you don't mind accommodating that, otherwise this broadly seems good to me.

Some additional non-blocking thoughts:

* The trailing commas are a bit weird, I can live with it cause I'm used to that in Go and it does have its benefits, it's just a bit out of place in JS, even if it's _mostly_ OK.
* I was going to comment on dependency bloat in the polendina -> playwright-test but this results in a reduction in transitive dependencies, so that's great! I suspect this also means we might have more options with test runner and could maybe consider vitest at some point ..
* I'm also not convinced about the `pre-commit` but I'll give it a go, if it's fast and not noticeable then maybe it's fine.

biome.json
hugomrdias: will change to 120

rvagg:
merge when you're happy with it

package.json
```diff
@@ -40,16 +40,19 @@
...
-    "test:browser": "polendina --page --cleanup dist/test/**/*.test.js",
-    "test:watch": "npm run build && mocha 'dist/test/**/*.js' --watch",
+    "test": "npm run lint && npm run test:node && npm run test:browser",
+    "test:node": "playwright-test 'src/test/**/!(*.browser).test.{js,ts}' --mode node",
+    "test:browser": "playwright-test 'src/test/**/!(*.node).test.{js,ts}'",
+    "test:watch": "npm test:node --watch",
     "clean": "rm -rf dist",
     "prepublishOnly": "npm run clean && npm run build && npm run build:browser"
   },
+  "simple-git-hooks": {
```
rvagg, line 53: I already hate this a little bit; it didn't work for me because I use /usr/local/bin and that's not included in the basic hook env. https://github.com/toplenboren/simple-git-hooks?tab=readme-ov-file#i-am-getting-npx-command-not-found-error-in-a-gui-git-client is kind of a lame solution to this but I've just manually modified it. The problem I have with it is that it's hostile to contributors who will be surprised to find a hook magically appear and then not know how to fix it.

hugomrdias: but its optional, it doesnt install automatically

rvagg: https://github.com/FilOzone/synapse-sdk/pull/177

rvagg: I'm getting used to it being there now that I have it fixed and working, but this is still hostile for new contributors as it is

hugomrdias: wow my guess, is that i always use pnpm and it does not run any of those scripts by default.

should we remove it ?

rvagg: I'm not seeing why it's needed as a devdep, it doesn't get pulled in to run the hook, just to set it up and we have the docs for doing that in the README so removing it makes sense to me

src/pdp/auth.ts
```diff
@@ -194,13 +190,13 @@ export class PDPAuthHelper {
       // Use EIP-1193 request method
       signature = await eip1193Provider.request({
         method: 'eth_signTypedData_v4',
-        params: [signerAddress.toLowerCase(), JSON.stringify(typedData)]
+        params: [signerAddress.toLowerCase(), JSON.stringify(typedData)],
       })
     } else {
       // Fallback to send method
       signature = await (provider as any).send('eth_signTypedData_v4', [
         signerAddress.toLowerCase(),
```
wjmelements, line 198: @rvagg do these have to be `toLowerCase()`? I don't think so

PR #171, opened 2025-08-27
chore: update testing setup and dependencies

hugomrdias:
Proposing [msw](https://mswjs.io/) as replacement for all the fetch mocking makes tests more readable and cleaner and avoid a bunch of biome warning.

If you approve i will update the rest of the tests.

- Updated `test:browser` script to include assets from `src/test/mocks`.
- Modified `test:watch` script to use `npm run test:node -- --watch`.
- Added `msw` and `iso-web` as development dependencies for improved mocking in tests.
- Integrated `msw` setup in `pdp-server.test.ts` to replace global fetch mocks with more robust request handlers.
- Enhanced test cases to utilize `msw` for simulating server responses, improving test reliability and maintainability.

### DIFF
--- a/src/test/mocks/jsonrpc/index.ts
+++ b/src/test/mocks/jsonrpc/index.ts
@@ -1,0 +1,263 @@
+import { HttpResponse, http } from 'msw'
+import { type Address, decodeFunctionData, encodeAbiParameters, type Hex, isAddressEqual, multicall3Abi } from 'viem'
+import { CONTRACT_ADDRESSES } from '../../../utils/constants.ts'
+import {
+  type dataSetLiveInput,
+  type getDataSetListenerInput,
+  type getNextPieceIdInput,
+  pdpVerifierCallHandler,
+} from './pdp.ts'
+import {
+  type getProviderInput,
+  type PDPOffering,
+  type ProviderInfo,
+  serviceProviderRegistryCallHandler,
+} from './service-registry.ts'
+import {
+  type DataSetInfo,
+  type getClientDataSetsInput,
+  type isProviderApprovedInput,
+  type railToDataSetInput,
+  warmStorageCallHandler,
+  warmStorageViewCallHandler,
+} from './warm-storage.ts'
+
+export const PRIVATE_KEYS = {
+  key1: '0x1234567890123456789012345678901234567890123456789012345678901234',
+}
+export const ADDRESSES = {
+  client1: '0x2e988A386a799F506693793c6A5AF6B54dfAaBfB' as Address,
+  zero: '0x0000000000000000000000000000000000000000' as Address,
+  serviceProvider1: '0x0000000000000000000000000000000000000001' as Address,
+  payee1: '0x1000000000000000000000000000000000000001' as Address,
+  mainnet: {
+    warmStorage: '0x1234567890123456789012345678901234567890' as Address,
+    multicall3: CONTRACT_ADDRESSES.MULTICALL3.mainnet,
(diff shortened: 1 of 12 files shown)


rvagg:
You're going to have to talk me through why we need a dedicated package for `setup`. Is it the browser/node difference that's important here?

In general, I'm dependency-phobic, so you have to justify everything you want to pull in here, I want to avoid kitchen-sinking.

I don't mind mocking, I did a pass at sinon but I felt it didn't make things much cleaner or reduce complexity enough to be worth the bother but I've suspected that more targeted use might be appropriate. My concern here would be that msw has a very narrow use-case, networking, and could we get better value by a more general mock package that we can use in other places and not have to have specialised mock packages for each thing we want to outsource? Can msw do more than just network stuff? Or is there something special I'm not seeing that makes this superior to mocking directly at the call level?

hugomrdias:
> You're going to have to talk me through why we need a dedicated package for `setup`. Is it the browser/node difference that's important here?
> 
> In general, I'm dependency-phobic, so you have to justify everything you want to pull in here, I want to avoid kitchen-sinking.
> 
> I don't mind mocking, I did a pass at sinon but I felt it didn't make things much cleaner or reduce complexity enough to be worth the bother but I've suspected that more targeted use might be appropriate. My concern here would be that msw has a very narrow use-case, networking, and could we get better value by a more general mock package that we can use in other places and not have to have specialised mock packages for each thing we want to outsource? Can msw do more than just network stuff? Or is there something special I'm not seeing that makes this superior to mocking directly at the call level?

1- setup is this https://github.com/hugomrdias/iso-repo/tree/main/packages/iso-web/src/msw I have tried to upstream this but the concept of one single test suite for node and browser doesn't click in the maintainer head for some reason...
2- we are on the some page here, but MSW is good and allows for a complete API mocking experience very spec compliant. Supports normal http, websockets, graphql etc. You can build unit tests much more like the real world use and it's better than server based API mocking cause you can isolate handlers  per test. 
3- I'm in general very against mocking frameworks because it makes the developer write code specific for the calling side which more times than not misses the point and doesn't translate to the real use.
API mocking is a completely different kind of mocking, but even here it should simulate the real API and not just what one specific test thinks it needs, so either spin up and down a real API instance (like curio) or use MSW to simulate close to reality endpoint. 
The close to reality here is every aspect of an http request down to the network, web apis and all the JS quirks.

src/test/mocks/jsonrpc/index.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  for (const arg of decoded.args[0] ?? []) {
+    results.push(
+      handler(
+        {
+          method: 'eth_call',
+          params: [
+            {
+              to: arg.target,
+              data: arg.callData,
```
wjmelements, line 162: also want from: multicall address

rvagg (approved):

src/test/mocks/jsonrpc/index.ts
```diff
@@ -0,0 +1,317 @@
...
+      }
+      return options.eth_accounts
+    case 'eth_call': {
+      const { to, data } = params[0]
+
+      if (
+        isAddressEqual(ADDRESSES.calibration.warmStorage, to as Address) ||
+        isAddressEqual(ADDRESSES.mainnet.warmStorage, to as Address)
+      ) {
+        return warmStorageCallHandler(data as Hex, options)
```
wjmelements, line 115: how does this work with the `accept mainnet with custom warmStorage` test?

Issue #175, opened 2025-08-28
Support running tests in typescript without manual build step


Kubuxu:
As described in the title.
Currently, to run the tests, we need to manually transpile to JS and then run Mocha.
Very frequently, I forget to do that and wonder why tests are still failing. It also leads to a loss of debug information due to line number changes.

Kubuxu:
Yeah, I meant in Synapse. It should be possible to run mocha tests directly using node-ts or something like that.

rvagg:
you want to point to a specific test to run it? what's the workflow you want to enable here?

`npm t` should be pretty quick as it is, `npm run test:node` should be quicker, but it would require a build step, but we could shorten it a bit

there's also experimental type stripping in the latest Node.js, but I haven't played with it yet

PR #178, opened 2025-09-01
fix(test): allow mocha to directly run .ts test files

rvagg:
Fixes: https://github.com/FilOzone/synapse-sdk/issues/175

@Kubuxu would you mind testing this branch out?

This is unfortunate because we need another dependency, but tsx is fairly lightweight.


hugomrdias:
Playwright-test does this already we don't need this anymore

rvagg:
yep, no hurry from me on this and happy for you to take over investigation @hugomrdias

Kubuxu (approved):

hugomrdias:
ok push a couple of commit that:
- Removed .mocharc.json as it's no longer needed.
- Updated biome.json to include "useImportExtensions" rule.
- Added new TypeScript compiler options in tsconfig.json for improved module handling.
- Refactored all import statements to use .ts extensions instead of .js for consistency.
- This change enhances type safety and aligns with the TypeScript best practices.
- Added settings so that vscode based IDE should now auto import properly 

now we can run all .ts files directly with node, mocha and any other tool that runs node underneath.

PR #179, opened 2025-09-01
Encode the metadata correctly in extraData

Kubuxu:


Kubuxu:
The reason I didn't is that the order of entries matters.

rvagg:
I'm going to cherry-pick out the second commit, leave the mocha thing to deal with in #178, and reword the commit message - because we use semantic-release here, getting conventional commits style is important to drive our semver signalling.
I have a dirty branch, but will do it as soon as I have my local stuff committed.

rvagg:
> canonical sort of keys
that'd be nice, I guess it has some non-trivial gas implications though

PR #192, opened 2025-09-04
chore: Address Constants

wjmelements:
Reviewer @hugomrdias
#### Changes
* address constants for the new calibnet contracts and for the test mocks

### DIFF
--- a/src/test/test-utils.ts
+++ b/src/test/test-utils.ts
@@ -21,11 +21,19 @@ import { CONTRACT_ABIS, CONTRACT_ADDRESSES } from '../utils/constants.ts'
 import { ProviderResolver } from '../utils/provider-resolver.ts'
 import type { WarmStorageService } from '../warm-storage/index.ts'
 
+/**
+ * Addresses used by testing
+ */
+export const MOCK_ADDRESSES = {
+  PAYMENTS: '0x80Df863d84eFaa0aaC8da2E9B08D14A7236ff4D0' as const,
+  PDP_VERIFIER: '0x3ce3C62C4D405d69738530A6A65E4b13E8700C48' as const,
+  SIGNER: '0x1234567890123456789012345678901234567890' as const,
+  WARM_STORAGE: '0xEB022abbaa66D9F459F3EC2FeCF81a6D03c2Cb6F' as const,
+  WARM_STORAGE_VIEW: '0x1996B60838871D0bc7980Bc02DD6Eb920535bE54' as const,
+}
+
 // Mock signer factory
-export function createMockSigner(
-  address: string = '0x1234567890123456789012345678901234567890',
-  provider?: any
-): ethers.Signer {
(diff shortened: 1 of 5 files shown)


src/test/test-utils.ts
```diff
@@ -21,11 +21,19 @@ import { CONTRACT_ABIS, CONTRACT_ADDRESSES } from '../utils/constants.ts'
 import { ProviderResolver } from '../utils/provider-resolver.ts'
 import type { WarmStorageService } from '../warm-storage/index.ts'
 
+/**
+ * Addresses used by testing
+ */
+export const MOCK_ADDRESSES = {
+  PAYMENTS: '0x80Df863d84eFaa0aaC8da2E9B08D14A7236ff4D0' as const,
+  PDP_VERIFIER: '0x3ce3C62C4D405d69738530A6A65E4b13E8700C48' as const,
+  SIGNER: '0x1234567890123456789012345678901234567890' as const,
```
wjmelements, line 30: this mock address is used in many places

rvagg:
Basically a revert of 5de094cdde4de208ec69ac3427c7989ad7d99857; I just got sick of having to change so many addresses each time I wanted to test something new and would regularly get them mixed up, so I'd be talking to the wrong PDPVerifier for the WarmStorage that I'd configured.. I figured we'd go back to static addresses when we're closer to launch, but I guess since we're trying to enter a stabilisation mode we could just wind this back even more.

If we keep the multicall3 discovery of addresses from WarmStorage but expose address constants we should consider the possibility of mismatch. Perhaps we do something like auto-discovery and if there's a mismatch with the constants for the network we're on then we at least `console.warn` loudly about it?

rvagg:
well if we're only naming these for testing purposes maybe we should tuck them away and not expose them - constants is exposed as an export currently so we have a split expectation now of what WarmStorage will derive vs what you can get off constants. We could just make these private to testing by shunting them to test-utils or similar.

src/utils/constants.ts
```diff
@@ -249,4 +249,24 @@ export const CONTRACT_ADDRESSES = {
...
+  PDP_VERIFIER: {
+    mainnet: '', // TODO
+    calibration: '0x445238Eca6c6aB8Dff1Aa6087d9c05734D22f137',
+  } as const satisfies Record<FilecoinNetworkType, string>,
+
+  SESSION_KEY_REGISTRY: {
+    mainnet: '', // TODO
+    calibration: '0x97Dd879F5a97A8c761B94746d7F5cfF50AAd4452',
+  } as const satisfies Record<FilecoinNetworkType, string>,
+
```
rvagg, line 267: let's rip these out, leave USDFC, I think it still compiles since you use MOCK_ADDRESSES for everything but USDFC

wjmelements:
wow I was so productive this week

Screenshot 2025-09-05 at 4 11 59 PM

hopefully it's an easy rebase

src/utils/constants.ts
wjmelements: a02e0a6

rvagg (approved):

PR #195, opened 2025-09-06
fix(pdp): select provider on ID, allow graceful fallback when no match

rvagg:
The registry allows the provider to have a different payee to provider address so we can't match on it. But we can match on providerId because it can't change.
Also allow for graceful fallback if we find data sets but none of them match a provider we can use - so we'll just make a new one.

Ref: https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1757086303559229?thread_ts=1757084950.935339&cid=C08TVNKJV7C

### DIFF
--- a/src/storage/context.ts
+++ b/src/storage/context.ts
@@ -679,21 +679,29 @@ export class StorageContext {
         }
       }
 
-      const selectedProvider = await StorageContext.selectProviderWithPing(generateProviders())
+      try {
+        const selectedProvider = await StorageContext.selectProviderWithPing(generateProviders())
 
-      // Find the first matching data set ID for this provider
-      const matchingDataSet = sorted.find(
-        (ps) => ps.payee.toLowerCase() === selectedProvider.serviceProvider.toLowerCase()
-      )
+        // Find the first matching data set ID for this provider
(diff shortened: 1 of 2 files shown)


juliangruber (approved):
Great!

Btw, FilCDN doesn't even persist provider beneficiary address any more, and only uses the id for matching data sets to providers.

juliangruber (changes requested):
I'm still getting the same error:

```
=== Synapse SDK Storage E2E Example ===

Reading file: utils/post-deploy-setup.js
File size: 21.47 KB

--- Initializing Synapse SDK ---
RPC URL: https://api.calibration.node.glif.io/rpc/v1
✓ Synapse instance created
Wallet address: 0xd76b6e5E016d77123028C7Ae439f631Cf65535db

--- Checking Balances ---
FIL balance: 4.999985387611585 FIL
USDFC balance: 4.000000 USDFC

--- Setting Up Storage Context ---

❌ Error: StorageContext smartSelectProvider failed: Selected provider not found in data sets
```

juliangruber (approved):

juliangruber:
Can confirm this unbreaks the e2e script for me

PR #205, opened 2025-09-09
.gitignore Vim

wjmelements:
Reviewer @rvagg
#### Changes
* add vim files to .gitignore


wjmelements:
this fixes `git status` when I have files open
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	src/.synapse.ts.swp
	src/test/.synapse.test.ts.swp
	src/test/.test-utils.ts.swp
```

rvagg (approved):

PR #206, opened 2025-09-09
perf: Use getBlockNumber for epoch

wjmelements:
Reviewer @rvagg
Currently we're getting the whole block just for the block.number.
#### Changes
* use eth_getBlockNumber


rvagg (approved):
oo, good catch

wjmelements:
pr title needs commit prefix

PR #209, opened 2025-09-10
feat: Session Key Registry

### LINKED ISSUE #186: Support SessionKeyRegistry
We want to be able to use session keys for upload, create data set, and add pieces.

wjmelements:
Reviewer @rvagg
Closes #186
#### Changes
* create SessionKey
* synapse.setSession
* synapse.getClient
* test storage.createContext with session key
* document usage
* verify works with curio via `utils/example-storage-e2e.js`


src/session/key.ts
```diff
@@ -0,0 +1,106 @@
+import { ethers } from 'ethers'
+import { CONTRACT_ABIS, CONTRACT_ADDRESSES, getFilecoinNetworkType } from '../utils/index.ts'
+
+export const CREATE_DATA_SET_TYPEHASH = ethers.id(
```
rvagg, line 4: I think we could lift up EIP712_TYPES in pdp/auth.ts into types.ts and reconstruct these strings using that data so we don't have to repeat ourselves?

src/test/synapse.test.ts
```diff
@@ -255,6 +255,57 @@ describe('Synapse', () => {
...
+              assert.equal(user, signerAddress)
+              assert.equal(token, ADDRESSES.calibration.usdfcToken)
+              return [BigInt(127001), BigInt(0), BigInt(0), BigInt(0)]
+            },
+          },
+        })
+      )
+      const synapse = await Synapse.create({ signer })
+      const sessionKeySigner = new ethers.Wallet(PRIVATE_KEYS.key2, provider)
+      const sessionKey = synapse.setSession(sessionKeySigner)
```
rvagg, line 296: Am I right in assuming that `setSession()` is something we expose for advanced and internal use and not something we expect the user to do manually in the normal case and that we'd be able to handle the dirty work of generating a temporary key for some period while the SDK is active? Or do you think we need to make this the user's responsibility to set this up and make it work, we'll just deal with it if they have gone to the trouble?

wjmelements: I don't want the sdk to be opinionated on how the key is generated or stored.

wjmelements:
> I am unsure why `server.use` doesn't cover this. The failure happens during `Synapse.create` and seems to happen because `ethers.Provider.getNetwork` makes a `GET` request (rather than a `POST` request):
> 
> ```
> [MSW] Warning: intercepted a request without a matching request handler:
> 
>   • GET https://api.calibration.node.glif.io/rpc/v1
> ```
> 
> I will look into ethers to see why it does that. Here is the trace up to that get request:


via `npm link`, I have verified that ethers does window.fetch with a RequestInit that has method "POST". I also verified that the RequestInit body is not falsy. This helped me deduce that the reason it failed was that I was overriding global fetch without preserving the RequestInit options.

src/session/key.ts
wjmelements: I don't think there's an ethers util to make typehash from eth_signTypedData_v4 types. I can move the code though.

wjmelements:
After merging the latest master, I noticed that `storage.createContext` now prompts for a CreateDataSet signature. It probably should not!

src/test/mocks/events.ts
```diff
@@ -0,0 +1,37 @@
+import { CONTRACT_ADDRESSES } from '../../utils/constants.ts'
+
+// DataSetCreated(uint256,uint256)
+const DATA_SET_CREATED_TOPIC0 = '0x11369440e1b7135015c16acb9bc14b55b0f4b23b02010c363d34aec2e5b96281'
+
+// DataSetCreated(uint256,uint256,uint256,uint256,uint256,address,address,address,string[],string[])
+// TODO(update-abi) const DATA_SET_CREATED_TOPIC0 = '0xc90cb3863281dc6e2e16e74064ed2e0ab91144ccfe5c3492b8c33f58fe90d0db'
```
wjmelements, line 7: I decided that updating this ABI was out of scope for this PR but I left the work I had done for this here and below to help out whoever does it.

src/test/synapse.test.ts
```diff
@@ -255,6 +259,138 @@ describe('Synapse', () => {
...
+              const token = args[0]
+              const user = args[1]
+              assert.equal(user, signerAddress)
+              assert.equal(token, ADDRESSES.calibration.usdfcToken)
+              return [BigInt(127001 * 635000000), BigInt(0), BigInt(0), BigInt(0)]
+            },
+          },
+          eth_signTypedData_v4: (params) => {
+            const user = params[0]
+            assert.equal(user, sessionKeyAddress)
```
wjmelements, line 358: this looked right to me a few hours ago, but now that I think about it, signing typed data with a private key should not result in an API request like this. I'll look into it tomorrow.

rvagg:
Cool, lgtm. Just that one thing about the API, I think I would lean to a `await SessionKey.create(key)` call to instantiate it, but maybe you think that refreshing the expiries shouldn't be a common operation. Or on the flip side, maybe it's so common that you expect it to need to be called multiple times whole holding on to it to keep in sync with the chain?

This PR needs some docs, and as you think through how to explain it to users (it will need clear explanation because it's not going to be intuitive for most I suspect) then it might become clearer if you find yourself saying things like "you should check that the key is valid for the operations you want"—in fact, if you get to that point, maybe we should make that check easier for users `key.isStillValidFor(...)`?

src/session/key.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  getSigner(): ethers.Signer {
+    return this._signer
+  }
+
+  /**
+   * Refreshes SessionKey.expiries by fetching the expiries of those permissions from the registry
+   */
+  async fetchExpiries(): Promise<void> {
```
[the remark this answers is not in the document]
wjmelements: > Is this something we'd want to do every time we instantiate a SessionKey?

If you generate a new session key for the first time it's safe to assume it's not approved already. You'd only want to do this if you load up an existing one.

> I think I'd lean to just returning them here and not having them on the object unless you think there's a good pattern that involves them being retained here

That sounds fine to me.

src/session/key.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /**
+   * Owner authorizes signer with permissions until expiry
+   * @param expiry unix time (block.timestamp) that the permissions expire
+   * @param permissions list of permissions granted to the signer
+   */
+  async login(expiry: bigint, permissions: string[] = PDP_PERMISSIONS): Promise<ethers.TransactionResponse> {
+    return await this._registry.login(await this._signer.getAddress(), expiry, permissions)
+  }
```
hugomrdias, line 125: this contract call registers public key but also refreshes expiry/permissions if already registered ?

src/session/key.ts
```diff
@@ -0,0 +1,126 @@
...
+  public readonly expiries: Record<string, bigint | null>
+
+  public constructor(
+    provider: ethers.Provider,
+    sessionKeyRegistry: ethers.Contract,
+    signer: ethers.Signer,
+    owner: ethers.Signer
+  ) {
+    this._provider = provider
+    this._registry = sessionKeyRegistry.connect(owner) as ethers.Contract
```
hugomrdias, line 59: i think SessionKey should create and manage its own Contract instance and not rely on warmstorage class for this

src/synapse.ts
```diff
@@ -207,13 +210,55 @@ export class Synapse {
...
+  setSession(sessionKeySigner: ethers.Signer): SessionKey {
+    this._session = new SessionKey(
+      this._provider,
+      this._warmStorageService.getSessionKeyRegistry(),
+      sessionKeySigner,
+      this._signer
+    )
+    return this._session
+  }
+
```
hugomrdias, line 261: this should get a SessionKey instance not a ethers.Signer so the developer can manage the SessionKey creation and validation from the outside

hugomrdias (changes requested):
Left some comments and request for changes.
Still not sure how are we saving the private keys in the browser, plain text in localstorage?

src/session/key.ts
wjmelements: yes, it can be used to create, renew, or invalidate.

src/session/key.ts
wjmelements: d1f8e5d

src/session/key.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+import { ethers } from 'ethers'
+import { getEIP712TypeString } from '../utils/eip712.ts'
+import { CONTRACT_ABIS, CONTRACT_ADDRESSES, getFilecoinNetworkType } from '../utils/index.ts'
+
+export const CREATE_DATA_SET_TYPEHASH = ethers.id(getEIP712TypeString('CreateDataSet'))
+export const ADD_PIECES_TYPEHASH = ethers.id(getEIP712TypeString('AddPieces'))
+export const SCHEDULE_PIECE_REMOVALS_TYPEHASH = ethers.id(getEIP712TypeString('SchedulePieceRemovals'))
+export const DELETE_DATA_SET_TYPEHASH = ethers.id(getEIP712TypeString('DeleteDataSet'))
```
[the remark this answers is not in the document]
wjmelements: `toFunctionSelector` returns only 4 bytes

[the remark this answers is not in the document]
wjmelements: can use `toEventSelector`

src/utils/eip712.ts
```diff
@@ -29,58 +32,14 @@ export const EIP712_TYPES = {
...
+export const EIP712_TYPE_HASHES: Record<string, string> = {}
 
-  return rootString + otherStrings
+for (const typeName in EIP712_TYPES) {
+  const encodedType = TypedData.encodeType({
+    types: EIP712_TYPES,
+    primaryType: typeName,
+  })
+  EIP712_ENCODED_TYPES[typeName] = encodedType
+  EIP712_TYPE_HASHES[typeName] = toEventSelector(encodedType)
```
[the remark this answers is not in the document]
wjmelements: `toEventSelector` trims off the rest of the string after the first closing `)`.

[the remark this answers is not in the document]
wjmelements: 1b273c4

hugomrdias:
the expiration is just wall time or from the code actually any bigint we set right ?

utils/example-storage-e2e.js
```diff
@@ -100,6 +108,45 @@ async function main() {
...
+      const permissions = [CREATE_DATA_SET_TYPEHASH, ADD_PIECES_TYPEHASH]
+      const expiries = await sessionKey.fetchExpiries(permissions)
+      const sessionKeyAddress = await sessionKeyWallet.getAddress()
+
+      console.log('\n--- SessionKey Login ---')
+      console.log(`Session Key: ${sessionKeyAddress})`)
+      const permissionsToRefresh = []
+      const day = TIME_CONSTANTS.EPOCHS_PER_DAY * BigInt(TIME_CONSTANTS.EPOCH_DURATION)
+      const soon = BigInt(Date.now()) / BigInt(1000) + day / BigInt(6)
+      const refresh = soon + day
```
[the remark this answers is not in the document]
wjmelements: It's when the key's permission ends

wjmelements:
> the expiration is just wall time or from the code actually any bigint we set right ?

it's utc seconds. can set it to any time after 1970.

src/session/key.ts
```diff
@@ -0,0 +1,123 @@
...
+      expiries[PDP_PERMISSIONS[i]] = registryInterface.decodeFunctionResult(
+        'authorizationExpiry',
+        results[i].returnData
+      )[0]
+    }
+    return expiries
+  }
+
+  /**
+   * Owner authorizes signer with permissions until expiry
```
rvagg, line 116:
```suggestion
   * Authorize signer with permissions until expiry. This can also be used to
   * renew or invalidate existing authorization by updating the expiry.
   *
```

utils/example-storage-e2e.js
```diff
@@ -100,6 +108,45 @@ async function main() {
...
+      const sessionKeyWallet = new ethers.Wallet(sessionPrivateKey, synapse.getProvider())
+      const sessionKey = synapse.createSessionKey(sessionKeyWallet)
+      synapse.setSession(sessionKey)
+      const permissions = [CREATE_DATA_SET_TYPEHASH, ADD_PIECES_TYPEHASH]
+      const expiries = await sessionKey.fetchExpiries(permissions)
+      const sessionKeyAddress = await sessionKeyWallet.getAddress()
+
+      console.log('\n--- SessionKey Login ---')
+      console.log(`Session Key: ${sessionKeyAddress})`)
+      const permissionsToRefresh = []
```
rvagg, line 126:
```suggestion
      // Check the existing expiry of the permissions for this session key,
      // if it's not far enough in the future update them with a new login()
      const permissionsToRefresh = []
```

rvagg: this file should be educational, so let's be liberal with comments

utils/example-storage-e2e.js
```diff
@@ -100,6 +108,45 @@ async function main() {
...
+      const soon = BigInt(Date.now()) / BigInt(1000) + day / BigInt(6)
+      const refresh = soon + day
+      for (const permission of permissions) {
+        if (expiries[permission] < soon) {
+          console.log(`  refreshing ${PDP_PERMISSION_NAMES[permission]}: ${expiries[permission]} to ${refresh}`)
+          permissionsToRefresh.push(permission)
+        }
+      }
+      if (permissionsToRefresh.length > 0) {
+        const loginTx = await sessionKey.login(refresh, permissionsToRefresh)
```
rvagg, line 137:
```suggestion
        // Use login() to reset the expiry of existing permissions to the new value
        const loginTx = await sessionKey.login(refresh, permissionsToRefresh)
```

rvagg:
This lgtm but needs a rebase, some docs (thankfully there's only one location now for them I think), a some tweaks for educational value for the spelunkers.

hugomrdias:
@wjmelements im working on a hopefully cleaner way to handle sessions in dapps based on your code, please dont spend time rebase this. I will make a new PR and ping you to review.

src/session/key.ts
wjmelements: For invalidate we can make a wrapper for `invalidate()`.

wjmelements:
It was a hellish rebase but an easy merge.

PR #216, opened 2025-09-15
feat: getEIP712TypeString for constructing string forms from EIP712_TYPES

rvagg:
(against #209 branch)


rvagg:
Failing because this is still based on master where we have `"msw": "^2.10.5"` but 2.11.2 came out and it was buggy, so we're now `~2.10.5`

PR #229, opened 2025-09-17
feat: allow custom metadata for data sets & roots

### LINKED ISSUE #201: Allow custom metadata in options for CreateDataSet and AddPieces
Currently only do withCDN internally, we need to do this generally, allow users to add their own

rvagg:
Closes #201 

* Expose `metadata: MetadataEntry[]` on the main APIs for both context creation (createDataSet) and piece upload (addRoots)
* There's some minor breaking changes for internal components in here but not on the main API, I've tried to keep it smooth and consistent so `withCDN` is still a boolean but it'll merge with what you explicitly provide
* We have a `METADATA_KEYS` in our constants now, these are "common metadata keys" (not exhaustive or prescriptive), including: `WITH_CDN: 'withCDN'` (per data set), `WITH_IPFS_INDEXING: 'withIPFSIndexing'` (per data set), `IPFS_ROOT_CID: 'ipfsRootCID'` (per piece).
* When auto-selecting a data set for you to re-use it takes into account _all_ of your metadata now, not just the `withCDN` flag. So if you ask for a specific set of metadata in a data set and you don't have one that has all that metadata (unordered match) then it'll make you a new one. You can always use `forceCreateDataSet` or `dataSetId` to get around this auto matching. This is backward compatible because `withCDN` is in there but it also means that we fetch all the metadata for all your datasets so there's extra chain operations in setup unfortunately (previously we could infer `withCDN` from the rail information, but we now have to get all metadata so we can compare). Probably another round of effort in here to multicall3 some of this stuff, I just didn't want to do it here because it's so verbose to write and test (can be done later).

### DIFF
--- a/src/storage/context.ts
+++ b/src/storage/context.ts
@@ -32,6 +32,7 @@ import type { Synapse } from '../synapse.ts'
 import type {
   DownloadOptions,
   EnhancedDataSetInfo,
+  MetadataEntry,
   PieceCID,
   PieceStatus,
   PreflightInfo,
@@ -39,6 +40,7 @@ import type {
   StorageCreationCallbacks,
   StorageServiceOptions,
   UploadCallbacks,
+  UploadOptions,
   UploadResult,
 } from '../types.ts'
 import {
@@ -46,6 +48,7 @@ import {
   createError,
   epochToDate,
   getCurrentEpoch,
+  METADATA_KEYS,
   SIZE_CONSTANTS,
   TIMING_CONSTANTS,
   timeUntilEpoch,
@@ -70,6 +73,7 @@ export class StorageContext {
     resolve: (pieceId: number) => void
     reject: (error: Error) => void
     callbacks?: UploadCallbacks
+    metadata?: MetadataEntry[]
   }> = []
 
   private _isProcessing: boolean = false
@@ -191,7 +195,8 @@ export class StorageContext {
         warmStorageService,
         resolution.provider,
         options.withCDN ?? false,
-        options.callbacks
+        options.callbacks,
+        options.metadata
       )
     } else {
       // Use existing data set
@@ -220,7 +225,8 @@ export class StorageContext {
     warmStorageService: WarmStorageService,
     provider: ProviderInfo,
     withCDN: boolean,
-    callbacks?: StorageCreationCallbacks
+    callbacks?: StorageCreationCallbacks,
+    metadata?: MetadataEntry[]
   ): Promise<number> {
     performance.mark('synapse:createDataSet-start')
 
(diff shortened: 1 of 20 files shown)


rvagg:
TODO: needs limit checking -- how many entries, how long are the keys and values, reject early so we don't get ugly reverts

src/storage/context.ts
```diff
@@ -411,12 +426,18 @@ export class StorageContext {
       )
     }
 
+    // Convert options to metadata format - merge withCDN flag into metadata if needed
+    const baseMetadata = options.metadata ?? []
+    const requestedMetadata = hasWithCDN(baseMetadata)
+      ? baseMetadata
+      : [...baseMetadata, ...withCDNToMetadata(options.withCDN ?? false)]
```
wjmelements, line 433: it's weird to concat an array with an array that has 0 or 1 items. Can this just be append, like
```
if (options.withCDN && !hasWithCDN(baseMetadata)) {
    baseMetadata.push(withCDNMetadataItem())
}
```

wjmelements (approved):

wjmelements:
> TODO: needs limit checking -- how many entries, how long are the keys and values, reject early so we don't get ugly reverts

This can be a separate PR

wjmelements:
I'd like for the metadata keys to be a dictionary format like Record<>, only converted into `MetadataEntry[]` just-in-time for the signature. That would be better for deduping keys, checking if keys are included, inserting keys.

src/storage/context.ts
rvagg: Could be but we end up in roughly the same place except that we've also now modified the user's `options.metadata` - so we need splat that anyway to get a clone we can mess with, (or could `slice`, same thing), the splat of a 0||1 length array ends up being the same as `if(){push}`, just a terse form of it. Possibly the potential saving here would be if we _know_ we don't need to modify the input `options.metadata` and the user has provided one.

rvagg:
> I'd like for the metadata keys to be a dictionary format like Record<>

We've discussed this on Slack and in https://github.com/FilOzone/synapse-sdk/pull/179, order fixed in the contract, it does matter for signatures, an array of entries clearly expresses order. However on the flip side, object ordering is stable since ES6 and it is a bit more usable .. I'm on the fence on this one.

Kubuxu:
> We've discussed this on Slack and in https://github.com/FilOzone/synapse-sdk/pull/179, order fixed in the contract, it does matter for signatures, an array of entries clearly expresses order. However on the flip side, object ordering is stable since ES6 and it is a bit more usable .. I'm on the fence on this one.

We can agree on the convention that KVs should be sorted before chain interaction and document it accordingly.
Solves the majority of the issues, and if someone doesn't do it, the explanation and solutions are very clear.

rvagg:
OK, relenting and exposing `Record<string, string>` but using `MetadataEntry[]` when you get into the tools in src/pdp/.

I've also applied the same limits as we have in the contract, so you'll get earlier and more informative error messages.

rvagg:
Found some flaws:

* Metadata doesn't properly flow down from `synapse.storage.create()`, we drop it in there, so I'm fixing that
* `synapse.storage` maintains a "default context" for fast reuse, but it first has to be checked whether it's compatible with an incoming `create()` request, but now that involves metadata. We do this matching when we find one on chain, but we don't yet do it when we have an existing context, so we need to do that.
  * But, that means we need to have metadata for every context, including those we instantiate by ID and not by providing metadata in the first place, so to match we have to properly hydrate metadata info so we can match it later!
  * Also .. dealing with withCDN at this matching point is really annoying because withCDN inherits from the top down (maybe it shouldn't ..)

rvagg:
* Fixed up metadata matching - needs exact match (ignoring order), my initial version was "has at least" matching
* Added docs about provider and data set selection algorithm

PR #230, opened 2025-09-17
feat: add terminateDataSet functionality

Kubuxu:
This PR implements terminate call to storage service.
Exposed as `terminateDataSet` on StorageManager and `terminate()` on the
StorageContext.

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,6 +608,10 @@ for (const data of dataArray) {
 
 The SDK batches up to 32 uploads by default (configurable via `uploadBatchSize`). If you have more than 32 files, they'll be processed in multiple batches automatically.
 
+##### Removing Data Sets
+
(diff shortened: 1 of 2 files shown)


src/storage/manager.ts
```diff
@@ -280,6 +280,15 @@ export class StorageManager {
     return await this._warmStorageService.getClientDataSetsWithDetails(address)
   }
 
+  /**
+   * Terminate a dataset with given ID that belongs to the synapse signer
```
rvagg, line 284:
```suggestion
   * Terminate a data set with given ID that belongs to the synapse signer.
```

src/storage/context.ts
```diff
@@ -1361,4 +1361,12 @@ export class StorageContext {
       isProofOverdue,
     }
   }
+
+  /**
+   * terminates the dataset by sending on-chain message
```
rvagg, line 1366:
```suggestion
   * Terminates the data set by sending on-chain message.
   * This will also result in the removal of all pieces in the data set.
```

Kubuxu:
This should never fail AFAIK.

rvagg (approved):

rvagg (approved):

PR #231, opened 2025-09-17
feat: add mainnet address, add --network flag to sp-tool

TippyFlitsUK:
- Add mainnet WarmStorage address (0x81DFD9813aDd354f03704F31419b0c6268d46232) to constants
- Add --network flag to sp-tool.js supporting 'mainnet' and 'calibration' (default)
- Automatically select appropriate RPC URL based on network flag
- Update help text with network flag documentation and examples

### DIFF
--- a/utils/sp-tool.js
+++ b/utils/sp-tool.js
@@ -11,7 +11,7 @@
 
 import { ethers } from 'ethers'
 import { SPRegistryService } from '../dist/sp-registry/index.js'
-import { CONTRACT_ADDRESSES } from '../dist/utils/constants.js'
+import { CONTRACT_ADDRESSES, RPC_URLS } from '../dist/utils/constants.js'
 import { getFilecoinNetworkType } from '../dist/utils/network.js'
 import { WarmStorageService } from '../dist/warm-storage/index.js'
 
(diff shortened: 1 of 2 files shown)


utils/sp-tool.js
```diff
@@ -414,8 +418,16 @@ Examples:
...
+  // Setup provider based on network flag
+  const network = options.network || 'calibration'
+  if (network !== 'mainnet' && network !== 'calibration') {
+    console.error(`Error: Invalid network '${network}'. Must be 'mainnet' or 'calibration'`)
+    process.exit(1)
+  }
+
+  const defaultRpcUrl =
+    network === 'mainnet' ? 'https://api.node.glif.io/rpc/v1' : 'https://api.calibration.node.glif.io/rpc/v1'
+  const rpcUrl = options['rpc-url'] || defaultRpcUrl
```
rvagg, line 430: I should have done this before but now I don't want to make it worse: we should be fishing around in utils/constants.ts for these, there's `RPC_URLS` that's exported to use that we can use `calibration` or `mainnet` to get defaults for and we should just use the websockets versions of these cause that should work (now)

After that, since we're now also accepting a `--network` we should run `getFilecoinNetworkType(provider)` (exported from synapse) and validate that it matches what they asked for and print it out at the beginning that we are using that network just so the user is absolutely sure what they are talking to, whatever args they provide.

rvagg: @copilot try and handle the above suggestion please

TippyFlitsUK: Thanks, Rod. Addressed here https://github.com/FilOzone/synapse-sdk/pull/231/commits/ff9da1f1efb8328c345680703fb6335974f9b2ab

PR #235, opened 2025-09-23
feat: enhance metadata selection and warm storage service tests

### LINKED ISSUE #181: Consider view contract's `clientDataSets` method instead of `getClientDataSets`
In `getClientDataSetsWithDetails` we have an awkward indirection from `getClientDataSets` and then using the `pdpRailId` of each data set to call `railToDataSet()` with it so we can then query the PDPVerifier about a data set. It's pretty smelly.

I notice we have a `clientDataSets` on the view contract which just returns IDs and think maybe we should just start with that and then enhance each one in parallel. Maybe that will smell less.

virajbhartiya:
- getClientDataSetsWithDetails now starts from view clientDataSets(client) IDs and enhances each in parallel (getDataSet, PDPVerifier, metadata).
- Removed railToDataSet indirection/dependence on getClientDataSets for this path.
- Tests and mocks updated; suite passes.

Closes #181

### DIFF
--- a/src/warm-storage/service.ts
+++ b/src/warm-storage/service.ts
@@ -348,17 +348,23 @@ export class WarmStorageService {
       try {
         const base = await this.getDataSet(pdpVerifierDataSetId)
 
+        // Parallelize independent calls
         const [isLive, listenerResult, metadata] = await Promise.all([
           pdpVerifier.dataSetLive(pdpVerifierDataSetId),
           pdpVerifier.getDataSetListener(pdpVerifierDataSetId).catch(() => null),
           this.getDataSetMetadata(pdpVerifierDataSetId).catch(() => Object.create(null) as Record<string, string>),
         ])
 
+        // Check if this data set is managed by our Warm Storage contract
         const isManaged =
           listenerResult != null && listenerResult.toLowerCase() === this._warmStorageAddress.toLowerCase()
 
-        if (onlyManaged && !isManaged) return null
+        // Skip unmanaged data sets if onlyManaged is true
+        if (onlyManaged && !isManaged) {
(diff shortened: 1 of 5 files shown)


src/warm-storage/service.ts
```diff
@@ -335,76 +335,50 @@ export class WarmStorageService {
...
-          }`
+          `Failed to get details for data set ${pdpVerifierDataSetId}: ${error instanceof Error ? error.message : String(error)}`
         )
       }
     })
 
-    // Wait for all promises to resolve
     const results = await Promise.all(enhancedDataSetsPromises)
-
-    // Filter out null values (from skipped data sets when onlyManaged is true)
```
rvagg, line 407: this comment was ~helpful so it can go back

src/warm-storage/service.ts
```diff
@@ -335,76 +335,50 @@ export class WarmStorageService {
...
 
-        // Check if this data set is managed by our Warm Storage contract
         const isManaged =
           listenerResult != null && listenerResult.toLowerCase() === this._warmStorageAddress.toLowerCase()
 
-        // Skip unmanaged data sets if onlyManaged is true
-        if (onlyManaged && !isManaged) {
-          return null // Will be filtered out
-        }
+        if (onlyManaged && !isManaged) return null
```
rvagg, line 360: this didn't need to change, just leave the original form (I'm generally not a fan single-line if's)

virajbhartiya:
@rvagg I've reverted the changes mentioned above

rvagg (approved):

rvagg:
cool, thanks @virajbhartiya!

PR #279, opened 2025-10-05
fix: generate-abi from correct contracts git tag

pali101:
This PR resolves test failures that occur after running `npm run generate-abi`

## Changes

- Updating the `baseUrl` in `wagmi.config.ts`
- Syncing mock data to include the required `providerId` property
- Regenerates ABI files

### DIFF
--- a/packages/synapse-sdk/wagmi.config.ts
+++ b/packages/synapse-sdk/wagmi.config.ts
@@ -2,6 +2,9 @@ import { defineConfig } from '@wagmi/cli'
 import { fetch } from '@wagmi/cli/plugins'
 import type { Address } from 'viem'
 
+const GIT_REF = 'tags/alpha/calibnet/0x80617b65FD2EEa1D7fDe2B4F85977670690ed348-v2'
+const BASE_URL = `https://raw.githubusercontent.com/FilOzone/filecoin-services/refs/${GIT_REF}/service_contracts/abi`
+
 const config = defineConfig(() => {
   const contracts = [
     {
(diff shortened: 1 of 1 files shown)


packages/synapse-sdk/wagmi.config.ts
```diff
@@ -57,7 +57,7 @@ const config = defineConfig(() => {
           cacheDuration: 100,
           request(contract) {
             const baseUrl =
-              'https://raw.githubusercontent.com/FilOzone/filecoin-services/refs/heads/main/service_contracts/abi'
+              'https://raw.githubusercontent.com/FilOzone/filecoin-services/refs/tags/alpha/calibnet/0x80617b65FD2EEa1D7fDe2B4F85977670690ed348-v2/service_contracts/abi'
```
rvagg, line 60: this is nice, but definitely belongs as a constant or two at the top of this file, how about you make `GIT_REF = 'tags/alpha/calibnet/0x80617b65FD2EEa1D7fDe2B4F85977670690ed348-v2'` and then `BASE_URL` which uses `GIT_REF`

rvagg (approved):
lgtm (I've changed the name of this @pali101 so the commit is more descriptive in the release when we squash merge it)

packages/synapse-sdk/wagmi.config.ts
pali101: Done.

rvagg:
nice, thanks @pali101 !

PR #283, opened 2025-10-07
feat: getSizeFromPieceCID(cid) to extract size from PieceCIDv2

rvagg:


packages/synapse-sdk/src/piece/piece.ts
```diff
@@ -204,3 +205,37 @@ export function createPieceCIDStream(): {
...
+ *
+ * PieceCIDv2 encodes the original data size in its multihash digest through
+ * the tree height and padding values. This function decodes those values to
+ * calculate the original raw data size.
+ *
+ * @param pieceCidInput - PieceCID as either a CID object or string
+ * @returns The raw size in bytes
+ * @throws {Error} If the input is not a valid PieceCIDv2
+ */
+export function rawSizeFromPieceCID(pieceCidInput: PieceCID | CID | string): number {
```
rvagg, line 220: I'm slightly indifferent to the `get` prefix to this. I think `getRawSizeFromPieceCID` might be too long, but the short version, `getRawSize`, isn't descriptive enough.

PR #285, opened 2025-10-07
feat: add piece deletion to storage context

### LINKED ISSUE #207: Allow client to delete a piece
Terminate a whole data set is here: https://github.com/FilOzone/synapse-sdk/issues/199

This would be exposing a simple `removePiece` or similar method on StorageContext, design happening in https://github.com/FilOzone/filecoin-services/issues/155, unclear yet whether we go through the Curio schedule-deletion process or do it directly on chain.

Kubuxu:
Uses Curio PDP APIs to trigger piece deletion resolves #207

### DIFF
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -1260,7 +1260,7 @@ export class StorageContext {
     return dataSetData.pieces.map((piece) => piece.pieceCid)
   }
 
-  async _getPieceIdByCID(pieceCID: string | PieceCID): Promise<number> {
+  private async _getPieceIdByCID(pieceCID: string | PieceCID): Promise<number> {
     const parsedPieceCID = asPieceCID(pieceCID)
     if (parsedPieceCID == null) {
       throw createError('StorageContext', 'deletePiece', 'Invalid PieceCID provided')


rvagg:
Opened https://github.com/FilOzone/synapse-sdk/issues/286 to deal with the extra call, I think we can make this nicer by just storing a copy of clientDataSetId when we have it on context creation. It's also used in AddPieces and we have another dedicated call for it then, although at least in that case we can do it in parallel with 3 other calls.

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -1260,6 +1260,28 @@ export class StorageContext {
     return dataSetData.pieces.map((piece) => piece.pieceCid)
   }
 
+  /**
+   * Delete a piece with given CID from this data set
+   * @param pieceCid - The PieceCID identifier
+   * @returns Transaction hash of the delete operation
+   */
+  async deletePiece(pieceCid: string | PieceCID): Promise<string> {
```
rvagg, line 1268: hey, this could be:


```suggestion
  async deletePiece(piece: string | PieceCID | number): Promise<string> {
```

and if it's a `number` then we skip the pieceId look-up and just use that.

Then, if we do https://github.com/FilOzone/synapse-sdk/pull/249#issuecomment-3379605263 and expose a `getPieces(): AsyncGenerator<{ cid: PieceCID, id: number }>` on the context, the user can be in charge of choosing which piece to delete and if there's duplicates they can pinpoint exactly which one.

Kubuxu: Yup, I've implemented accepting the pieceID

Issue #264, opened 2025-10-01
Support new addPieces with create dataset


hugomrdias:
https://github.com/FilOzone/pdp/pull/201

rvagg:
Depends on https://github.com/filecoin-project/curio/issues/678

rvagg:
Quick thought about API here: I think we can just lazily create the data set. When we `createContext()` we either end up with a state where we have an existing data set, or we have no data set. Then when we add the first piece(s) we use CreateAndAddPieces if we have no data set, otherwise just AddPieces. Then it should just work transparently from the user's perspective and everything is just quicker.

Unless I'm not seeing something obvious.

Some methods on `StorageContext` would have to have new behaviours to handle this, like `getPieces` would just auto-return an empty array instead of querying.

rvagg:
Tricky thing to resolve: `StorageCreationCallbacks` has `onDataSetCreationStarted` and `onDataSetCreationProgress` which are going to tell a different story. Considering `StorageCreationCallbacks` further, we have `onProviderSelected` and `onDataSetResolved`, which still can return useful information, but is it really that useful if `createContext()` is returning fairly quickly and will either have a data set or not, and you could just get them off the context itself.

So how about we `@deprecate` `callbacks?: StorageCreationCallbacks` on `StorageServiceOptions` (and while in there, rename that to `StorageContextOptions`, it should have been renamed when we introduced contexts). We can continue to call `onProviderSelected` and `onDataSetResolved`, but the other two don't have any meaningful data that we can call them with. So we are going to have to break this experience for devs unfortunately, hopefully that won't be too painful.

To properly deprecate them, we should make both the data set ID and the provider information available off `StorageContext`, so let's add getters for both of those, but lets clone `ProviderInfo` when we return it so it's a strictly read-only getter. Then when you get a context, you can immediately access `context.dataSetId` and `context.provider`. `dataSetId` should be `null` when we don't have one.

We could think about modifying `UploadCallbacks` to include data set information, but I'm not sure it's that helpful. You can just read `dataSetId` before and after an upload to see it having been created. Let's defer that one until someone asks for insight or we see the workflow and decide it needs something.

rvagg:
@timfong888 that's correct, if you are not using session keys, as per the current demos, when you first encounter an app, after you're set up with payments deposits and approvals, then your first wallet interaction with FWSS will be _after_ you've uploaded your first piece, then you'll get a prompt to sign for the single create+add interaction.

PR #287, opened 2025-10-08
feat: add createDataSetWithPieces method for combined dataset creation and piece addition

virajbhartiya:
- Introduced `createDataSetWithPieces` method in `PDPServer` to allow creating a new data set with pieces in a single operation.
- Added corresponding type `CreateDataSetWithPiecesResponse` for the new response structure.
- Implemented `uploadAndCreate` method in `StorageContext` and `StorageManager` to streamline data upload and dataset creation.
- Enhanced tests to cover scenarios for successful dataset creation with and without pieces, including validation for piece metadata and CID.

Ref https://github.com/filecoin-project/curio/issues/678
Curio PR: https://github.com/filecoin-project/curio/pull/687

### DIFF
--- a/packages/synapse-sdk/src/pdp/index.ts
+++ b/packages/synapse-sdk/src/pdp/index.ts
@@ -13,6 +13,7 @@ export { PDPAuthHelper } from './auth.ts'
 export type {
   AddPiecesResponse,
   CreateDataSetResponse,
+  CreateDataSetWithPiecesResponse,
   DataSetCreationStatusResponse,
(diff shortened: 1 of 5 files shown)


rvagg:
@virajbhartiya please see the notes I left in https://github.com/FilOzone/synapse-sdk/issues/264 about how this should go - it's not a new method, it replaces the current flow and has some subtleties to how it all works. You need to pay careful attention to how the pieces are moving in here and not just add new functionality.

PR #316, opened 2025-10-14
feat: deposit to any address

lordshashank:
This pull request adds support for depositing funds to a different recipient address using the `PaymentsService.deposit()` method, and updates related documentation and type definitions to reflect this new flexibility. The change replaces the previous `DepositCallbacks` parameter with a more general `DepositOptions` object, which now includes an optional `to` property for specifying the recipient address, as well as callback functions for transaction visibility.

**SDK API enhancements:**

* The `PaymentsService.deposit()` method now accepts an optional `DepositOptions` object, which includes an optional `to` property to specify a recipient address other than the signer. This replaces the previous `DepositCallbacks` parameter. [[1]](diffhunk://#diff-295e3bf6a9ceaf5092588db43bd925bc9ce2d8a02262c11f6cba7dbaecad50b6L606-R608) [[2]](diffhunk://#diff-295e3bf6a9ceaf5092588db43bd925bc9ce2d8a02262c11f6cba7dbaecad50b6L22-R26) [[3]](diffhunk://#diff-8f0930977d7e061a4c5737f596245fa7502178779699d9fa919910e3c276fc1cL2-R2) [[4]](diffhunk://#diff-8f0930977d7e061a4c5737f596245fa7502178779699d9fa919910e3c276fc1cL11-R11)
* The deposit transaction now uses the address provided in `options.to` (if present) as the recipient; otherwise, it defaults to the signer's address. [[1]](diffhunk://#diff-295e3bf6a9ceaf5092588db43bd925bc9ce2d8a02262c11f6cba7dbaecad50b6R621) [[2]](diffhunk://#diff-295e3bf6a9ceaf5092588db43bd925bc9ce2d8a02262c11f6cba7dbaecad50b6L661-R664)
* All deposit operation callbacks are now accessed through the `DepositOptions` object.

**Documentation updates:**

* The documentation and code examples in `AGENTS.md`, `getting-started.mdx`, and `components.mdx` have been updated to describe and demonstrate the new `to` option for deposits, including sample usage for depositing to another address. [[1]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L20-R20) [[2]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R211) [[3]](diffhunk://#diff-c94cc37d907ce4c63e716de33e64f0319937e83fbd6ce0ac471e2c58b691b80bR173-R176) [[4]](diffhunk://#diff-c627cf6cc7f69732e42b9d6151505d03e5d8020c6d646ada96bec5d37e4d3ea2R25-R28) [[5]](diffhunk://#diff-c94cc37d907ce4c63e716de33e64f0319937e83fbd6ce0ac471e2c58b691b80bL380-R384)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,7 +17,7 @@ This document serves as context for LLM agent sessions working with the Synapse 
 ### Key Components
 
 - `Synapse`: Main SDK entry; minimal interface with `payments` property and `storage` manager; strict network validation (mainnet/calibration).
-- `PaymentsService`: Payment operations - deposits, withdrawals, balances, service approvals.
+- `PaymentsService`: Payment operations - deposits, withdrawals, balances, service approvals. The `deposit()` method accepts an optional `DepositOptions` object with `to` property to deposit funds to a different address than the signer, plus callback functions for visibility.
 - `SPRegistryService`: Service provider registry - registration, updates, product management, provider discovery.
 - `WarmStorageService`: Storage coordination - costs, allowances, data sets. Factory method `WarmStorageService.create(provider, address)`. Source of all contract addresses via discovery.
 - `StorageManager/StorageContext`: Storage operations with auto-managed or explicit contexts.
@@ -208,6 +208,7 @@ WarmStorageService (storage coordination)
 - Manages payment rails between parties
 - Supports operator approvals for account management
 - Address discovered from WarmStorage
+- **Deposit Flexibility**: The `deposit(token, to, amount)` function allows depositing to any address, not just the signer. SDK's `PaymentsService.deposit()` exposes this via `DepositOptions` object with optional `to` property.
 
 #### 5. Curio Storage Provider (Service Node)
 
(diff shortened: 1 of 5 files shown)


rvagg:
pretty ugly, I don't like this being right at the end after the callbacks.

I think @hugomrdias wants to turn all of the arguments to these calls into `{}` eventually, which would be much easier to deal with. But for now, how about you rename `DepositCallbacks` to `DepositOptions` and put a `to` property on that. Then you're not doing an `undefined` but a `{ to: wallet }`

rvagg:
you can type alias and `@deprecate` if you're worried about breakage there

another alternative is to make `DepositOptions` take your `to` but also a `callbacks: DepositCallbacks`, and then in that last position we take `DepositOptions | DepositCallbacks` and check for the `callbacks` property to figure out what we got - do that as an interim measure, document clearly that it's going to change, remove docs for the old way and we can ship that without breakage

We can also break things, that's ok, just semver-minor. People will pick up the type changes when the break downstream, that's one nice thing about TypeScript.

wjmelements (approved):

PR #327, opened 2025-10-18
feat(upload): add exponential backoff to retry logic

### LINKED ISSUE #293: fix: add exponential backoff for findPiece
https://github.com/FilOzone/synapse-sdk/blob/5a35d4328f8fc84c8ff038d74ff12cca117f04f6/packages/synapse-sdk/src/storage/context.ts#L969-L971

instead of only increasing timeout delay by interval on each failure, we should add exponential backoff and potentially jitter

DarkLord017:
This pull request updates the polling logic in the `StorageContext` class to use exponential backoff with jitter when waiting for a piece to be "parked" (ready). 

Fixes #293 

Improvements to polling and retry logic:

* Replaced the fixed polling interval with an exponential backoff strategy, introducing a `baseDelay`, `maxDelay`, and an `attempt` counter to increase the delay between retries.
* Added random jitter to each backoff delay to help prevent synchronized retries and further reduce contention.

### DIFF
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -1015,7 +1015,9 @@ export class StorageContext {
 
       // Poll for piece to be "parked" (ready)
       const maxWaitTime = TIMING_CONSTANTS.PIECE_PARKING_TIMEOUT_MS
-      const pollInterval = TIMING_CONSTANTS.PIECE_PARKING_POLL_INTERVAL_MS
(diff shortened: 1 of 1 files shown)


rvagg (changes requested):
I don't love this, I'm not really convinced we need this, it's a normal process of waiting for piece parking, IMO backoff is good for error conditions, but this isn't an error condition, all this will do is make it longer for a piece park wait to happen

blocking this for now, needs more consideration before merging, maybe someone else wants to make a stronger argument for this

PR #331, opened 2025-10-20
docs: Add Introduction Section

### LINKED ISSUE #322: Docs: Introduction
The introduction section should include:
- Overview 
- About the Filecoin Onchain Cloud
- Why Filecoin Onchain Cloud

longfeiWan9:
Closes #322 

This PR introduces a new "Introduction" section to the documentation with two key pages that provide a comprehensive overview of Filecoin Onchain Cloud (FOC) for new users and developers.

### Changes Made

 📄 New Pages Added

- **`intro/about.md`** - Comprehensive overview of what Filecoin Onchain Cloud is, its architecture, and core components
- **`intro/why.md`** - Detailed explanation of why developers should choose FOC over traditional cloud solutions

🏗️ Re-structure the content for coming content, includes:
- Quick Start
- Concepts
- Developer Guides


docs/src/content/docs/intro/why.md
```diff
@@ -0,0 +1,54 @@
...
+- 🚀 **Foster innovation** by letting teams build, fork, and combine services into new decentralized applications
+
+### Key Differentiators
+
+| Feature | Traditional Cloud | Filecoin Onchain Cloud |
+|---------|------------------|------------------------|
+| Data Control | Provider controlled | User owned |
+| Transparency | Limited | Full blockchain verification |
+| Programmability | Provider APIs | Smart contracts |
+| Payment Models | Fixed | Flexible & programmable |
```
timfong888: Not a strong concern for me, but I could see a developer wondering whether this means what *they pay and how they pay* for the primary offering, which is the warm storage, is flexible to them.  I think developers want clarity on how they pay for the service.

I believe this is intended for in the event the developer creates their own service and want to bill their own customers, they can do that with Filecoin Pay?

If so, can we frame it as we enable them to bill their customers flexibly?  Versus they can pay for the Storage Service we provide out of the box flexibly?

docs/src/content/docs/guides/rails-settlement.mdx
timfong888: It references for the user to keep settling regularly as normal operations....is this essential for the developer to be settling their funds?  

Second: I think a section should say, "Terminate service" - so they understand how to stop paying, get unused funds back.  The developer probably wants to know: 1. How do I pay for it; 2. If i don't like it, how do I stop.

timfong888 (changes requested):
I made some suggestions, none earth shattering but for clarity and priority:

1. Payment Rails: when I read it, the flexibility sounds like I, as the developer, can pay flexibility for the warm storage service; even if we could, I don't think that's what they want, it adds cognitive load.  My understanding is they can set up rails for services they offer to their clients, so it's like a way to enable billing to customers for their own services.

2. Related to payment rails: a core goal and metric is exposure of their customers to pay directly for the storage as a preferred mechanism.  If we could nudge that way in the text and example.

3. The rest are minor organizational and text changes.

docs/src/content/docs/guides/components.mdx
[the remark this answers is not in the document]
longfeiWan9: The quick start will be added in a separate PR. And I do not think it is necessary for the develoeprs to know much about operators when they use Synapse SDK

docs/src/content/docs/intro/about.md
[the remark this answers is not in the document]
longfeiWan9: The quick start guide for uploads will be on the "Getting Started" page. This page serves as an introduction, focusing on explaining what FoC is.

docs/src/content/docs/intro/why.md
longfeiWan9: This payment models is just general point of paying the Filecion warm storage service. At this point, we take developers as our audience regardless how they bill their own customers. 

In terms of developers want clarity on how they pay for the service, this will be covered in the developer guides under payment rails.

PR #334, opened 2025-10-21
Update ABIs for next contracts

Kubuxu:


hugomrdias:
I removed all these docs build jobs in master for now, docs build in a different workflow and cloudflare on node lts.

PR #335, opened 2025-10-21
chore: merge core and react

hugomrdias:


packages/synapse-core/src/session-key/secp256k1.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export class Secp256k1Key extends TypedEventTarget<WalletEvents> implements SessionKey {
+  private privateKey: Hex
+  permissions: SessionKeyPermissions[]
+  expiresAt: number | undefined
+  type: 'secp256k1'
+  account: Account
+  private isConnecting: boolean = false
+  private isConnected: boolean = false
+  private connectPromise: Promise<TransactionReceipt | void> | undefined
```
wjmelements: lint complains about this:
```
  ℹ Unsafe fix: Use undefined instead.

     40  40 │     private isConnecting: boolean = false
     41  41 │     private isConnected: boolean = false
     42     │ - ··private·connectPromise:·Promise<TransactionReceipt·|·void>·|·undefined
         42 │ + ··private·connectPromise:·Promise<TransactionReceipt·|·undefined>·|·undefined
     43  43 │
     44  44 │     constructor(props: Secp256k1SessionKeyProps) {


Checked 196 files in 637ms. No fixes applied.
```

PR #344, opened 2025-10-23
feat: use core curio client in sdk pdp

hugomrdias:
- synapse-core curio client adds polling, better errors 
- removes `piece_cid` deprecated prop

closes #331


rvagg:
@hugomrdias sorry, I didn't realise this had error decoding stuff in it and didn't register it as a priority.

I'm fine with this approach, looks good, I'd just prefer the naming be less specific as discussed already.

Issue #348, opened 2025-10-23
bug: guarding upload success by requiring status===204


SgtPooki:
https://github.com/FilOzone/synapse-sdk/blob/1167900aa117026d084842708b760eb1f3ffd203/packages/synapse-sdk/src/pdp/server.ts#L577

im not sure if this is some intentional expectation, but this feels wrong.

HTTP 204 is "No Content"

I am integrating telemetry (#328) and have errors emitting now by running `tsx utils/example-storage-e2e.js ~/Downloads/monero-gui-mac-armv8-v0.18.3.3.dmg`.. The pdp server is responding with a `HTTP 201`, which means `Created`, but since we don't get a 204, we emit a `StorageContext uploadPiece failed: Failed to upload piece to service provider - Failed to upload piece: 500 Internal Server Error - Failed to store piece data` error.

See the breadcrumbs here:

<img width="1660" height="1524" alt="Image" src="https://github.com/user-attachments/assets/f79bb56e-ee40-4652-920b-b17029fc39e0" />

References:

* https://filoz.sentry.io/issues/6969186662/?query=is%3Aunresolved&referrer=issue-stream
* https://filecoinproject.slack.com/archives/C095WFA0QK1/p1761241667586879?thread_ts=1761227812.361289&cid=C095WFA0QK1

SgtPooki:
@rvagg ok I think I got mixed up. When I first saw this, I was pretty sure the server responded to the final call with "201" and that the error I saw was from https://github.com/FilOzone/synapse-sdk/blob/1167900aa117026d084842708b760eb1f3ffd203/packages/synapse-sdk/src/pdp/server.ts#L579

I can understand preferring 204 to 201 after re-reading https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.5. Thanks for the context. 

The only issue I could see in the future is if someone is running software other than curio and returns an objectively valid 200 or 201, but I imagine we're not trying to focus on other software backends at the moment.

PR #349, opened 2025-10-23
feat: add telemetry

### LINKED ISSUE #328: Self-service way for maintainers to get client side errors/logs
## Done Criteria
The Synapse SDK captures and reports frontend errors, logs, and user actions to a centralized location with request/trace IDs that correlate with backend Curio PDP SP logs, enabling maintainers to quickly debug issues end-to-end.

Basically a user should be able to let us know they had a problem and we can quickly get to debugging what happened with minimal intervention from them or the backend SP. 

## Why Important
Between now and our November GA launch, making our product robust and reliable is a key focus. We'll be spending significant time debugging issues across distributed service providers. Centralizing frontend telemetry alongside backend logs (see [curio#712](https://github.com/filecoin-project/curio/issues/712)) should reduce our debugging cycle time by allowing maintainers to:
- See the full user journey from browser to backend and back
- Correlate frontend errors with backend behavior using shared request IDs
- Understand what users were doing when issues occurred
- Reproduce issues with full context (browser, network conditions, user actions)
- Debug async operations that span frontend and backend

## User/Customer
1. **Maintainers/implementers** - Can see the complete picture of user issues including frontend state, actions, and errors correlated with backend logs
2. **End users** - Benefit from faster issue resolution and fewer requests to "tell us what you were doing"

## Notes

This is a short-term measure.  It's not intended to be "the long term solution".  The tracking issue to cleanup and/or improve at GA is https://github.com/FilOzone/synapse-sdk/issues/363

### Principles
1. **Speed over perfection** - Focus on getting something functional quickly. We can iterate and improve after launch.
2. **Ease of operation** - Favor managed SaaS solutions that integrate quickly over building custom instrumentation.
3. **User privacy** - Be mindful of what data we collect. Focus on technical debugging data, not sensitive user information.
4. **Request correlation** - Generate request/trace IDs on the frontend and pass them to backend so we can tie logs together.

### What to Capture
- Uncaught JavaScript errors and promise rejections
- Synapse SDK method calls and parameters (sanitized)
- Backend "API" requests/responses (status codes, timing, errors) - whether to SPs or RPCs providers querying the chain
- State transitions within the SDK?
- Browser user agent
- Request/trace IDs for correlation with backend

### Potential Technology Options to Evaluate
**Frontend-focused SaaS:**
- **Sentry** - For error tracking, automatic breadcrumbs, release tracking
- **LogRocket** - Session replay + error tracking (may be overkill initially)
- **Datadog RUM** - If we use Datadog for backend, can use same platform
- **Better Stack** - Can handle both frontend and backend logs in one place

**DIY Approach (likely not recommended):**
- Custom beacon endpoint to our own logging service.  I assume more work to build, but gives full control

### Implementation Approach
1. Select frontend error/log tracking service (target: < 1 day)
2. Integrate synapse SDK with tracking service
3. Implement request ID generation and propagation to backend API calls
4. Add SDK configuration options for opt-in/debugging mode if needed
5. Document for maintainers how to correlate frontend/backend logs using request IDs

### Integration with Backend Logging
- Ensure request IDs generated in SDK are passed to Curio backend (headers, query params, or request body)
- Backend logs should include these request IDs (see [curio#712](https://github.com/filecoin-project/curio/issues/712))
- Document the full tracing flow for maintainers

### Success Metrics
- Capture most JavaScript errors that occur
- Request IDs successfully correlate frontend and backend logs
- Time from issue report to full context access < 5 minutes
- Reduced median time-to-resolution for debugging issues
- Minimal performance overhead

## Related Issues
- [filecoin-project/curio#712](https://github.com/filecoin-project/curio/issues/712) - Backend logging for Curio SPs

SgtPooki:
Adds telemetry functionality in `packages/synapse-sdk/src/telemetry`

Attempts to touch existing code as little as possible..

There is a singleton to prevent us from passing a telemtry instance through multiple layers and touching a lot of code.

What do we get from this?

e.g. https://github.com/FilOzone/synapse-sdk/issues/348

What type of data do we get?

* breadcrumbs showing related operations of failures..
* tracing to see timing of different events (currently the operation is based on the `extractSPInfo` function) in the fetch wrapper
* some automatic breadcrumbs from console.log statements (when usint utils/logger.ts)

### DONE
- [x] Synapse Integration: Telemetry properly integrated into Synapse.create() with config options
- [x] HTTP Tracking: Global fetch wrapper
- [x] Error Integration: Automatic error capture via enhanced createError() function
- [x] Debug Support: debugDump() method for consumers to output to help with tracking issues
- [x] test in filecoin-pin CLI: https://github.com/filecoin-project/filecoin-pin/pull/154

### TODO:
- [ ] Test in filecoin-pin/upload-action
- [x] test in filecoin-pin-website
- [x] Ensure telemetry is not emitting during testing.
- [x] Fix `${prefix}.${operation}` errors mapping and normalize OPERATIONS constant to match. (it should make debugging easier, but idk that we can do the mapping easily.. probably good enough for EOW?)
- [x] Global Error Handlers - Add unhandledrejection and uncaughtException handlers
- [x] Environment Detection - Fix SYNAPSE_TELEMETRY_DISABLED environment variable support
- [x] SDK Version - Get version from package.json instead of hardcoded "0.34.0"
- [x] Add some unit tests
- [x] add documentation to readme about "opt-out" (enabled by default) and how to disable.

### Deferred

- [ ] Service Instrumentation - Add trackOperation() calls to key services (PDPServer, PaymentsService, StorageContext, etc.)
- [ ] Figure out if we want to pipe through pino logging with something like https://www.npmjs.com/package/pino-sentry

Fixes #328

### DIFF
--- a/docs/src/content/docs/guides/telemetry.md
+++ b/docs/src/content/docs/guides/telemetry.md
@@ -3,7 +3,7 @@ title: Telemetry
 description: Notes about the telemetry functionality that is within Synapse.
 ---
 
-To help us validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack, starting from the SDK, telemetry is **temporarily enabled by default for the calibration network** in Synapse.  We are currently leveraging sentry.io as discussed in [issue #328](https://github.com/FilOzone/synapse-sdk/issues/328).
+To help maintainers validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack, starting from the SDK, telemetry is **temporarily enabled by default for the calibration network** in Synapse.  We are currently leveraging sentry.io as discussed in [issue #328](https://github.com/FilOzone/synapse-sdk/issues/328).
 
 ### How to disable telemetry
 
@@ -33,7 +33,7 @@ We are not capturing:
 - Personal identifiable information (PII).  We explicitly [disable sending default PII to Sentry](https://docs.sentry.io/platforms/javascript/configuration/options/#sendDefaultPii).
 - Metrics on static asset (e.g., CSS, JS, image) retrieval.  
 
-(One can verify these claims in [src/telemetry/service.ts](src/telemetry/service.ts).)
+(One can verify these claims in [telemetry/service.ts](https://github.com/FilOzone/synapse-sdk/blob/master/packages/synapse-sdk/src/telemetry/service.ts).)
 
 ### Why is telemetry collecting happening a library like Synapse
 Collecting telemetry through Synapse with [issue #328](https://github.com/FilOzone/synapse-sdk/issues/328) is done as short a term dev-resource efficient decision.  In this season of focusing on stability, the goal is to capture request failures and other client-side errors as broadly and quickly as possible so we have an enumeration of the problems and their impact.  By setting up telemetry at the Synapse layer, we can broadly get telemetry from some of the first consumers by default without requiring extra on them (e.g., filecoin-pin,filecoin-pin-website, synapse demo websites).  This is a short term measure.
--- a/docs/src/content/docs/intro/getting-started.mdx
+++ b/docs/src/content/docs/intro/getting-started.mdx
@@ -54,7 +54,7 @@ Before diving into the code, understand these fundamental concepts:
 
 ## Telemetry
 
-To help maintainers validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack, starting from the SDK, telemetry is **temporarily enabled by default for the calibration network** in Synapse.  [Read more (including how to disable)](../guides/telemetry.md).
+To help maintainers validate functionality and iron out problems throughout the whole Filecoin Onchain Cloud stack, starting from the SDK, telemetry is **temporarily enabled by default for the calibration network** in Synapse.  [Learn more (including how to disable)](/guides/telemetry).
 
 ## Usage
 


BigLep:
The other thing in practice that I think we'll want is for filecoin-pin to be able to measure interactions as well.  For example, we want to know how long it's taking from the client-side perspective from "file was uplaoded to SP" but how long before it is retrievable (because it's discoverable through cid.contact → filecoinpin.contact).  I assume there is thought that needs to be done to support this (as ideally we should be capturing the filecoin-pin version).  

Or should filecoin-pin just have it's own sentry instance?  I guess that is probably cleanest... ?  (If so, I can create an issue for it)

SgtPooki:
> I assume there is thought that needs to be done to support this (as ideally we should be capturing the filecoin-pin version).

> Or should filecoin-pin just have it's own sentry instance? I guess that is probably cleanest... ? (If so, I can create an issue for it)

the `synapse.telemetry.createSpan` will allow us to do that. I plan to test in filecoin-pin website and CLI and get example telemetry (with PRs for those repos) before this PR moves out of draft.

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
```diff
@@ -0,0 +1,170 @@
...
+        duration: event.durationMs,
+        requestId: event.requestId,
+        sp_hostname: event.spHostname,
+        sp_path: event.spPath,
+        sp_operation: event.spOperation,
+        http_method: event.method,
+        http_status: event.status?.toString(),
+      },
+      level: event.ok ? 'info' : 'error',
+    })
```
SgtPooki: idk if it matters that the breadcrumb is emitted after the span?

SgtPooki (commented):
self review

packages/synapse-sdk/src/telemetry/fetch-wrapper.ts
```diff
@@ -0,0 +1,233 @@
...
+      const durationMs = Date.now() - startTime
+
+      // Extract SP information from URL
+      const spInfo = extractSPInfo(parsedUrl, init?.method || 'GET')
+
+      const httpEvent: HTTPEvent = {
+        type: 'http',
+        method: init?.method || 'GET',
+        urlTemplate: `${parsedUrl.protocol}//${parsedUrl.host}${parsedUrl.pathname}`,
+        status: 500,
```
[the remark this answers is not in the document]
BigLep: We get into this block because of a network error right?
- Network connection fails (no internet, server unreachable)
- DNS lookup fails
- Request is blocked by browser (CORS preflight failure, mixed content, etc.)
- Request times out (in environments that implement timeouts)
- SSL/TLS certificate errors

As a result, I don't think we should give this a status code since we didn't get one.  Maybe this shouldn't even been of type "http"?  (I'm no expert here, but in my searching it doesn't look like you can glean too much about the error cause.  Maybe we just log the whole error object as a string?)

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,443 @@
...
+        durationMs: Date.now() - startTime,
+        requestId,
+        ts: new Date().toISOString(),
+      }
+
+      this.addToBuffer(event)
+      this.adapter.captureOperation(event)
+
+      return result
+    } catch (error) {
```
BigLep: how much do we even need this?  Are any errors even expected?  Do we just let our global error catcher handle it?

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,443 @@
...
+   * Enable telemetry explicitly (even if disabled by environment)
+   * Useful for testing or when you need to force telemetry on
+   */
+  enable(): void {
+    if (!this.enabled) {
+      this.enabled = true
+      this.adapter.init(
+        {
+          enabled: true,
+          environment: this.context.runtime === 'browser' ? 'development' : 'production',
```
BigLep: why is the environment dependent on the context.rutime?

packages/synapse-sdk/src/telemetry/types.ts
```diff
@@ -0,0 +1,147 @@
...
+   * Whether to enable telemetry.
+   *
+   * You can also control this via the environment variable `SYNAPSE_TELEMETRY_DISABLED` or the global variable `SYNAPSE_TELEMETRY_DISABLED`.
+   *
+   * This value will also be false if `NODE_ENV === 'test'`.
+   *
+   * @default true
+   */
+  enabled?: boolean
+  environment?: Environment // optional: deployment environment
```
BigLep: do we need this?  I'm wondering if is overkill right now and we maybe wait until we identify a need?

packages/synapse-sdk/src/telemetry/types.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export interface ErrorEvent extends BaseTelemetryEvent {
+  type: 'error'
+  name: string
+  message: string
+  stack?: string
+  operation?: OperationType // SDK operation context when error occurred
+}
+
+export interface HTTPEvent extends BaseTelemetryEvent {
```
BigLep: Are these object keys certain things that Sentry is expectry?

packages/synapse-sdk/src/telemetry/types.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+export interface CustomEvent extends BaseTelemetryEvent {
+  type: 'custom'
+  name: string
+  data: Record<string, any>
+  level: 'info' | 'warning' | 'error'
+}
+
+export const OPERATIONS = {
```
BigLep: Is there a reason this is an object with key=value, rather than just an arrary?

BigLep: Is it to support "'custom.operation': `{string}.{string}`"

packages/synapse-sdk/src/test/telemetry.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      const synapse = await Synapse.create({ provider })
+
+      // Verify that telemetry is disabled
+      assert.isFalse(synapse.telemetry.isEnabled())
+
+      // Verify that debug dump shows disabled state
+      const debugDump = synapse.telemetry.debugDump()
+      assert.isFalse(debugDump.context.enabled)
+      assert.equal(debugDump.context.runtime, 'node')
+      assert.equal(debugDump.context.network, 'calibration')
```
BigLep: Is this even relevant?

packages/synapse-sdk/src/test/telemetry.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      const synapse = await Synapse.create({ provider })
+
+      // Get debug dump
+      const debugDump = synapse.telemetry.debugDump()
+
+      // Verify structure
+      assert.isObject(debugDump)
+      assert.isArray(debugDump.events)
+      assert.isObject(debugDump.context)
+      assert.isString(debugDump.timestamp)
```
BigLep: Are these asserts even import?  Just assert that we're disabled and have no events?

packages/synapse-sdk/src/test/telemetry.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  describe('Explicit Enable', () => {
+    it('should allow enabling telemetry explicitly even when disabled by environment', async () => {
+      // Set up proper mock responses for contract calls
+      server.use(JSONRPC(presets.basic))
+
+      // Create a real provider that will use the mocked responses
+      const provider = new ethers.JsonRpcProvider('https://api.calibration.node.glif.io/rpc/v1')
+
+      // Create Synapse instance - telemetry should be disabled by default in test environment
+      const synapse = await Synapse.create({ provider })
```
BigLep: I don't know current best practices, but these lines are duplicated a few times.

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
```diff
@@ -0,0 +1,170 @@
...
+      runtime: tags.runtime,
+      network: tags.network,
+      ua: tags.ua,
+      appName: tags.appName,
+    })
+
+    // Set global tags
+    Sentry.setTag('sdk_version', tags.sdkVersion)
+    Sentry.setTag('runtime', tags.runtime)
+    Sentry.setTag('network', tags.network)
```
BigLep: Why do we have some of these as global, and some of them set to a context?

BigLep:
FYI that I have been going through this PR.  I'm not done yet (still active in progress) because I'm also getting up to speed on Sentry more in the process.

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    const span = Sentry.startInactiveSpan({
+      name: `${event.method} ${event.urlTemplate}`,
+      op: event.spOperation,
+      startTime: startTime,
+      attributes: {
+        // Use standard searchable properties
+        action: event.method, // HTTP method (GET, POST, etc.)
+        domain: event.spHostname, // Hostname for searching
+        description: `${event.method} ${event.urlTemplate}`, // Full description
```
BigLep: Do we need this?  Isn't it covered in name above?

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
```diff
@@ -0,0 +1,170 @@
...
+  captureHTTP(event: HTTPEvent): void {
+    const startTime = new Date(event.ts).getTime()
+
+    const span = Sentry.startInactiveSpan({
+      name: `${event.method} ${event.urlTemplate}`,
+      op: event.spOperation,
+      startTime: startTime,
+      attributes: {
+        // Use standard searchable properties
+        action: event.method, // HTTP method (GET, POST, etc.)
```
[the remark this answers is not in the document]
BigLep: oh I see, you're saying if we want them searchable, we need to set it as "action".

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  captureHTTP(event: HTTPEvent): void {
+    const startTime = new Date(event.ts).getTime()
+
+    const span = Sentry.startInactiveSpan({
+      name: `${event.method} ${event.urlTemplate}`,
+      op: event.spOperation,
+      startTime: startTime,
+      attributes: {
+        // Use standard searchable properties
```
BigLep: Can we link to the docs that defines these standard properties?  It looks like https://docs.sentry.io/concepts/search/searchable-properties/spans/ ?

Given tags are also searchable, I'm wondering the value of trying to conform to their standard properties vs. us just using our own set of tags that are maybe more self documenting?

SgtPooki:
currently trying to solve browser imports in filecoin-pin-website without having to finagle the build here in synapse-sdk

packages/synapse-sdk/README.md
```diff
@@ -42,6 +42,24 @@ Note: `ethers` v6 is a peer dependency and must be installed separately.
 
 Check the documentation [website](https://synapse.filecoin.services/)
 
+## Telemetry
+
+Telemetry is enabled by default to help us validate functionality while moving towards GA launch. Disable it at runtime if you prefer:
```
rvagg: We need to be super up-front and clear about this, it's going to be a very sensitive topic for much of our user-base.

```suggestion
To help us validate functionality and iron out problems throughout the whole stack, starting from the SDK, telemetry is **enabled by default** in Synapse. This will be the case during the current development period prior to our upcoming General Availability launch.

You are welcome to disable it in the runtime when you use Synapse if you would prefer to opt out:
```

Can we say anything more about what's collected, that it's "not personally identifying information" or anything like that?

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,443 @@
...
+   * Uses setTimeout instead of setInterval to avoid keeping the process alive.
+   */
+  private setupIntervalFlushing(): void {
+    const scheduleNext = () => {
+      setTimeout(() => {
+        this.flush().finally(() => {
+          // Schedule the next flush only after the current one completes
+          scheduleNext()
+        })
+      }, 5000)
```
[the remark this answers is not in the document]
SgtPooki: This auto flushing has been removed.. It was causing some issues

packages/synapse-sdk/src/telemetry/types.ts
SgtPooki: Ive boiled this down to a simple operation string type instead of a key map.

Will reduce further

packages/synapse-sdk/src/telemetry/types.ts
```diff
@@ -0,0 +1,147 @@
...
+  captureCustomEvent(event: CustomEvent): void
+  setContext(tags: Record<string, string>): void
+
+  /**
+   * Start a span for manual timing control
+   *
+   * NOTE: For SDK operations, use TelemetryService.trackOperation() instead.
+   * This is for advanced cases where you need manual span lifecycle control.
+   */
+  startSpan(
```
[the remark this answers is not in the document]
SgtPooki: Im imagining this is used by clients to start a span prior to a collection of synapse events.

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
SgtPooki: Tags were not searchable for me when i was testing. Maybe i missed something

packages/synapse-sdk/src/test/telemetry.test.ts
SgtPooki: since telemetry is instantiated when synapse.create happens, thats the test entry. the provider and server.use could probably be reused though

packages/synapse-sdk/src/telemetry/adapters/sentry.ts
SgtPooki: sdk_version will allow us to search based off it, but we can also search based off of "release" which is now set, so i will clean this up.. only planning on `setTag` for now, and Context `environment` for things we don't care about searching but might be important when looking into breadcrumbs/events

SgtPooki (commented):
responding to some comments... large reduction in code incoming

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,217 @@
...
+      this.sentry.setTag('network', this.context.network as 'mainnet' | 'calibration') // only two values, useful for searching
+    }
+  }
+  private initSentry(): void {
+    this.sentry.init({
+      dsn: 'https://3ed2ca5ff7067e58362dca65bcabd69c@o4510235322023936.ingest.us.sentry.io/4510235328184320',
+      // Setting this option to false will prevent the SDK from sending default PII data to Sentry.
+      // For example, automatic IP address collection on events
+      sendDefaultPii: false,
+      // environment: config.environment || 'production',
```
SgtPooki: remove
```suggestion
```

packages/synapse-sdk/src/telemetry/config.ts
```diff
@@ -0,0 +1,87 @@
...
+  return config?.enabled ?? true
+}
+
+/**
+ * Get environment-specific telemetry configuration
+ *
+ * @param userConfig - User-provided configuration
+ * @returns Resolved configuration with environment defaults
+ */
+export function resolveTelemetryConfig(
```
BigLep: Looks like we can get rid of this function since it's only be called once, and it's just being called to get at whether we should enabled telemetry.  Instead call call `shouldEnableTelemetry` directly?

packages/synapse-sdk/src/synapse.ts
```diff
@@ -221,6 +247,23 @@ export class Synapse {
...
+   * const dump = synapse.telemetry.debugDump()
+   * console.log(JSON.stringify(dump, null, 2))
+   *
+   * // Track custom events
+   * synapse.telemetry.captureCustomEvent('user-action', { action: 'upload' })
+   * ```
+   */
+  get telemetry(): TelemetryService {
+    return this._telemetry
+  }
```
BigLep: We have two access patterns for the TelemetryService singleton: via synapse.telemetry and via the telemetry.singleton.  How about we just have one?

I assume the easiest is to move the construction of the TelemetryService into the singleton (passing it the telemetry config and context that are generated here) and then having everyone access via the singleton class?

packages/synapse-sdk/src/telemetry/errors.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Telemetry-aware error creation utilities
+ *
+ * Provides error creation functions that automatically capture errors to telemetry
+ * when telemetry is enabled. Uses the global telemetry singleton.
```
BigLep:
```suggestion
 * Telemetry-aware wrapper around `../utils/errors.ts`
 * Provides the same error creation functions, but also automatically capture errors to telemetry when telemetry is enabled. 
 * Uses the global telemetry singleton.
 * Generic error handling of uncaught errors is configured outside this class.
```

packages/synapse-sdk/src/telemetry/fetch-wrapper.ts
```diff
@@ -0,0 +1,57 @@
...
+    // If telemetry disabled, use the original fetch
+    // OR, we have an active span, and fetch calls will be instrumented by Sentry automatically
+    const sentry = getGlobalTelemetry()?.sentry
+    if (!sentry || sentry.getActiveSpan() != null) {
+      return originalFetch(input, init)
+    }
+    const url = input instanceof Request ? new URL(input.url) : new URL(input.toString())
+    const method = input instanceof Request ? input.method : init?.method || 'GET'
+
+    // currently showing up as TWO items in the sentry UI..you can filter these out in the sentry Trace explorer with `!span.op:http.wrapper`
```
BigLep:
```suggestion
    // This means there will be two currently showing up as TWO items in the sentry UI..you can filter these out in the sentry Trace explorer with `!span.op:http.wrapper`
```

packages/synapse-sdk/src/telemetry/get-sentry.ts
```diff
@@ -0,0 +1,27 @@
...
+export type Sentry = typeof SentryNode.default | typeof SentryBrowser.default
+
+export async function getSentry(): Promise<{ Sentry: Sentry; integrations: any[] }> {
+  if (isBrowser) {
+    const SentryBrowser = await import('@sentry/browser')
+    return {
+      Sentry: SentryBrowser,
+      integrations: [
+        SentryBrowser.browserTracingIntegration({
+          ignoreResourceSpans: ['resource.script', 'resource.img', 'resource.css', 'resource.link'],
```
BigLep:
```suggestion
          // We don't want to clutter the Sentry Trace Explorer with traces for static assets currently so we stay focused on the HTTP calls to RPC providers and SPs.
          ignoreResourceSpans: ['resource.script', 'resource.img', 'resource.css', 'resource.link'],
```

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+   *
+   * You can also control this via the environment variable `SYNAPSE_TELEMETRY_DISABLED` or the global variable `SYNAPSE_TELEMETRY_DISABLED`.
+   *
+   * This value will also be false if `NODE_ENV === 'test'`.
+   *
+   * @default true
+   */
+  enabled?: boolean
+  /**
+   * The name of the application using the SDK.
```
BigLep: Let's be clear about what SDK we're talking about.

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+    this.sentry.setTag('synapseSdkVersion', `@filoz/synapse-sdk@v${this.context.sdkVersion}`) // useful to know which version of the SDK is being used
+    this.sentry.setTag('runtime', this.context.runtime) // only two values, useful for searching
+    this.sentry.setTag('network', this.context.network as 'mainnet' | 'calibration') // only two values, useful for searching
+  }
+
+  protected sanitizeError(event: any): any {
+    this.addToBuffer(event)
+
+    return event
+  }
```
BigLep: I think we can remove this and just have the caller call "addToBuffer" directly.

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+  private initSentry(): void {
+    this.sentry.init({
+      dsn: 'https://3ed2ca5ff7067e58362dca65bcabd69c@o4510235322023936.ingest.us.sentry.io/4510235328184320',
+      // Setting this option to false will prevent the SDK from sending default PII data to Sentry.
+      // For example, automatic IP address collection on events
+      sendDefaultPii: false,
+      release: `@filoz/synapse-sdk@v${this.context.sdkVersion}`,
+      beforeSend: this.sanitizeError.bind(this),
```
BigLep:
```suggestion
      beforeSend: this.addToBuffer.bind(this),
```

Doesn't seem to be any point adding extra indirection now

BigLep (commented):
I haven't fully finished going through but this is the next checkpoint of feedback I have.  I'll finish going through the rest of the files later today.  

I'm generally just looking at how we can make this easier to follow and reason about.

packages/synapse-sdk/src/telemetry/singleton.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Telemetry singleton manager
+ *
+ * Provides a single global telemetry instance that can be used by all telemetry
+ * components (fetch wrapper, error tracking, etc.).
```
BigLep:
```suggestion
 * Telemetry singleton manager.
 * Sets up and provides a single global telemetry instance that can be used by all telemetry components:
 * - fetch wrapping via fetch-wrapper.ts
 * - error wrapping via errors.ts
 * Synapse SDK consumers (e.g., filecoin-pin) via #getGlobalTelemetry.
 * 
 * #setupGlobalTelemetry is the entry point and #getGlobalTelemetry is the expected access point within Synapse and beyond.
 * 
 * Setting up the fetch wrapper is managed here.
 * Wrapping of error handling is wired in by `src/utils/index.ts` exporting `src/telemetry/errors.ts#createError()`, which wraps `src/utilts/errors.ts`
```

I'm not saying this needs to be the exact wording but I think paints a clearer picture about this class' role.

packages/synapse-sdk/src/telemetry/singleton.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+  return null
+}
+
+/**
+ * Remove the global telemetry instance
+ *
+ * Useful for testing or when telemetry should be disabled.
+ */
+export function removeGlobalTelemetry(): void {
```
BigLep: In this case, shouldn't we remove global fetch wrapping?

packages/synapse-sdk/src/telemetry/singleton.ts
```diff
@@ -0,0 +1,52 @@
...
+  telemetryInstance = null
+}
+
+/**
+ * Check if global telemetry is enabled
+ *
+ * @returns True if telemetry is initialized and enabled
+ */
+export function isGlobalTelemetryEnabled(): boolean {
+  return telemetryInstance?.isEnabled() ?? false
```
BigLep: If we incorporate feedback elsewhere about the telemetryservice not magaing whether it's enabled or not (but rather having the singleton be aware) then this would need to change.

packages/synapse-sdk/src/telemetry/config.ts
BigLep: Do we just inline the relevant parts into the singletonn manager if it's in charge of determining whether global telemetry is enabled?

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+      // Enable tracing/performance monitoring
+      tracesSampleRate: 1.0, // Capture 100% of transactions for development (adjust in production)
+      // Integrations configured per-runtime in sentry-dep files
+      integrations,
+    })
+
+    // things that we don't need to search for in sentry UI, but may be useful for debugging should be set as context
+    this.sentry.setContext('environment', {
```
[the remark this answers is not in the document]
BigLep: Maybe have 

```
setContect("Runtime", {
   type : "browser" // or "node" from the current RuntimeContext.runtime
   userAgent : this.context.userAgent
}
```

packages/synapse-sdk/src/telemetry/get-sentry.ts
```diff
@@ -0,0 +1,27 @@
+import type * as SentryBrowser from '@sentry/browser'
+import type * as SentryNode from '@sentry/node'
+
+// Dynamically import the correct Sentry package based on environment
+const isBrowser =
+  typeof (globalThis as any).window !== 'undefined' && typeof (globalThis as any).document !== 'undefined'
+
+export type Sentry = typeof SentryNode.default | typeof SentryBrowser.default
+
+export async function getSentry(): Promise<{ Sentry: Sentry; integrations: any[] }> {
```
BigLep: Somewhere we want to callout that we're adding additional integrations to the defaults that Sentry enables per https://docs.sentry.io/platforms/javascript/configuration/integrations/

packages/synapse-sdk/src/telemetry/get-sentry.ts
```diff
@@ -0,0 +1,27 @@
...
+        SentryBrowser.browserTracingIntegration({
+          ignoreResourceSpans: ['resource.script', 'resource.img', 'resource.css', 'resource.link'],
+        }),
+      ],
+    }
+  }
+  const SentryNode = await import('@sentry/node')
+  return {
+    Sentry: SentryNode,
+    integrations: [SentryNode.httpIntegration()],
```
BigLep: Is this even needed?  It looks to be enabled by default per https://docs.sentry.io/platforms/javascript/guides/connect/configuration/integrations/http/

Although it's confusing that it isn't listed at https://docs.sentry.io/platforms/javascript/configuration/integrations/

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    return this.sentry.close(timeout)
+  }
+
+  /**
+   * Add event to circular buffer
+   * @internal
+   */
+  private addToBuffer(event: any): void {
```
BigLep:
```suggestion
  private addToEventBuffer(event: any): void {
```

to be a little more self describing?

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+interface RuntimeContext {
+  sdkVersion: string
+  runtime: 'browser' | 'node'
+  network: 'mainnet' | 'calibration'
+  userAgent?: string
+  appName?: string
+}
+
+export interface DebugDump {
+  lastEventId: string | undefined
```
BigLep: Why is this needed?

packages/synapse-sdk/src/telemetry/shutdown-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { Sentry as SentryType } from './get-sentry.ts'
+
+export function setupShutdownHooks(sentry: SentryType, opts: { timeoutMs?: number } = {}) {
+  const g = globalThis as any
+  const timeout = opts.timeoutMs ?? 2000
+  let shuttingDown = false
+
+  // NOTE: sentry handles uncaughtException and unhandledRejection. see https://docs.sentry.io/platforms/javascript/troubleshooting/#third-party-promise-libraries
+
+  if (typeof g.window !== 'undefined' && typeof g.document !== 'undefined') {
```
BigLep: I assume we have a more shorthand / self-documenting way at this point for knowing whether we're in a browser or not right?  Basically, we already have the right type of Sentry instance so we should already know whether we're handling a browser a not without having to these less readable checks.

packages/synapse-sdk/src/telemetry/shutdown-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    // -------- Browser runtime --------
+    const flush = () => {
```
BigLep: Why do we only flush in browsers vs. close?  Lets document that.  Is there any harm in just closing?

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+   * If not set, the SDK will use 'synapse-sdk' as the default app name.
+   */
+  appName?: string
+  tags?: Record<string, string> // optional: custom tags
+}
+
+/**
+ * Configuration for runtime detection and context
+ */
+interface RuntimeContext {
```
BigLep: I think if other feedback is incorporated, this goes away.

BigLep (commented):
I know my comments have come in incrementally in different reviews.  I know that isn't ideal.  I'm hoping this review comment can give some clarity.  

## Problems to solve
It took me awhile to grok what is getting setup where and why.  I think we can simplify.

_But isn't this telemetry temporary and we shouldn't worry too much about it?_
1. I suspect this telemetry code is going to live beyond getting to GA.  We'll of course always have ways to disable it, but it's going to be critical for us in being able to troubleshoot and improve the project.  
2. Synapse is the reference project for interfacing with Filecoin Onchain Cloud, which people will be encouraged to use, inspect, and learn from.  As a result, I think we need to hold any code we add to it to a high bar for flexing the muscle of insisting on high standards.

## High level thoughts/proposal
1. the TelemetryService instance gets created in singleton.ts.  singelton.ts only creates the instance if Telemetry should be enabled.
3. synapse.ts just calls `singleon.ts#setupGlobalTelemetry(options.telemetry)` after setting `options.telemetry.sentryTags.fileconNetwork = network` (since the Synapse constructor it is in a position to know what network is being used).  It doesn't do anything else.  It doesn't even store a reference to the TelemetryService.  All TelemetryService access is through singleton.ts.  (I think it would be great if we didn't need to use a singleton, but  I am assuming we do per "Why have all access of the TelemetryService through singelton.ts?" below).
4. singleton.ts is responsible for creating the TelemetryService, wrapping fetch, setting up shutdown hooks.  It's also response for tearing things down / unwrapping when directed.
5. As much as possible, we just bleed through the Sentry configuration options without our own wrappers.  That way we have less code and it's possible for someone to make configuration changes to change the behavior.  We just decorate the [Sentry configuration options](https://docs.sentry.io/platforms/javascript/configuration/options/) with good defaults if they haven't been explicitly set. 

## Why have all access of the TelemetryService through singelton.ts?
I think we're forced into that because of errors.ts.  It needs to get ahold of the TelemetryService instance, and the singleton accessors in singleton.ts is its means.

If there is a way for errors.ts to get the TelemetryService without a static accessor, then that would be great.  In that case, synapse.ts can create the TelemetryInstance and hold a reference.  

My thinking on priorities:
1. Be consistent on how TelemetryService can be accessed.  Either access through a static singleton method or through the Synapse instance, but lets not do both.
2. If it's possible to always access through the Synapse instance only, then lets do that and drop the Singleton static access.

## Do we need so many files under telemetry directory?
If we do make any of the structural proposals below, lets check if all the files in the directory still make sense, or if some of them should be combined.  I would need to see how the dust settles to have a stronger opinion.  Lets document though the responsibility of each .ts file so it's clear, and I think we should optimize for readability (which I realize is subjective).

## Side observation
When looking at https://filoz.sentry.io/explore/traces/trace/d16dbdb3d307d472145fea2ce986345d/?node=span-33e56d623b76f721&query=&source=traces&statsPeriod=14d&targetId=33e56d623b76f721&timestamp=1761338760, I'm curious why there so many HTTP requests that are breadcrumbs in the wrapper span.  I would have thought each of those HTTP request breadcrumb item would have triggered their own span, in which case any of the breadcrumbs would have been sent out earlier.  Basically I'm surprised we get so many HTTP related breadcrumbs within our HTTP wrapper span, which only mapps to a single HTTP requests.  

## Next steps
1. If my spew of written comments isn't clear, then lets align verbally the morning of 2025-10-27.  
2. I know more needs to be included in README.md.  I can draft this up 2025-10-27.

packages/synapse-sdk/src/telemetry/service.ts
SgtPooki: see [`07b6cbc` (#349)](https://github.com/FilOzone/synapse-sdk/pull/349/commits/07b6cbc711fa9a0e30a03f4b0b2727da7bcc9838)

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  constructor(config: TelemetryConfig, context: RuntimeContext) {
+    const resolvedConfig = resolveTelemetryConfig(config)
+    this.enabled = resolvedConfig.enabled
+    this.context = context
+
+    this.sentry = Sentry
+    if (this.enabled) {
+      this.initSentry()
+    }
```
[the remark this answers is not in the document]
SgtPooki: makes sense

packages/synapse-sdk/src/telemetry/config.ts
SgtPooki: moved to singleton.ts. I'm not a huge fan of large files and would much rather prefer tiny files with little helper methods, which is why these were all split out.

utils/example-piece-status.js
```diff
@@ -518,7 +519,9 @@ async function main() {
     if (error.cause) {
       console.error(`Caused by: ${error.cause.message}`)
     }
-    process.exit(1)
+    process.exitCode = 1
+  } finally {
+    synapse.telemetry.flush()
```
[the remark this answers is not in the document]
SgtPooki: I actually want to remove the changes to existing utils.. ideally, telemetry would emit without changing these, but flush was the original method.. I didn't have both flush and close when this was added.

packages/synapse-sdk/src/telemetry/get-sentry.ts
SgtPooki: removed

BigLep (commented):
Posting comments on singleton.ts after looking from our 2025-10-27 conversation.

packages/synapse-sdk/src/telemetry/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+import { createError as originalCreateError } from '../utils/errors.ts'
+import { getGlobalTelemetry } from './singleton.ts'
+
+// Dynamically import the correct Sentry package based on environment
+export const isBrowser =
+  typeof (globalThis as any).window !== 'undefined' && typeof (globalThis as any).document !== 'undefined'
+
+export type Sentry = typeof SentryNode.default | typeof SentryBrowser.default
+
+export async function getSentry(): Promise<{ Sentry: Sentry; integrations: any[] }> {
```
BigLep:
```suggestion
/**
 * Dynamically import the correct Sentry package for whether we're running in a browser or Node.
 */
export async function getSentry(): Promise<{ Sentry: Sentry}> {
```

(I'll leave a comment about this in service.ts but I think it makes more sent to have the integration definitions happen in service.ts rather than being tucked away here.)

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.context = context
+    this.config = config
+
+    // Initialize sentry always.. singleton.ts will not construct this service if telemetry is disabled.
+    void this.initSentry()
+  }
+
+  private async initSentry(): Promise<void> {
+    const { Sentry, integrations } = await getSentry()
+    this.sentry = Sentry
```
BigLep:
```suggestion
    this.sentry = await getSentry()
    if (isBrowser) {
      runtime = "browser"
      integrations = [
        // Disable telemetry on static asset retrieval.  
        // It's noisy and potentially more identifiable information.
        SentryBrowser.browserTracingIntegration({
          ignoreResourceSpans: ['resource.script', 'resource.img', 'resource.css', 'resource.link'],
        })
      ]
    } else {
       runtime = "node"
       integrations = []
    }
```

We also want to enable and configure certain [Sentry integrations](https://docs.sentry.io/platforms/javascript/configuration/integrations/), based on our runtime environment.

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.sentry.init({
+      dsn: 'https://3ed2ca5ff7067e58362dca65bcabd69c@o4510235322023936.ingest.us.sentry.io/4510235328184320',
+      // Setting this option to false will prevent the SDK from sending default PII data to Sentry.
+      // For example, automatic IP address collection on events
+      sendDefaultPii: false,
+      // Enable tracing/performance monitoring
+      tracesSampleRate: 1.0, // Capture 100% of transactions for development (adjust in production)
+      // Integrations configured per-runtime in sentry-dep files
```
BigLep:
```suggestion
```

Remove if we incorporate the suggestion above of defining integrations in this file.

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,150 @@
...
+ * You should use `synapse.telemetry.sentry` directly to capture events.
+ *
+ * ## Debug Dumps
+ *
+ * Get recent events for support tickets:
+ *
+ * ```typescript
+ * const dump = synapse.telemetry.debugDump()
+ * console.log(JSON.stringify(dump, null, 2))
+ * ```
```
BigLep:
```suggestion
 * TelemetryService - Main telemetry service for Synapse SDK.
 * Per [issue #328](https://github.com/FilOzone/synapse-sdk/issues/328) this is primarily a thin wrapper around sentry.io.
 * It allows a caller to pass through [Sentry configuration options], where Synapse will apply some defaults if they aren't otherwise set.
 * (See the constructor for more information.)
 * The underlying Sentry instance can be accessed via `.sentry` for invoking any other [Sentry APIs](https://docs.sentry.io/platforms/javascript/apis/).  
 *
 * In addition, to help with support tickets, the TelemetryService can be queried to get recent events:
 *
 * ```typescript
 * const dump = synapse.telemetry.debugDump()
 * console.log(JSON.stringify(dump, null, 2))
 * ```
```

packages/synapse-sdk/src/telemetry/service.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  constructor(config: TelemetryConfig, context: TelemetryRuntimeContext) {
+    this.context = context
+    this.config = config
+
+    // Initialize sentry always.. singleton.ts will not construct this service if telemetry is disabled.
+    void this.initSentry()
+  }
+
+  private async initSentry(): Promise<void> {
```
BigLep:
```suggestion
  // This is a separate function rather than being in the constructor because it is async.
  // This does means that a TelemetryService instance can be accessible without the sentry object being instantiated.
  // We are fine with this in practice because in the worst case it means some initial telemetry events get missed.
  // Consuming code of the Synapse telemetry module should be fine because it already protects against a null sentry instance in case telemetry is disabled. 
  private async initSentry(): Promise<void> {
```

Ignore if this is generally obvious.

packages/synapse-sdk/src/telemetry/singleton.ts
```diff
@@ -0,0 +1,262 @@
+/**
+ * Telemetry singleton manager.
+ * Sets up and provides a single global TelemetryService instance.
+ * #initGlobalTelemetry is the entry point.
+ * #getGlobalTelemetry is the expected access point within Synapse and beyond.
```
SgtPooki:
```suggestion
 * #getGlobalTelemetry is the expected access point within Synapse
```

we want consumers to access it (from outside synapse) with `synapse.telemetry` and should likely do similar if they have access to the synapse instance, instead of calling `getGlobalTelemetry`

packages/synapse-sdk/src/telemetry/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  // Capture to telemetry if enabled
+  getGlobalTelemetry()?.sentry?.captureException(error, {
+    tags: { operation: `${prefix}.${operation}` },
+    extra: {
+      synapseErrorPrefix: prefix,
+      synapseErrorOperation: operation,
+      synapseErrorDetails: details,
+      originalError,
+    },
+  })
```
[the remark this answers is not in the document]
SgtPooki: I was trying to make it easy to rip out in the future (delete telemetry folder, replace import, update synapse.ts, done). I guess it wouldnt be much harder to just remove it in the createError after `rm -rf src/telemetry` and seeing what build errors are.

packages/synapse-sdk/src/test/mocks/jsonrpc/index.ts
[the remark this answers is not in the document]
SgtPooki: we needed the changes here to get the `mainnet` tests passing in telemetry.test.ts. I'm fine to remove those tests if we want, or follow better patterns.

If I replaced `warmStorage` with the real contract address (which is what returns when I mock the way I do in mainnet tests) then other tests break because they're expecting the `0x1234567890123456789012345678901234567890` value, and I didn't want to go change a bunch of other things.. but I can.

How would you like me to resolve it?

1. remove mainnet telemetry test?
1. mock mainnet differently in telemetry.test.ts?
1. replace warmStorage address with the real one in jsonrpc mock?
1. fix other code that doesn't use the real address?

packages/synapse-sdk/src/telemetry/utils.ts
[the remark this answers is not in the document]
hugomrdias: yep and avoid changing the imports back to original createError

utils/simple-span-test.ts
```diff
@@ -0,0 +1,89 @@
...
+  })
+
+  const response = await makeRequest(1)
+  console.log('outside span', response)
+}
+
+// Run the test
+testSpanTest(process.env.RPC_URL || RPC_URLS.calibration.websocket)
+  .then(() => {
+    throw new Error('test error')
```
rvagg: this error is showing up in the collection because it's an unhandled?

BigLep:
> * can a dev use sentry in their app and not get into trouble because synapse already using sentry?

If a dev wants to use Sentry in their app, then they would need to configure Sentry through their app (pass configuration options into Sentry).  Alternatively, they need to disable Synapse Sentry and they their telemetry will work (and Synapse's will be excluded).  I will update the README to be clear about this.

> * this doesnt solve catching unhandled exceptions right ? its only going after the known errors the sdk already catches?

I believe we are using the Sentry defaults which will catch unhandled exceptions anywhere, not just the known erros that the SDK catches.  Is that a blocker for this pre-GA period?

BigLep:
> I will update the README to be clear about this.

README updated in https://github.com/FilOzone/synapse-sdk/pull/349/commits/0fbabc07fefceda34ae68922c728ab454b148074

Issue #343, opened 2025-10-23
Use random nonces for AddPieces


rvagg:
Adapt to https://github.com/FilOzone/filecoin-services/pull/317

The 712 payload is broken, so at a minimum we need to fix that and insert a nonce into there. The nonce should be random, but it could even be the same as we have now, but we have nice randomness for the CreateDataSet nonces so it shouldn't be a stretch.

PR #350, opened 2025-10-24
feat: use random nonce for AddPieces operations

rvagg:
- Replace sequential firstPieceId with random nonce in AddPieces signature
- Generate nonce using randU256() for replay protection
- Update AddPieces EIP-712 type: firstAdded → nonce
- Encode nonce as first parameter in addPieces extraData
- Replace getAddPiecesInfo() with validateDataSet() + getDataSet()
- Remove AddPiecesInfo interface (no longer needed)
- Update PDPServer.addPieces() to remove nextPieceId parameter
- Validate piece ID count matches batch size in status polling

Ref: https://github.com/FilOzone/filecoin-services/pull/317
Closes: https://github.com/FilOzone/synapse-sdk/issues/343


rvagg:
This does depend on https://github.com/filecoin-project/curio/pull/733

Kubuxu:
Delay merging this until new contract set.

packages/synapse-sdk/src/test/rand.test.ts
```diff
@@ -1,6 +1,7 @@
 /* globals describe it */
+
+import { fallbackRandIndex, fallbackRandU256, randIndex, randU256 } from '@filoz/synapse-core/rand'
```
rvagg: @hugomrdias can you tell me about what your intention is for testing the components in synapse-core? I see it has no tests at the moment, are you intending to add some or are we supposed to be testing it all from the outside? I had to move `rand` into there but now the unit tests for it are sitting over here, awkwardly.

packages/synapse-core/src/rand.ts
rvagg: @hugomrdias I rebased on top of your changes but I'm putting rand.ts into src/ and putting it as a top-level export as per my original rather than in utils/ so we don't have the same file in both places because we need it to make nonces in the sdk.

Issue #360, opened 2025-10-28
StorageInfo from getStorageInfo() should reflect new storage pricing


rvagg:
CDN being egress based and floor pricing.

rvagg:
Renaming this and using it as a parent for a bunch of related work

BigLep:
@rvagg : I put an estimate of 2 dev days.  Please adjust if that seems off.

Issue #358, opened 2025-10-28
"dev" selector should use `serviceStatus=dev` in SP capabilities list


rvagg:
Noting this here so it's remembered either as part of #356 or as a follow-up.

New capabilities list in the SP Registry doesn't allow empty keys and is now good at doing key=value fetching so we can afford to be more verbose. The current `dev` option in the SP selector should instead look for a `serviceStatus` key with value `dev`.

Then @Kubuxu's @ZenGround0 's and my SPs can be marked this way in the new registry.

rvagg:
This has been done in dealbot now

PR #361, opened 2025-10-28
feat: update to latest abi, including SP registry changes

rvagg:
This is an awkward mashup of the work @wjmelements is doing in #356 and some ABI update work I was doing as well.

It's likely not perfect, hasn't been properly tested, but _should_ match the latest filecoin-services ABI.

There's some follow-ups in here:

* https://github.com/FilOzone/synapse-sdk/issues/360 - service pricing is a mess in here, the main exported API needs to be updated
* https://github.com/FilOzone/synapse-sdk/issues/358 - I don't think `dev` will work as-is now, needs to be changed, capabilities need values
* There's a bunch of ignored `catch`s in the registry code, I think this is what @wjmelements was mentioning the other day? And I see in #356 that he was trying to undo them. I've left them in place here but added `console.warn` -- if we get a provider and can't decode it then we need to skip it. It could indicate an error in talking to the contract so maybe we need to be more careful with what we ignore and what we allow to propagate. I don't have time to think further about that unfortunately.

### DIFF
--- a/packages/synapse-sdk/package.json
+++ b/packages/synapse-sdk/package.json
@@ -73,6 +73,9 @@
       ],
       "output": [
         "dist/**"
+      ],
+      "dependencies": [
+        "../synapse-core:build"
       ]
     },
     "lint": {
--- a/packages/synapse-sdk/src/pdp/server.ts
+++ b/packages/synapse-sdk/src/pdp/server.ts
@@ -27,8 +27,8 @@
  */
 
 import * as Piece from '@filoz/synapse-core/piece'
-import { randU256 } from '@filoz/synapse-core/rand'
 import * as SP from '@filoz/synapse-core/sp'
+import { randU256 } from '@filoz/synapse-core/utils'
 import { ethers } from 'ethers'
 import type { Hex } from 'viem'
 import { asPieceCID, downloadAndValidate } from '../piece/index.ts'
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -22,7 +22,7 @@
  * ```
  */
 
-import { randIndex, randU256 } from '@filoz/synapse-core/rand'
+import { randIndex, randU256 } from '@filoz/synapse-core/utils'
 import type { ethers } from 'ethers'
 import type { PaymentsService } from '../payments/index.ts'
 import { PDPAuthHelper, PDPServer } from '../pdp/index.ts'
--- a/packages/synapse-sdk/src/test/rand.test.ts
+++ b/packages/synapse-sdk/src/test/rand.test.ts
@@ -1,6 +1,6 @@
 /* globals describe it */
 
-import { fallbackRandIndex, fallbackRandU256, randIndex, randU256 } from '@filoz/synapse-core/rand'
+import { fallbackRandIndex, fallbackRandU256, randIndex, randU256 } from '@filoz/synapse-core/utils'
 import { assert } from 'chai'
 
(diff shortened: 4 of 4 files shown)


Kubuxu:
I've created https://github.com/FilOzone/synapse-sdk/issues/362

packages/synapse-sdk/src/test/sp-registry-service.test.ts
```diff
@@ -434,22 +413,19 @@ describe('SPRegistryService', () => {
...
-            payee: mockProviderAddress,
-            name: 'Test Provider',
-            description: 'A test storage provider',
-            isActive: true,
-          },
-          product: {
-            productType: 0n,
-            capabilityKeys: [],
-            isActive: false,
+        getPDPService: async () => ({
```
wjmelements: This is wrong. There is no getPDPService method! Why did you revert my fix that switched to the new method `getProviderWithProduct`?

packages/synapse-sdk/src/test/sp-registry-service.test.ts
```diff
@@ -1,8 +1,6 @@
 /* globals describe it beforeEach */
 import { assert } from 'chai'
 import { ethers } from 'ethers'
-import type { Hex } from 'viem'
-import { boolToBytes, bytesToHex, numberToBytes, stringToHex } from 'viem'
```
wjmelements: bro!

packages/synapse-core/src/warm-storage/upload.ts
```diff
@@ -21,23 +22,33 @@ export async function upload(client: Client<Transport, Chain, Account>, options:
...
+    args: [dataSet.providerId, 0], // productType PDP = 0
   })
 
+  // Decode capabilities to get service URL
+  const capabilities = providerWithProduct.product.capabilityKeys.reduce(
+    (acc, key, i) => {
+      acc[key] = providerWithProduct.productCapabilityValues[i]
+      return acc
+    },
+    {} as Record<string, `0x${string}`>
```
wjmelements: use `Hex`!

packages/synapse-sdk/src/sp-registry/service.ts
```diff
@@ -108,16 +107,33 @@ export class SPRegistryService {
     const registrationFee = await contract.REGISTRATION_FEE()
 
     // Prepare product data and capabilities
-    let productType = 0 // ProductType.PDP
-
-    const [ capabilityKeys, capabilityValues ] = this._convertPDPOfferingToCapabilities(info.pdpOffering, info.capabilities)
```
wjmelements: bro!

packages/synapse-sdk/src/test/sp-registry-service.test.ts
```diff
@@ -389,10 +368,10 @@ describe('SPRegistryService', () => {
         ipniPiece: true,
         ipniIpfs: true,
         storagePricePerTibPerDay: BigInt(2000000),
-        minProvingPeriodInEpochs: 2880,
+        minProvingPeriodInEpochs: 2880n,
         location: 'EU-WEST',
-        paymentTokenAddress: '0x0000000000000000000000000000000000000000',
-      } as const
+        paymentTokenAddress: '0x0000000000000000000000000000000000000000' as `0x${string}`,
```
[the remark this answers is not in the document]
wjmelements: I don't think this type assertion is required (as it is for ethers.ZeroAddress), because it's a literal.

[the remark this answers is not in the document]
wjmelements: another fix is to use `as const` for the object

PR #364, opened 2025-10-28
feat(ServiceProviderRegistry): support latest ABI

wjmelements:
Reviewer @rvagg
This partially reverts d70899d48baa361536622ec84d59e4f2cb037855, mostly in sp-registry/service and sp-registry-service.test.
#### Changes
* prefer Hex to `0x{string}`
* restore product encoding accidentally reverted by d70899d48baa361536622ec84d59e4f2cb037855
* add bool parameter to `getProvidersByProductType`
* fix tests

### DIFF
--- a/packages/synapse-sdk/src/test/storage.test.ts
+++ b/packages/synapse-sdk/src/test/storage.test.ts
@@ -1,10 +1,9 @@
 /* globals describe it beforeEach afterEach */
 import { assert } from 'chai'
 import { ethers } from 'ethers'
-import type { Hex } from 'viem'
 import { StorageContext } from '../storage/context.ts'
 import type { Synapse } from '../synapse.ts'
-import type { PieceCID, ProviderInfo, UploadResult } from '../types.ts'
+import type { PieceCID, ProviderInfo } from '../types.ts'
 import { SIZE_CONSTANTS } from '../utils/constants.ts'
 import { createMockProviderInfo, createSimpleProvider, setupProviderRegistryMocks } from './test-utils.ts'
 


packages/synapse-sdk/src/sp-registry/service.ts
```diff
@@ -429,30 +404,22 @@ export class SPRegistryService {
...
       const result = await contract.getProviderWithProduct(providerId, 0) // 0 = ProductType.PDP
 
-      // Check if provider exists
-      if (result.providerInfo.serviceProvider === ethers.ZeroAddress) {
-        return null
-      }
-
-      // Decode capability keys and values
-      const keys = Array.from(result.product.capabilityKeys) as string[]
-      const values = Array.from(result.productCapabilityValues) as string[]
```
wjmelements: `productCapabilityValues` is already `Hex[]`

packages/synapse-core/src/session-key/permissions.ts
```diff
@@ -1,5 +1,5 @@
 import { TypedData } from 'ox'
-import { keccak256, stringToHex } from 'viem'
+import { type Hex, keccak256, stringToHex } from 'viem'
```
rvagg: type imports need to be separate I think for the linter, currently linter isn't running on synapse-core in GHA but this will become a problem later I think, pull this up to a separate import

packages/synapse-core/src/warm-storage/providers.ts
```diff
@@ -1,5 +1,5 @@
 import type { AbiParametersToPrimitiveTypes, ExtractAbiFunction } from 'abitype'
-import { type Chain, type Client, hexToBigInt, hexToString, type Transport } from 'viem'
+import { type Chain, type Client, type Hex, hexToBigInt, hexToString, type Transport } from 'viem'
```
rvagg: separate type import

packages/synapse-sdk/src/sp-registry/service.ts
```diff
@@ -548,82 +507,38 @@ export class SPRegistryService {
...
       }
 
       try {
-        const decoded = iface.decodeFunctionResult('getProvider', results[providerResultIndex].returnData)
-        const rawProvider = decoded[0]
-
-        // Process PDP service if available
-        const products = this._extractProductsFromMulticallResult(results[pdpServiceResultIndex], iface)
+        const [, rawProvider, product, productCapabilityValues] = iface.decodeFunctionResult(
+          'getProviderWithProduct',
```
rvagg: nice 👌

packages/synapse-sdk/src/sp-registry/service.ts
```diff
@@ -22,8 +22,9 @@
  * ```
  */
 
-import { decodeCapabilities } from '@filoz/synapse-core/warm-storage'
```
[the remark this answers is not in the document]
rvagg: looking at this now I think maybe we shouldn't put them in `warm-storage` but they should be in an `sp-registry` export since this isn't warm-storage specific

not a blocker for this PR but if we don't do it now, let's open an issue and follow up later

packages/synapse-core/src/utils/decode-pdp-errors.ts
```diff
@@ -1,5 +1,5 @@
 import type { AbiError } from 'abitype'
-import { AbiErrorSignatureNotFoundError, decodeErrorResult } from 'viem'
+import { AbiErrorSignatureNotFoundError, decodeErrorResult, type Hex } from 'viem'
```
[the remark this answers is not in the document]
wjmelements: f145461

packages/synapse-sdk/wagmi.config.ts
```diff
@@ -3,7 +3,7 @@ import { fetch } from '@wagmi/cli/plugins'
 import type { Address } from 'viem'
 
 // GIT_REF can be one of: '<branch name>', '<commit>' or 'tags/<tag>'
-const GIT_REF = '43e122981042a1498fc642d51376e8b70a760161'
+const GIT_REF = '8e162d676f3e83c495f104989b3014b3961e2f05'
```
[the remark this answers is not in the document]
rjan90: I think we would only need to upgrade the `ServiceProviderRegistry` contract. Issue is that the [upgrade-script in filecoin-services](https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/tools/upgrade.sh), only deals with the FilecoinWarmStorageService contracts.

And the only other option would then be to use the [deploy-all-warm-storage](https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/tools/deploy-all-warm-storage.sh) script, but reuse everything else apart from the ServiceRegistryContract. But that would create entirely new proxy and implementation addresses, and that would create additional work for Kuba (and our E2E testing effort), as he would need to reset and get his Curio node accepted in the registry again.

packages/synapse-core/src/warm-storage/data-sets.ts
```diff
@@ -82,7 +82,9 @@ export async function getDataSets(client: Client<Transport, Chain>, options: Get
     })
 
     // getProviderWithProduct returns {providerId, providerInfo, product, productCapabilityValues}
-    const pdpCaps = decodeCapabilities(pdpOffering.product.capabilityKeys, pdpOffering.productCapabilityValues)
+    const pdpCaps = decodePDPCapabilities(
+      capabilitiesListToObject(pdpOffering.product.capabilityKeys, pdpOffering.productCapabilityValues)
+    )
```
hugomrdias: instead of this can we just call decodePDPCapabilities with keys and values and call capabilitiesListToObject inside?

wjmelements: yes that is the part I understood. I was asking about the second part.

wjmelements: The reason not to do this is because in synapse-sdk we need the capabilities object too.

packages/synapse-sdk/src/sp-registry/service.ts
```diff
@@ -429,30 +404,19 @@ export class SPRegistryService {
...
 
-      // Decode capabilities into PDPOffering
-      const offering = decodeCapabilities(keys, values as readonly `0x${string}`[])
-      const capabilities = this._convertCapabilitiesToObject(keys, values)
+      const capabilities = capabilitiesListToObject(result.product.capabilityKeys, result.productCapabilityValues)
 
       return {
-        offering,
+        offering: decodePDPCapabilities(capabilities),
         capabilities,
```
wjmelements: @hugomrdias here

hugomrdias (approved):

Issue #340, opened 2025-10-22
Remove approval requirement for provider selection


rvagg:
I only just discovered this today, `git grep 'provider.*not.*approved'`.

As of https://github.com/FilOzone/filecoin-services/pull/302 this should no longer be a concern of the SDK. We should _use_ the approved list to find providers, but we shouldn't block on the user providing an ID or address that's not in the approved list.

I _think_ it's just a matter of removing some `providerResolver.getApprovedProvider*` calls.

PR #369, opened 2025-10-30
feat: remove approvedProvider gating for use

rvagg:
* FWSS approvedProvier list is just advisory now and used for Synapse SP selection
* Where providerId or providerAddress is provided, approvedProvider isn't checked for permission
* Removed ProviderResolver, simplified lots of code

Fixes: https://github.com/FilOzone/synapse-sdk/issues/340


wjmelements:
We still enforce approval for random selection, even if there are no approved providers
```
    // No existing data sets - select from all approved providers. First we get approved IDs from
    // WarmStorage, then fetch provider details.
    const approvedIds = await warmStorageService.getApprovedProviderIds()
    const approvedProviders = await spRegistry.getProviders(approvedIds)
    const allProviders = approvedProviders.filter(
      (provider: ProviderInfo) => excludeProviderIds?.includes(provider.id) !== true
    )

    if (allProviders.length === 0) {
      throw createError('StorageContext', 'smartSelectProvider', 'No approved service providers available')
    }
```

PR #375, opened 2025-10-31
refactor(sdk): ServiceInfo rename, accurate pricing separation, reusable primitives

rvagg:
Two commits in here. I was originally solving for #360 and then the deeper I got the worse I realised it was. We have a lot of CDN stuff still in there, floor pricing impacts a lot of things, there's mismatch in arguments and return types. So I've tried to be consistent; I've removed "perDay" and "perEpoch" from a bunch of places, leaving just the standardised "perMonth" we do everywhere else.

I'm hoping that this will be able to replace a bunch of things in filecoin-pin. @SgtPooki when you have a spare moment would you mind having a quick think about the coverage of functionality we have here? filecoin-pin is currently not dealing with the floor price, although it is updated to the 30 day lockup. I've also pulled in all of the detailed service readiness checks in here.

---

refactor(sdk): decompose preflight into reusable primitives

- Add PaymentsService.checkServiceReadiness() for any service
- Rename calculateStorageCost() to calculateUploadCost(), expose floor pricing
- Simplify prepareStorageUpload() signature (dataSize only, monthly cost only)
- Remove checkAllowanceForStorage()
- Add getContractAddress(), export ServiceReadinessCheck type

refactor(sdk): rename StorageInfo → ServiceInfo, restructure pricing

- Rename getStorageInfo() → getServiceInfo()
- Restructure pricing fields for clarity
- Simplify allowance checks

Fixes: https://github.com/FilOzone/synapse-sdk/issues/360

---

So we can now get this kind of information from Synapse, this is from example-storage-info.js with my wallet that's not set up for current contracts:

```
--- Pricing Information ---
Token: USDFC (0xb3042734b608a1B16e9e86B374A3f3e389B4cDf0)

Base Storage:
  Per TiB per month:        2.500000 USDFC
  Minimum price per month:  0.060000 USDFC

CDN Usage-Based Pricing (only for data sets with CDN enabled):
  CDN egress per TiB:       7.000000 USDFC
  Cache miss egress per TiB: 7.000000 USDFC

--- Service Providers ---
Total providers: 2

Provider #2:
  Name:        ezpdpz-calib2
  Description: PDP Calibration node based in the UK
  Address:     0xbCdf1bdc1a97D071a5a8EF03F1F05225b6E2a1Ba
  Payee:       0xbCdf1bdc1a97D071a5a8EF03F1F05225b6E2a1Ba
  Active:      true
  Service URL: https://calib2.ezpdpz.net
  PDP Service:
    Min size:  1 MB
    Max size:  1 GB
    Price:     0.000000 USDFC/TiB/month
    Location:  C=GB;ST=Gloucestershire;L=Cheltenham

Provider #1:
  Name:        Kubuxu's dev node
  Description: Do not rely on it at all
  Address:     0x8c8c7a9BE47ed491B33B941fBc0276BD2ec25E7e
  Payee:       0x8c8c7a9BE47ed491B33B941fBc0276BD2ec25E7e
  Active:      true
  Service URL: https://pdp-dev.kubuxu.com
  PDP Service:
    Min size:  1 MB
    Max size:  64 GB
    Price:     0.000000 USDFC/TiB/month
    Location:  C=FI;ST=Uusimaa;L=Helsinki

--- Service Parameters ---
Network:          calibration
Epochs per month: 86,400
Epochs per day:   2,880
Epoch duration:   30 seconds
Min upload size:  127 Bytes
Max upload size:  200 MB

Contract Addresses:
  Warm Storage:                  0x02925630df557F957f70E112bA06e50965417CA0
  Payments:                      0x09a0fDc2723fAd1A7b8e3e00eE5DF73841df55a0
  PDP Verifier:                  0x85e366Cf9DD2c0aE37E963d9556F5f4718d6417C
  Service Provider Registry:     0x839e5c9988e4e9977d40708d0094103c0839Ac9D
  Session Key Registry:          0x97Dd879F5a97A8c761B94746d7F5cfF50AAd4452

--- Current Allowances ---
Service: 0x02925630df557F957f70E112bA06e50965417CA0

Rate:
  Allowance:  0.000000 USDFC
  Used:       0.000000 USDFC
  Available:  0.000000 USDFC

Lockup:
  Allowance:  0.000000 USDFC
  Used:       0.000000 USDFC
  Available:  0.000000 USDFC

--- Account Balance ---
Total funds:      1.000000 USDFC
Available funds:  1.000000 USDFC
Locked up:        0.000000 USDFC

--- Upload Cost Examples ---

1 MiB:
  Base monthly cost:  0.000002 USDFC
  With floor pricing: 0.060000 USDFC

100 MiB:
  Base monthly cost:  0.000238 USDFC
  With floor pricing: 0.060000 USDFC

1 GiB:
  Base monthly cost:  0.002441 USDFC
  With floor pricing: 0.060000 USDFC

10 GiB:
  Base monthly cost:  0.024414 USDFC
  With floor pricing: 0.060000 USDFC

--- Wallet Readiness Check ---
Checking readiness for 1 GiB upload (0.060000 USDFC/month)...

Readiness: ❌ NOT READY

Checks:
  ✓ Operator approved:     ❌
  ✓ Sufficient funds:      ✅
  ✓ Rate allowance:        ❌
  ✓ Lockup allowance:      ❌
  ✓ Valid lockup period:   ❌

Required actions:
  - Increase rate allowance by 0.000001 USDFC
  - Increase lockup allowance by 0.060000 USDFC
  - Extend lockup period by 86400 epochs
```

### DIFF
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -54,6 +54,8 @@ import {
   getCurrentEpoch,
   METADATA_KEYS,
   SIZE_CONSTANTS,
+  TIME_CONSTANTS,
+  TOKENS,
   timeUntilEpoch,
 } from '../utils/index.ts'
 import { combineMetadata, metadataMatches, objectToEntries, validatePieceMetadata } from '../utils/metadata.ts'
@@ -785,34 +787,70 @@ export class StorageContext {
 
   /**
    * Static method to perform preflight checks for an upload
-   * @param size - The size of data to upload in bytes
-   * @param withCDN - Whether CDN is enabled
+   * Composes cost calculation from WarmStorageService and readiness checks from PaymentsService
    * @param warmStorageService - WarmStorageService instance
    * @param paymentsService - PaymentsService instance
+   * @param size - The size of data to upload in bytes
    * @returns Preflight check results without provider/dataSet specifics
    */
   static async performPreflightCheck(
     warmStorageService: WarmStorageService,
     paymentsService: PaymentsService,
-    size: number,
-    withCDN: boolean
+    size: number
   ): Promise<PreflightInfo> {
     // Validate size before proceeding
     StorageContext.validateRawSize(size, 'preflightUpload')
 
-    // Check allowances and get costs in a single call
-    const allowanceCheck = await warmStorageService.checkAllowanceForStorage(size, withCDN, paymentsService)
+    // Calculate upload cost
+    const cost = await warmStorageService.calculateUploadCost(size)
 
-    // Return preflight info
-    return {
-      estimatedCost: {
-        perEpoch: allowanceCheck.costs.perEpoch,
-        perDay: allowanceCheck.costs.perDay,
-        perMonth: allowanceCheck.costs.perMonth,
+    // Calculate rate per epoch from floor-adjusted monthly price
+    const pricing = await warmStorageService.getServicePrice()
+    const ratePerEpoch = cost.withFloorPerMonth / pricing.epochsPerMonth
+
+    // Calculate lockup requirements
+    const lockupEpochs = BigInt(TIME_CONSTANTS.DEFAULT_LOCKUP_DAYS * TIME_CONSTANTS.EPOCHS_PER_DAY)
+    const lockupNeeded = ratePerEpoch * lockupEpochs
+
+    // Check payment readiness
+    const readiness = await paymentsService.checkServiceReadiness(
+      warmStorageService.getContractAddress(),
+      {
+        rateNeeded: ratePerEpoch,
+        lockupNeeded,
+        lockupPeriodNeeded: lockupEpochs,
       },
+      TOKENS.USDFC
+    )
+
+    // Build error message if not sufficient
+    let message: string | undefined
+    if (!readiness.sufficient) {
+      const issues: string[] = []
+      if (!readiness.checks.hasSufficientFunds && readiness.gaps?.fundsNeeded) {
+        issues.push(`Insufficient funds: need ${readiness.gaps.fundsNeeded} more`)
+      }
+      if (!readiness.checks.isOperatorApproved) {
+        issues.push('Operator not approved for service')
+      }
+      if (!readiness.checks.hasRateAllowance && readiness.gaps?.rateAllowanceNeeded) {
+        issues.push(`Insufficient rate allowance: need ${readiness.gaps.rateAllowanceNeeded} more`)
+      }
+      if (!readiness.checks.hasLockupAllowance && readiness.gaps?.lockupAllowanceNeeded) {
+        issues.push(`Insufficient lockup allowance: need ${readiness.gaps.lockupAllowanceNeeded} more`)
+      }
+      if (!readiness.checks.hasValidLockupPeriod && readiness.gaps?.lockupPeriodNeeded) {
+        issues.push(`Lockup period too short: need ${readiness.gaps.lockupPeriodNeeded} more epochs`)
+      }
+      message = issues.join('; ')
+    }
+
+    return {
+      estimatedCostPerMonth: cost.withFloorPerMonth,
       allowanceCheck: {
-        sufficient: allowanceCheck.sufficient,
-        message: allowanceCheck.message,
+        sufficient: readiness.sufficient,
+        message,
+        checks: readiness.checks,
       },
       selectedProvider: null,
       selectedDataSetId: null,
@@ -829,8 +867,7 @@ export class StorageContext {
     const preflightResult = await StorageContext.performPreflightCheck(
       this._warmStorageService,
       this._synapse.payments,
-      size,
-      this._withCDN
+      size
     )
 
     // Return preflight info with provider and dataSet specifics
--- a/packages/synapse-sdk/src/warm-storage/service.ts
+++ b/packages/synapse-sdk/src/warm-storage/service.ts
@@ -228,6 +228,13 @@ export class WarmStorageService {
   }
 
   /**
+   * Get the Warm Storage contract address
+   */
+  getContractAddress(): string {
+    return this._warmStorageAddress
+  }
+
+  /**
    * Get the provider instance
    * @returns The ethers provider
    */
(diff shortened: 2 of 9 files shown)


SgtPooki (commented):
There are a few things here that will help reduce specifics in filecoin-pin: 

1. **Readiness checks** — `checkServiceReadiness()` can replace parts of our custom `validatePaymentCapacity()` and `checkUploadReadiness()` logic. We'll still need the runway-aware checks below.

1. **Cost calculation** — `calculateUploadCost()` with floor pricing will replace our `calculateRequiredAllowances()` calls, once we integrate floor pricing.

**What is missing that filecoin-pin needs (runway management)**:

filecoin-pin adds runway/duration management not covered here:

1. **Runway calculations** — `computeTopUpForDuration()` computes how many days current funding covers based on `rateUsed` + `lockupUsed` + balance.
2. **Adjustment calculations** — `computeAdjustmentForExactDays()` / `computeAdjustmentForExactDaysWithPiece()` calculate exact deposits needed to maintain N days runway (including after adding new content).
1. **Top-up orchestration** — `calculateRequiredTopUp()` combines piece upload needs + runway requirements with reason codes.

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -683,34 +685,70 @@ export class StorageContext {
...
-        perEpoch: allowanceCheck.costs.perEpoch,
-        perDay: allowanceCheck.costs.perDay,
-        perMonth: allowanceCheck.costs.perMonth,
+    // Calculate rate per epoch from floor-adjusted monthly price
+    const pricing = await warmStorageService.getServicePrice()
+    const ratePerEpoch = cost.withFloorPerMonth / pricing.epochsPerMonth
+
+    // Calculate lockup requirements
+    const lockupEpochs = BigInt(TIME_CONSTANTS.DEFAULT_LOCKUP_DAYS * TIME_CONSTANTS.EPOCHS_PER_DAY)
+    const lockupNeeded = ratePerEpoch * lockupEpochs
```
hugomrdias: really dont like this per month thing its inaccurate and forces us to do this back and forth stuff epoch<>month exchange.

also we just called getServicePrice in calculateUploadCost and we call it again just to get

https://github.com/FilOzone/filecoin-services/blob/980e25aeb920bb894eab71a3dba9e8c9bccddcae/service_contracts/src/FilecoinWarmStorageService.sol#L200

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -895,14 +797,25 @@ export class WarmStorageService {
...
-    ])
-
-    // Select the appropriate costs based on CDN option
-    const selectedCosts = (options.withCDN ?? false) ? costs.withCDN : costs
+    // Calculate upload cost
+    const cost = await this.calculateUploadCost(dataSize)
+
+    // Calculate requirements
+    const pricing = await this.getServicePrice()
+    const ratePerEpoch = cost.withFloorPerMonth / pricing.epochsPerMonth
```
hugomrdias: again calculateUploadCost and getServicePrice just for pricing.epochsPerMonth

hugomrdias (changes requested):
This all seems very messy, i think we are trying to shoehorn the old methods into a system that changed a bit (plus cdn egress that i dont understand).
We should redesign from first principles a better api.

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -874,17 +783,10 @@ export class WarmStorageService {
    * ```
    */
   async prepareStorageUpload(
```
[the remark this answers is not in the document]
rvagg: I didn't know about this CDN thing, it is kind of important. I've opened an issue for the contract here: https://github.com/FilOzone/filecoin-services/issues/339 and I'll consider it here as well.

packages/synapse-sdk/src/storage/context.ts
rvagg: point taken on the double call to `getServicePrice`, could just do that once.

> really dont like this per month thing

but the problem is twofold:

1. the contract defines pricing per month
2. downscaling to per-epoch loses a lot of precision (a surprising amount since we're dealing with fairly low numbers in the first place)

the logic here is intentionally keeping things at per-month for as long as possible to avoid the early precision loss of going per-epoch and then back again.

rvagg:
> We should redesign from first principles a better api.

Yeah, I agree .. the more I look at this the more I hate it from a consumption perspective. A lot of this isn't as helpful as it should be.

From the experience with Filecoin Pin (and what we're trying to do with https://github.com/filecoin-project/filecoin-pin/issues/76, and also see https://github.com/filecoin-project/filecoin-pin/pull/202#issuecomment-3483826121) is solve for: _do one transaction (if required) that sets me up to use the system_. Taken further, we when we add storage to that, we should also be able to make it: _do one transaction (if required) that sets me up to perform this upload_.

The combination of `deposit` and `depositWithPermitAndApproveOperator` I think gives us just one transaction in each of the necessary cases. And we also have a bunch of failure conditions where we can't perform that operation (wallet not found, don't have USDFC, don't have enough USDFC).

How about we do this: strip out all of this complexity and replace it with a single method: `synapse.storage.prepare(): Transaction` (could be something more helpful like: `{ type: ['deposit', 'deposit-and-approve-operator', 'approve-operator'], Transaction }`). It throws in the cases it can't solve, but where it can solve it it's allowed to submit one transaction. It can do the calculations you need based on what the system requires, and what your input requirements are and this can solve for. We take the various options we've added into Filecoin Pin wrt runway calculation and add them here. We also allow for proving a list of contexts (multi-context style, for multi-uploads) so it can perform calculations based on what you already have in those data sets and get you the runway you need.

Basically we want to take away the pain of figuring out all of this stuff for practical use: _Set me up to store this thing, on top of what I'm already storing, and I want it to be paid for this far into the future._

hugomrdias:
storage.prepare sounds good! what about current costs and current prepared upload costs?

hugomrdias:
so prepare would run checks and return the costs for the current upload?

i chatted with filbeam and:
- with cdn enable we need to make sure 1 usdfc is available in pay just for filbeam
- for current overall cost it will be harder because theres no aggregation for this anywhere
- they will PR a egress available per dataset (2 rails) method and we will go from there

rjan90:
It looks like this PR is quite outdated at this point? Is this PR still worth keeping around, or should it be closed?

PR #376, opened 2025-10-31
fix: use `serviceStatus=dev` for dev provider filtering

rjan90:
Closes: #358 

Updates the dev provider selector to use `serviceStatus=dev` in the SP capabilities list instead of checking for an empty `dev` key. This change aligns with the new ServiceProvider Registry capabilities format that requires proper key-value pairs. 

Providers can now be marked as dev by including `serviceStatus: 'dev'` in their capabilities, matching the implementation in dealbot and enabling proper filtering in the updated registry system.

### DIFF
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -643,11 +643,17 @@ export class StorageContext {
             continue
           }
 
-          if (!dev && provider.products.PDP?.capabilities?.serviceStatus === '0x646576') {
+          const serviceStatus = provider.products.PDP?.capabilities?.serviceStatus
+          console.log(
+            `[DEBUG] Provider ${provider.id}: dev=${dev}, serviceStatus=${serviceStatus}, typeof=${typeof serviceStatus}`
+          )
+          if (!dev && serviceStatus === '0x646576') {
             // "dev" in hex
+            console.log(`[DEBUG] Filtering out dev provider ${provider.id}`)
             continue
           }
 
+          console.log(`[DEBUG] Yielding provider ${provider.id}`)
           yield provider
         }
       }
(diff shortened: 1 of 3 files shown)


packages/synapse-sdk/src/storage/context.ts
```diff
@@ -638,16 +638,41 @@ export class StorageContext {
...
+                .match(/.{1,2}/g)
+                ?.map((byte) => parseInt(byte, 16)) ?? []
+            )
+            serviceStatus = new TextDecoder().decode(hexBytes).replace(/\0/g, '')
+          } else {
+            serviceStatus = serviceStatusHex
+          }
+          if (serviceStatus === 'dev') {
+            continue
+          }
```
rvagg: I think this is all correct, but the right way to do this is to just compare it to the hex.

```suggestion
        if (serviceStatusHex === '0x646576') { // "dev" has hex
          continue
        }
```

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -638,16 +638,41 @@ export class StorageContext {
...
+          let serviceStatus: string
+          if (typeof serviceStatusHex === 'string' && serviceStatusHex.startsWith('0x')) {
+            // Decode hex to string using web standard APIs
+            const hexBytes = new Uint8Array(
+              serviceStatusHex
+                .slice(2)
+                .match(/.{1,2}/g)
+                ?.map((byte) => parseInt(byte, 16)) ?? []
+            )
+            serviceStatus = new TextDecoder().decode(hexBytes).replace(/\0/g, '')
```
rvagg: this is an interesting inclusion, although I suspect for an edge case that doesn't exist? I think we can rely on it being strictly "dev" as a hex string of the correct length when it comes out of the capabilities decode

rvagg (approved):
I think this is ok, we should make sure we actually test this live before assuming it works and come back and fix it if not

rjan90:
Rebased this on master to resolve the conflict.

Issue #388, opened 2025-11-03
perf: `selectProviderWithPing` should use `Promise.race`


wjmelements:
`selectProvidersWithPing` should use `Promise.race` instead of trying them all sequentially.

_Originally posted by @wjmelements in https://github.com/FilOzone/synapse-sdk/pull/368#discussion_r2488025782_

This is a first step toward parallelizing `createContexts`. It would also speed up `createContext`. Lastly, the user would more often get a better storage provider.

wjmelements:
Consequently, fallback provider selection would not be random anymore.

PR #390, opened 2025-11-04
perf: Select Storage Provider by a ping race

wjmelements:
Reviewer @rvagg @hugomrdias
Closes https://github.com/FilOzone/synapse-sdk/issues/388
This makes progress toward a parallel `createContexts`.
By pinging in parallel rather than sequentially, the performance of `createContext` and `createContexts` should be dramatically increased, especially in the case where many providers are down.
Ties among providers are resolved according to which can ping first.
The prior priority tiers, such as existing pieces and existing data sets, are preserved.
#### Changes
* destroy async generator
* Promise.race the pings
* fix tests

### DIFF
--- a/packages/synapse-core/src/utils/rand.ts
+++ b/packages/synapse-core/src/utils/rand.ts
@@ -30,18 +30,3 @@ export function randU256(): bigint {
 export function fallbackRandIndex(length: number): number {
   return Math.floor(Math.random() * length)
 }
-
-/**
- * Provides a random index into an array of supplied length (0 <= index < length)
- * @param length - exclusive upper boundary
- * @returns a valid index
- */
-export function randIndex(length: number): number {
-  if (crypto?.getRandomValues != null) {
-    const randomBytes = new Uint32Array(1)
-    crypto.getRandomValues(randomBytes)
-    return randomBytes[0] % length
-  } else {
-    return fallbackRandIndex(length)
-  }
-}
--- a/packages/synapse-sdk/src/test/rand.test.ts
+++ b/packages/synapse-sdk/src/test/rand.test.ts
@@ -1,52 +1,49 @@
 /* globals describe it */
 
-import { fallbackRandIndex, fallbackRandU256, randIndex, randU256 } from '@filoz/synapse-core/utils'
+import { fallbackRandIndex, fallbackRandU256, randU256 } from '@filoz/synapse-core/utils'
 import { assert } from 'chai'
 
-const randIndexMethods = [randIndex, fallbackRandIndex]
-randIndexMethods.forEach((randIndexMethod) => {
-  describe(randIndexMethod.name, () => {
-    it('should return 0 for length 1', () => {
-      for (let i = 0; i < 32; i++) {
-        assert.equal(0, randIndexMethod(1))
+describe('fallbackRandIndex', () => {
+  it('should return 0 for length 1', () => {
+    for (let i = 0; i < 32; i++) {
+      assert.equal(0, fallbackRandIndex(1))
+    }
+  })
+  it('returns both 0 and 1 for length 2', () => {
+    const counts = [0, 0]
+    for (let i = 0; i < 32; i++) {
+      counts[fallbackRandIndex(counts.length)]++
+    }
+    // this test can fail probabilistically but the probability is low
+    // each bit should be independent with 50% likelihood
+    // the probability of getting the same index N times is 2**(1-N)
+    // so if this test fails, the 50% assumption is likely wrong
+    assert.isAtLeast(counts[0], 1)
+    assert.isAtLeast(counts[1], 1)
+  })
+  it('has at least 10 random bits', () => {
+    const counts = []
+    for (let i = 0; i < 10; i++) {
+      counts.push([0, 0])
+    }
+    for (let i = 0; i < 32; i++) {
+      let index = fallbackRandIndex(1024)
+      assert.isAtLeast(index, 0)
+      assert.isAtMost(index, 1023)
+      for (let j = 0; j < 10; j++) {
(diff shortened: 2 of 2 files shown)


packages/synapse-sdk/src/test/storage.test.ts
```diff
@@ -2463,11 +2461,8 @@ describe('StorageService', () => {
         }
 
         try {
-          await (StorageContext as any).selectRandomProvider(testProviders)
-          assert.fail('Should have thrown error')
-        } catch (error: any) {
-          assert.include(error.message, 'StorageContext selectProviderWithPing failed')
-          assert.include(error.message, 'All 2 providers failed health check')
+          const provider = await (StorageContext as any).selectProviderWithPing(testProviders)
+          assert.isNull(provider)
```
wjmelements: I prefer returning null over throwing and catching because catching can hide other errors

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -615,48 +615,38 @@ export class StorageContext {
...
+      // Prefer data sets with pieces
+      const [hasNoPieces, hasPieces] = managedDataSets
+        .reduce<[Set<EnhancedDataSetInfo>, Set<EnhancedDataSetInfo>]>(
+          (results: [Set<EnhancedDataSetInfo>, Set<EnhancedDataSetInfo>], managedDataSet: EnhancedDataSetInfo) => {
+            results[managedDataSet.currentPieceCount > 0 ? 1 : 0].add(managedDataSet)
+            return results
+          },
+          [new Set(), new Set()]
+        )
+        .map((deduped) => [...deduped])
```
rvagg: does `createContexts` land us with duplicates at this point now? why are we worried about deduping with Sets?

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -615,48 +615,38 @@ export class StorageContext {
...
 
-          if (withIpni && provider.products.PDP?.data.ipniIpfs === false) {
-            continue
-          }
+      for (const managedDataSets of [hasPieces, hasNoPieces]) {
+        const providers: ProviderInfo[] = (
+          await Promise.all(
+            managedDataSets.map((dataSet: EnhancedDataSetInfo) => spRegistry.getProvider(dataSet.providerId))
```
rvagg:
```suggestion
      for (const dataSets of [hasPieces, hasNoPieces]) {
        const providers: ProviderInfo[] = (
          await Promise.all(
            dataSets.map((dataSet: EnhancedDataSetInfo) => spRegistry.getProvider(dataSet.providerId))
```

shadowing `managedDataSets` here makes it confusing

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -708,72 +704,41 @@ export class StorageContext {
...
-        // Create a temporary PDPServer for this specific provider's endpoint
-        if (!provider.products.PDP?.data.serviceURL) {
-          // Skip providers without PDP products
-          continue
-        }
-        const providerPdpServer = new PDPServer(null, provider.products.PDP.data.serviceURL)
-        await providerPdpServer.ping()
-        return provider
-      } catch (error) {
+        return await Promise.race(pings)
```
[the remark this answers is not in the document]
wjmelements: Moving the logging into the reject block would actually be noisy if we abort though.

rvagg:
> The prior priority tiers, such as existing pieces and existing data sets, are preserved

But they're not quite preserved, currently you'll always be pulled back to data sets with the most pieces in it, so the choice between the one with 1 piece and 20 will always land you with the 20 data set as long as you can ping the provider, so the behaviour is now changed. Maybe this is OK, but it's a change we'd need to deal with and think through the implications of.

The other major change with this is that we now end up talking to the closest & fastest SPs and completely ignore others. TTFB is now our main selection metric. This may be an OK design decision, but it's going to have some implications for the network and what it means to be an SP and how to compete for business. The existing randomness was helping us distribute the network a bit more.

Also, for the multi-context case, aren't we going to be hitting this same code multiple times, so pinging the same providers multiple times (with exclusion)?

How about an alternative form of this: The main purpose of the ping was to weed out providers that we can't talk to, it's an easy and quick test. When we get our list of top-level providers inside createContexts, we could do a bulk parallel ping of all of them, with a short timeout, maybe 500ms max, then that trims our initial list, and our smart select no longer needs to perform the ping because we've done it at the top level and that trimmed our list. Perhaps "smartSelectWithPing" is now different to "smartSelect" and createContexts only uses the latter while createContext uses the former.

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -615,48 +615,38 @@ export class StorageContext {
     )
 
     if (managedDataSets.length > 0 && !forceCreateDataSet) {
-      // Prefer data sets with pieces, sort by ID (older first)
-      const sorted = managedDataSets.sort((a, b) => {
-        if (a.currentPieceCount > 0 && b.currentPieceCount === 0) return -1
-        if (b.currentPieceCount > 0 && a.currentPieceCount === 0) return 1
-        return a.pdpVerifierDataSetId - b.pdpVerifierDataSetId
```
wjmelements: > But they're not quite preserved, currently you'll always be pulled back to data sets with the most pieces in it

No, the current tie breaker is dataset ID

wjmelements:
> Also, for the multi-context case, aren't we going to be hitting this same code multiple times, so pinging the same providers multiple times (with exclusion)?

Yes. I have a local diff that changes these functions to take a count and return an array but that will be more work and I will be prioritizing upload first.

rvagg:
provider selection all replaced now

PR #402, opened 2025-11-04
fix: onPieceAdded arg type is Hex

wjmelements:
Reviewer @hugomrdias
This fixes a minor issue where the transaction hash wasn't logged, and some other inconsistencies in docs and examples.
It seems this used to pass a transaction object and now it just passes the hash.
#### Changes
* fix onPieceAdded arg: transaction -> transactionHash

### DIFF
--- a/docs/src/content/docs/intro/getting-started.mdx
+++ b/docs/src/content/docs/intro/getting-started.mdx
@@ -586,11 +586,11 @@ const result = await context.upload(data, {
   onUploadComplete: (pieceCid) => {
     console.log(`Upload complete! PieceCID: ${pieceCid}`)
   },
-  onPieceAdded: (transaction) => {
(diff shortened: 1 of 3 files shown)


wjmelements:
before the fix:
```
✓ [File 1/2] Piece addition transaction: undefined
✓ [File 2/2] Piece addition transaction: undefined
```

after:
```
✓ Piece addition transaction: 0xbb5d5e586075325caa750095488250640756b43c2267e04ac4230e4d1154a66f
```

rvagg:
I thought we changed this originally to be the transaction itself, so you could `.wait()` or do whatever else you wanted with it. It seems like an appropriate place to be passing a `Transaction` object so maybe this is an oversight?

hugomrdias (approved):

PR #405, opened 2025-11-05
feat: streaming upload support

rvagg:
Depends on https://github.com/filecoin-project/curio/pull/783 being deployed to the active servers.

* Lifts the upload limit to 1GiB minus fr32 expansion amount, matching Curio
* `upload()` now accepts an `AsyncIterable<Uint8Array>` which should be generic enough to support most types of byte streams
* synapse-core's `sp.ts` does all of the hard work, but it _only_ accepts an `AsyncIterable<Uint8Array>`, so we convert it in sdk if we receive a `Uint8Array`
* `onProgress` callback called whenever we send a chunk
* Streaming CommP used now, so we no longer have the two-step bottleneck of calc then upload
* I've removed `src/piece` from sdk and put all of that functionality into core and pulled it in from there, this is a breaking change but I assume almost nobody is using that or if they are they know how to switch to core - @hugomrdias you may have opinions on this, please take a look at that

### DIFF
--- a/packages/synapse-core/src/piece.ts
+++ b/packages/synapse-core/src/piece.ts
@@ -19,16 +19,16 @@ const SHA2_256_TRUNC254_PADDED = 0x1012
 /**
  * Maximum upload size currently supported by PDP servers.
  *
- * 1 GiB adjusted for fr32 expansion: 1 GiB * (126/127) = 1,065,353,216 bytes
+ * 1 GiB adjusted for fr32 expansion: 1 GiB * (127/128) = 1,065,353,216 bytes
  *
- * Fr32 encoding adds 1 bit of padding per 254 bits, resulting in 127 bytes
- * of padded data for every 126 bytes of raw data.
+ * Fr32 encoding adds 2 bits of padding per 254 bits of data, resulting in 128 bytes
+ * of padded data for every 127 bytes of raw data.
  *
  * Note: While it's technically possible to upload pieces this large as Uint8Array,
  * streaming via AsyncIterable is strongly recommended for non-trivial sizes.
  * See SIZE_CONSTANTS.MAX_UPLOAD_SIZE in synapse-sdk for detailed guidance.
  */
-export const MAX_UPLOAD_SIZE = 1065353216 // 1 GiB * 126/127
+export const MAX_UPLOAD_SIZE = 1_065_353_216 // 1 GiB * 127/128
 
 /**
  * PieceCID - A constrained CID type for Piece Commitments.
@@ -95,6 +95,15 @@ function parseLegacyPieceCID(pieceCidString: string): LegacyPieceCID | null {
 }
 
 /**
+ * Type guard to check if a value is a CID
+ * @param value - The value to check
+ * @returns True if it's a CID
+ */
+function isCID(value: unknown): value is CID {
+  return typeof value === 'object' && value !== null && CID.asCID(value as CID) !== null
+}
+
+/**
  * Check if a CID is a valid PieceCID
  * @param cid - The CID to check
  * @returns True if it's a valid PieceCID
@@ -127,14 +136,12 @@ export function asPieceCID(pieceCidInput: PieceCID | CID | string | null | undef
     return parsePieceCID(pieceCidInput)
   }
 
-  if (typeof pieceCidInput === 'object' && CID.asCID(pieceCidInput as CID) !== null) {
-    // It's already a CID, validate it
-    if (isValidPieceCID(pieceCidInput as CID)) {
-      return pieceCidInput as PieceCID
+  if (isCID(pieceCidInput)) {
+    if (isValidPieceCID(pieceCidInput)) {
+      return pieceCidInput
     }
   }
 
-  // Nope
   return null
 }
 
@@ -152,9 +159,10 @@ export function asLegacyPieceCID(
     return null
   }
 
-  const pieceCid = asPieceCID(pieceCidInput as CID | string)
+  // Try converting as PieceCID first (handles PieceCID and CID types)
+  const pieceCid = asPieceCID(pieceCidInput as PieceCID | CID | string | null | undefined)
   if (pieceCid != null) {
-    // downgrade to LegacyPieceCID
+    // Downgrade PieceCID to LegacyPieceCID
     const digest = Digest.create(SHA2_256_TRUNC254_PADDED, pieceCid.multihash.digest.subarray(-32))
     return Link.create(FIL_COMMITMENT_UNSEALED, digest) as LegacyPieceCID
   }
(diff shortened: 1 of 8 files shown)


packages/synapse-core/src/sp.ts
```diff
@@ -374,6 +414,154 @@ export async function uploadPiece(options: UploadPieceOptions) {
...
+
+  const uploadUuid = locationMatch[1]
+
+  // Stream data with CommP calculation
+
+  // Create CommP calculator stream
+  const { stream: pieceCidStream, getPieceCID } = Piece.createPieceCIDStream()
+
+  // Convert AsyncIterable to ReadableStream
+  const dataStream = asyncIterableToReadableStream(options.data)
```
SgtPooki: if we accepted a readableStream we wouldn't need to convert here

packages/synapse-core/src/sp.ts
```diff
@@ -374,6 +414,154 @@ export async function uploadPiece(options: UploadPieceOptions) {
...
+    throw finalizeResponse.error
+  }
+
+  if (finalizeResponse.result.status !== 200) {
+    throw new PostPieceError(`Expected 200 OK for finalization, got ${finalizeResponse.result.status}`)
+  }
+
+  return {
+    pieceCid,
+    size: bytesUploaded,
```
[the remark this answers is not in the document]
rvagg: Unfortunately they can't be parameters now, we only figure out both of these values _after_ we've uploaded since we can't see the entire stream - we don't know how many bytes until the end, and we're calculating CommP during streaming and can't know it until we're done with the upload

we could consider some refactoring here though, but the problem is that this is a 3 step process:

1. create an upload "session" with curio
2. stream upload data to to session endpoint
3. post our calculated commp to curio and have curio agree that it's correct

It's that final step which presents the biggest challenges for parallel uploads, we only want to calculate commp once per stream, but if we're forking the stream to send to multiple endpoints we need to be calculating it elsewhere and making it available for this third step. Possibly we extract this third step to a separate method that we can follow-up later on, or, as per my suggestion above, we provide a stream _and_ a "here's a function you can call to get commp that I calculated via other means".

packages/synapse-core/src/piece.ts
```diff
@@ -16,7 +41,138 @@ import * as Link from 'multiformats/link'
...
+export function asPieceCID(pieceCidInput: PieceCID | CID | string | null | undefined): PieceCID | null {
+  if (pieceCidInput === null || pieceCidInput === undefined) {
+    return null
+  }
+
+  if (typeof pieceCidInput === 'string') {
+    return parsePieceCID(pieceCidInput)
+  }
+
+  if (typeof pieceCidInput === 'object' && CID.asCID(pieceCidInput as CID) !== null) {
```
SgtPooki: if this were a typeguard, all the typecasting inside should go away

packages/synapse-core/src/piece.ts
```diff
@@ -89,3 +262,234 @@ export function calculate(data: Uint8Array) {
...
+ * @returns An object with the TransformStream and a getPieceCID function to retrieve the result
+ *
+ * @example
+ * const { stream, getPieceCID } = createPieceCIDStream()
+ * await fetch(url, {
+ *   method: 'PUT',
+ *   body: dataStream.pipeThrough(stream)
+ * })
+ * const pieceCid = getPieceCID() // Available after stream completes
+ */
```
SgtPooki: nice

packages/synapse-core/src/piece.ts
```diff
@@ -89,3 +262,234 @@ export function calculate(data: Uint8Array) {
...
+ * )
+ * ```
+ */
+export async function downloadAndValidateFromUrl(
+  url: string,
+  expectedPieceCid: string | PieceCID
+): Promise<Uint8Array> {
+  const response = await fetch(url)
+  return await downloadAndValidate(response, expectedPieceCid)
+}
```
SgtPooki: I feel like downloadAndValidate is nice enough to where we probably don't need this. If we do.. will we ever need to override fetch options? should we accept InitOptions as well as URL | string?

SgtPooki (commented):
made it through about ~600 lines so far..

packages/synapse-core/src/piece.ts
```diff
@@ -89,3 +262,234 @@ export function calculate(data: Uint8Array) {
...
+ */
+export async function calculateFromIterable(data: AsyncIterable<Uint8Array>): Promise<PieceCID> {
+  const hasher = Hasher.create()
+
+  for await (const chunk of data) {
+    hasher.write(chunk)
+  }
+
+  const digest = hasher.digest()
+  return Link.create(Raw.code, digest) as PieceCID
```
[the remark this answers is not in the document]
rvagg: going to leave this one, I don't mind explicitness in these kinds of situations

packages/synapse-sdk/src/utils/constants.ts
```diff
@@ -161,10 +162,26 @@ export const SIZE_CONSTANTS = {
...
+   * non-trivial sizes. In-memory operations with large byte arrays can:
+   * - Cause memory pressure and garbage collection issues
+   * - Block the JavaScript event loop during allocation/copying
+   * - Trigger browser tab slowdowns or "unresponsive script" warnings
+   *
+   * For optimal performance and resource utilization, prefer streaming for pieces
+   * larger than a few megabytes, especially in browser environments.
+   *
+   * Imported from @filoz/synapse-core/piece
    */
```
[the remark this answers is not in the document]
rvagg: might get to this when I enable streaming on the parent `upload()`, for now it'll just throw if you give it a stream

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -146,26 +159,55 @@ export class StorageManager {
...
+    // Multi-context upload handling
+    if (contexts.length > 1) {
+      // Multi-context uploads require Uint8Array to calculate pieceCid once
+      if (!(data instanceof Uint8Array)) {
+        throw createError(
+          'StorageManager',
+          'upload',
+          'Multi-context uploads currently only support Uint8Array. ' +
+            'For streaming uploads to multiple providers, convert your stream to Uint8Array first.'
+        )
```
SgtPooki: i'm assuming this is a temporary solution for the forking-stream + streaming-pieceCid-calculation problem?

The reason i'm asking is i'm curious what folks' solutions' are for getting around this once merged? Things i'm thinking:

1. convert to uint8array as error states..
2. perform multiple explicit upload calls with your stream (but they still need to calculate pieceCid in this case, right?)

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -146,26 +159,55 @@ export class StorageManager {
...
+
+      // Upload to single context
+      const result = await context.upload(data, {
+        ...options?.callbacks,
+        metadata: options?.metadata,
+        pieceCid,
+        signal: options?.signal,
+      })
+
+      return [{ status: 'fulfilled' as const, value: result }]
```
SgtPooki: not necessarily an issue for this PR, but:

I know I commented about us returning `promise.allSettled` instead of `promise.all` previously, but idk how I feel about an SDK returning `Promise<PromiseSettledResult<UploadResult>[]>`.

definitely something consumers should be able to handle, and keeps it flexible without maintainer burden on handling all the edge-cases, but that means consumers will need to, right?

I don't think it's a bad call, just something I think should have an explicit agreement from the core maintainers upon, and maybe some callouts for consumers.

packages/synapse-core/src/sp.ts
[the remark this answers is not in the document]
rvagg: so ... I'm just going to leave these return values here for now, they're not essential, but they are lossy if you use the streaming flow independently and allow it to take care of commp and size calculation for you since it'll do it streaming and we as yet don't have a way to pass in a deferred commp value -- I'll get to that in the next PR but I still think this is a reasonable return value for this streaming version

PR #410, opened 2025-11-06
feat!(StorageManager): upload for contexts

### LINKED ISSUE #312: Default Replication
We would like for storage contexts to upload to multiple storage providers by default. It should still be possible to select only one storage provider, so this will be configurable.

Without proof of replica, it is possible for replicated service providers to cheat proof of data possession. Therefore we plan to implement an endorsement system in the short term to certify availability. It should be possible to notice by latency if a service provider only has data possession by proxy.

My current plan to implement this is to have a replica count configuration parameter for createContext, and a wrapper replicated storage context that uses `Promise.all` on child context operations. This would need to handle individual SP failures gracefully. However because there is not a deadline on operations like AddPieces, it is still possible that a SP may succeed after we have determined that it failed, so simply replacing an apparently bad SP can result in paying for extra replicas.

I also want to make prior replica data sets somewhat discoverable in the same way that we discover your prior data sets. There are ways this can fail though. For example, a prior replicated data set may have been terminated or created with a different number of replicas. I don't have a good solution for this but I'll fallback to large existing relationships and use the new storage context preferences for new pieces.

Followups to this changeset:
* helper to identify user's piece CID duplicates and their statuses
* helper to boost (or repair) replication for previously uploaded pieces or data sets
* upload via URL for faster uploads (requires Curio change)
* prefer to use at least one endorsed SP by default

wjmelements:
Reviewer @rvagg @hugomrdias
Closes #312
Parallel uploading to different Storage Providers will use StorageManager.
I tried not to break StorageContext.upload.
As a follow up we would want to upload via URL in some cases, but that requires a curio change.
#### Changes
* change return type of StorageManager.upload to array
* calculate the UploadResponse fields pieceId and size pre-flight
* convert buffer to UInt8Array and calculate pieceId only once
* unit test StorageManager.upload

### DIFF
--- a/packages/synapse-sdk/src/test/storage.test.ts
+++ b/packages/synapse-sdk/src/test/storage.test.ts
@@ -14,7 +14,12 @@ import { SIZE_CONSTANTS } from '../utils/constants.ts'
 import { WarmStorageService } from '../warm-storage/index.ts'
 import { ADDRESSES, JSONRPC, PRIVATE_KEYS, PROVIDERS, presets } from './mocks/jsonrpc/index.ts'
 import { mockServiceProviderRegistry } from './mocks/jsonrpc/service-registry.ts'
-import { findPieceHandler, postPieceHandler, uploadPieceHandler } from './mocks/pdp/handlers.ts'
+import {
+  createAndAddPiecesHandler,
+  findPieceHandler,
+  postPieceHandler,
+  uploadPieceHandler,
+} from './mocks/pdp/handlers.ts'
 import { PING } from './mocks/ping.ts'
 
 // MSW server for JSONRPC mocking
@@ -1271,12 +1276,7 @@ describe('StorageService', () => {
         http.get('https://pdp.example.com/pdp/piece', async () => {
           return HttpResponse.json({ pieceCid: testPieceCID })
         }),
-        http.post('https://pdp.example.com/pdp/data-sets/create-and-add', () => {
-          return new HttpResponse(null, {
-            status: 201,
-            headers: { Location: `/pdp/data-sets/created/${mockTxHash}` },
-          })
-        }),
+        createAndAddPiecesHandler(mockTxHash, pdpOptions),
         http.get('https://pdp.example.com/pdp/data-sets/created/:tx', async () => {
           return HttpResponse.json(
             {
@@ -1344,12 +1344,7 @@ describe('StorageService', () => {
         http.get('https://pdp.example.com/pdp/piece', async () => {
           return HttpResponse.json({ pieceCid: testPieceCID })
         }),
-        http.post('https://pdp.example.com/pdp/data-sets/create-and-add', () => {
-          return new HttpResponse(null, {
-            status: 201,
-            headers: { Location: `/pdp/data-sets/created/${mockTxHash}` },
-          })
-        }),
+        createAndAddPiecesHandler(mockTxHash, pdpOptions),
         http.get('https://pdp.example.com/pdp/data-sets/created/:tx', async () => {
           return HttpResponse.json(
             {


packages/synapse-sdk/src/pdp/server.ts
```diff
@@ -494,13 +484,13 @@ export class PDPServer {
    * @param data - The data to upload
    * @returns Upload response with PieceCID and size
    */
-  async uploadPiece(data: Uint8Array | ArrayBuffer): Promise<UploadResponse> {
+  async uploadPiece(data: Uint8Array, pieceCid: PieceCID): Promise<void> {
     // Convert ArrayBuffer to Uint8Array if needed
-    const uint8Data = data instanceof ArrayBuffer ? new Uint8Array(data) : data
```
wjmelements: I removed the ability of PDPServer.uploadPiece to accept ArrayBuffer but if we actually need that I can put it back.

wjmelements:
Tomorrow I will write some tests using `src/test/mocks/pdp/handlers.ts`. I will need to add some missing upload handlers such as `PUT` `pdp/piece/upload/${uploadUuid}` and `POST` `pdp/piece`.

rvagg:
Yep, good by me, 🚢

I'm going to have to change a bit of this for #405 but I'll just work on top of this branch now assuming this is the structure we're going with. Pulling PieceCID calculation out is good, and helps #405.

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -133,14 +137,32 @@ export class StorageManager {
...
-    })
+    // Get the context to use
+    const contexts = options?.contexts ?? [options?.context ?? (await this.createContext(options))]
+    // Convert once
+    const dataBytes = data instanceof ArrayBuffer ? new Uint8Array(data) : data
+    // Calculate pieceCid once
+    const pieceCid = Piece.calculate(dataBytes)
+
+    // Upload using the contexts with piece metadata
+    return await Promise.all(
```
SgtPooki: If one of 5 contexts fails, this will reject all of them.. 

Will those background uploads be aborted? Currently doesn't look like it (no `signal` support)
Should we do `allSettled` and have a return that includes successes/failures?

wjmelements:
Happy case has test. Going to test some failure cases after dinner.

packages/synapse-core/src/sp.ts
```diff
@@ -313,14 +314,13 @@ export type UploadPieceResponse = {
...
-export async function uploadPiece(options: UploadPieceOptions) {
+export async function uploadPiece(options: UploadPieceOptions): Promise<void> {
   const size = options.data.length
   if (size < SIZE_CONSTANTS.MIN_UPLOAD_SIZE || size > SIZE_CONSTANTS.MAX_UPLOAD_SIZE) {
     throw new InvalidUploadSizeError(size)
   }
 
-  const pieceCid = Piece.calculate(options.data)
-
+  const pieceCid = options.pieceCid
```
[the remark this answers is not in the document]
rvagg: didn't you listen to Gozala @hugomrdias? https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/, hence the `asPieceCID` and no `isPieceCID` in the original implementaiton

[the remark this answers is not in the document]
hugomrdias: lol pls dont open old wounds, every time i see Link i still remember

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -835,26 +836,29 @@ export class StorageContext {
   /**
    * Upload data to the service provider
    */
-  async upload(data: Uint8Array | ArrayBuffer, options?: UploadOptions): Promise<UploadResult> {
+  async upload(data: Uint8Array | ArrayBuffer, options?: UploadOptions, pieceCid?: PieceCID): Promise<UploadResult> {
```
rvagg: I've just realised that I should have called out this breaking API, `pieceCid` should be in `UploadOptions`

In my streaming PR I'm putting it in there, so we'll have a break & unbreak

Issue #392, opened 2025-11-04
Synapse API to query the remaining egress allowances for a given data set


bajtos:
In order to allow FilBeam users to understand how much egress allowances they have remaining for CDN and cache-miss requests, Synapse SDK should provide an API to obtain the remaining allowances for a given data set.

We should also update Synapse documentation to explain developers how to:
- Check the remaining CDN & cache-miss balance
- Top up the CDN balance and the cache-miss balance

Task list:

- [x] Add Synapse API for checking the remaining balances → https://github.com/FilOzone/synapse-sdk/pull/430
- [ ] Docs: how to check the remaining balances
- [ ] Docs: how to top up the balances

pyropy:
Reopening as we didn't yet document how to check and top-up egress quotas on docs.filbeam.com

PR #430, opened 2025-11-12
feat: Add API for querying remaining egress

pyropy:
Extends existing Synapse SDK API with FilBeam service allowing for querying the remaining data set egress. 

Closes https://github.com/FilOzone/synapse-sdk/issues/392

### DIFF
--- a/.github/release-please-manifest.json
+++ b/.github/release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-  "packages/synapse-sdk": "0.36.0",
-  "packages/synapse-core": "0.1.3",
-  "packages/synapse-react": "0.1.3"
+  "packages/synapse-sdk": "0.36.1",
+  "packages/synapse-core": "0.1.4",
+  "packages/synapse-react": "0.1.4"
 }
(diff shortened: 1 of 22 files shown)


packages/synapse-sdk/src/synapse.ts
```diff
@@ -166,6 +168,9 @@ export class Synapse {
       pieceRetriever = new FilBeamRetriever(baseRetriever, network)
     }
 
+    // Create FilBeamService
+    const filbeamService = await FilBeamService.create(provider)
```
rvagg: You should change this to just take `network`, not `provider` because that's all you're using it for. Then it doesn't need to be async, you can just use a constructor instead of the async factory, and you don't have yet another `getNetworkType` call in the Synapse initialisation path (it should be cached, but we don't have a guarantee of that). Nice and fast to get this set up.

packages/synapse-sdk/src/filbeam/index.ts
```diff
@@ -0,0 +1,94 @@
...
+ *     json: async () => ({
+ *       cdnEgressQuota: '1099511627776', // 1 TiB in bytes
+ *       cacheMissEgressQuota: '549755813888' // 0.5 TiB in bytes
+ *     })
+ *   } as Response
+ * }
+ *
+ * const service = new FilBeamService('mainnet', mockFetch)
+ * const stats = await service.getDataSetStats('test')
+ * ```
```
rvagg:
```suggestion
```

this is confusing for user-facing docs, let's not

PR #439, opened 2025-11-14
feat(storage): add dataSetId to piece confirmation callbacks

### LINKED ISSUE #409: emit dataSetId when it's confirmed created
https://github.com/FilOzone/synapse-sdk/blob/833e14bbedb27ffe22b4fa18f9f1ddee94d8e8a6/packages/synapse-sdk/src/storage/context.ts#L1026

currently, there is no exposure to consumers that a dataSetId exists once it's created with the new createAndAdd functionality.. 

we should add another callback, or provide that dataSetId in `onPieceConfirmed`

virajbhartiya:
- Add `onPiecesConfirmed(dataSetId, pieces)` callback exposing dataset ID when pieces are confirmed
- Add `onPiecesAdded(transaction, pieces)` callback with piece CIDs
- Deprecate `onPieceAdded` and `onPieceConfirmed` 
- Update docs and examples to demonstrate new callbacks

Closes #409


virajbhartiya:
@rvagg @SgtPooki can you please review

virajbhartiya:
@SgtPooki rebased

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -1045,7 +1058,16 @@ export class StorageContext {
 
         confirmedPieceIds.push(...(confirmedPieces.confirmedPieceIds ?? []))
 
+        const dataSetId = this.dataSetId
+        if (dataSetId == null) {
+          throw createError('StorageContext', 'addPieces', 'Data set ID missing during confirmation callbacks')
+        }
```
hugomrdias: why this check ? this should never be null

hugomrdias (changes requested):

packages/synapse-sdk/src/types.ts
```diff
@@ -404,14 +404,23 @@ export interface PreflightInfo {
  * 2. Piece addition (transaction submitted to chain)
  * 3. Confirmation (transaction confirmed on-chain)
  */
+export interface PieceIdentifiers {
```
SgtPooki: should this be PiecePair removed entirely? `{ pieceCid: PieceCID; pieceId: number }` is used in quite a few places, a good name for this representation of a Piece could be useful..

if the core maintainers disagree we could just remove this.

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -1045,7 +1058,16 @@ export class StorageContext {
...
+        const piecesForConfirmed: PieceIdentifiers[] = confirmedPieceIds.map((pieceId, index) => ({
```
SgtPooki:
```suggestion
        const confirmedPiecePairs: PieceIdentifiers[] = confirmedPieceIds.map((pieceId, index) => ({
```

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -997,14 +999,24 @@ export class StorageContext {
...
         const addPiecesResponse = await SP.pollForAddPiecesStatus(addPiecesResult)
 
         // Handle transaction tracking if available
         confirmedPieceIds.push(...(addPiecesResponse.confirmedPieceIds ?? []))
 
+        const dataSetId = this.dataSetId
+        if (dataSetId == null) {
+          throw createError('StorageContext', 'addPieces', 'Data set ID missing during confirmation callbacks')
+        }
+        const piecesForConfirmed: PieceIdentifiers[] = confirmedPieceIds.map((pieceId, index) => ({
```
SgtPooki:
```suggestion
        const confirmedPiecePairs: PieceIdentifiers[] = confirmedPieceIds.map((pieceId, index) => ({
```

rvagg:
Looks good! Just that one doc fix and I think this is OK to go.

PR #445, opened 2025-11-16
chore: replace filfox with blockscout

### LINKED ISSUE #428: Update contract links to Blockscout
Filfox's contract verifier/tool is unreliable and often fails to verify contracts, even when verification scripts run successfully. Blockscout has been more reliable with verification scripts and contains more valuable information. We should update references and links across the stack to use Blockscout instead of Filfox as the default block explorer to explore contract verification code.

akronim26:
# Fixes #428 
- This updates the default block explorer links from Filfox to Blockscout. In a few places where both Filfox and Blockscout links were already present, I left those unchanged, since the issue seemed to target only the default explorer.

### DIFF
--- a/.github/release-please-manifest.json
+++ b/.github/release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-  "packages/synapse-sdk": "0.35.3",
-  "packages/synapse-core": "0.1.2",
-  "packages/synapse-react": "0.1.2"
+  "packages/synapse-sdk": "0.36.0",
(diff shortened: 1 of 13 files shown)


packages/synapse-core/src/abis/generated.ts
hugomrdias: please remove all changes to this file, this is a auto generated files and should be change manually

hugomrdias (changes requested):

Issue #437, opened 2025-11-13
Add intelligent synapse.payments.fund(amount) (improve golden path)


hugomrdias:
```js
import { Synapse, RPC_URLS, TIME_CONSTANTS } from "@filoz/synapse-sdk";
import { ethers } from "ethers";

const synapse = await Synapse.create({
  privateKey: "YOUR_PRIVATE_KEY",
  rpcURL: RPC_URLS.calibration.http,
});

const tx = await synapse.payments.depositWithPermitAndApproveOperator(
  ethers.parseUnits("10", 18),
  synapse.getWarmStorageAddress(),
  ethers.MaxUint256,
  ethers.MaxUint256,
  30n * TIME_CONSTANTS.EPOCHS_PER_DAY,
);
await tx.wait();

const dataBytes = new TextEncoder().encode(
  ":rocket: Welcome to decentralized storage on Filecoin Onchain Cloud!"
);
const result = await synapse.storage.upload(dataBytes);

await synapse.storage.download(result[0].value.pieceCid);
```

- add defaults to `depositWithPermitAndApproveOperator` should only require ammount
- `depositWithPermitAndApproveOperator` can be just `depositAndApprove`
- the new upload return type is awkward `result[0].value.pieceCid`

rvagg:
Discussion from in person today: we should make a new method on `synapse.payments` called `fund()` (name suggestions welcome) that's distinct from `deposit` and `depositWithPermitAndApproveOperator` but has the intelligence to figure out which one to use depending on the options provided. It should have the defaults presented above, i.e. be WarmStorage specific. This leaves the existing methods be pass-throughs to the contract but provides a simple method that can do what's needed.

rvagg:
Let's scope this down strictly to just adding a new `fund(amount)` method that does `depositWithPermitAndApproveOperator` or `deposit()` depending on what it finds about your status. Checks are similar to what Filecoin Pin already does. Defaults for `depositWithPermitAndApproveOperator` are as above in the issue (same as Filecoin Pin).

PR #446, opened 2025-11-17
fix!: revert golden path upload() API + minor fixes

rvagg:
This is a temporary revert of the top-level synapse.storage.upload() API to default to a single provider and return the original Promise<UploadResult> type, restoring the original functionality as presented in our docs. Multi-provider uploads both makes this API too complex to use (mainly thanks to the return type of allSettled) yet also doesn't present enough information to make it useful for an API (e.g. callbacks are not individuated).

We'll continue to work on this and consume it in our demo apps to get the API ergonomics right, but for now we'll restore the original DX of this top-level API to match our existing docs and avoid the confusion and complexity for developers.

Ref: https://github.com/FilOzone/synapse-sdk/issues/437

Also:

* Only Uint8Array and ReadableStream on top-level APIs for now
* Fix some pieceCid calculation issues w/streaming (wasn't stream calculating in most cases where it could), so we shouldn't have to pause-calculate it for single uploads
* Lots of fixes in example script
*

### DIFF
--- a/packages/synapse-sdk/src/storage/manager.ts
+++ b/packages/synapse-sdk/src/storage/manager.ts
@@ -124,17 +124,17 @@ export class StorageManager {
    * Uses the storage contexts or context provided in the options
    * Otherwise creates/reuses default context
    *
-   * Accepts Uint8Array, AsyncIterable<Uint8Array>, or ReadableStream<Uint8Array>.
-   * For large files, prefer streaming types to minimize memory usage.
+   * Accepts Uint8Array or ReadableStream<Uint8Array>.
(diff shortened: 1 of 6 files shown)


packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -168,6 +168,7 @@ export class StorageManager {
         : await this.createContexts({
             withCDN: options?.withCDN,
             withIpni: options?.withIpni,
+            count: 1, // single context by default for now - this will be changed in a future version
```
wjmelements: why not this version?

Issue #453, opened 2025-11-19
Dutch Auction support


wjmelements:
We would like to add support for the FilecoinPay Dutch Auction to this SDK.

This would help with an auction frontend.

PR #454, opened 2025-11-20
feat: auctionPriceAt

wjmelements:
Reviewer @rvagg @hugomrdias
Closes https://github.com/FilOzone/synapse-sdk/issues/453
#### References
* https://github.com/PaulRBerg/prb-math/blob/main/src/ud60x18/Constants.sol
* https://github.com/PaulRBerg/prb-math/blob/main/src/ud60x18/Math.sol 
* https://github.com/PaulRBerg/prb-math/blob/main/src/Common.sol
* https://github.com/FilOzone/filecoin-pay/blob/main/src/Dutch.sol
* https://github.com/FilOzone/filecoin-pay/blob/main/test/Dutch.t.sol
#### Changes
* helper functions for getting auction info and accumulated fees
* auctionPriceAt matching Dutch.sol
* testing to match Dutch.t.sol

### DIFF
--- a/packages/synapse-core/src/auction/auction.ts
+++ b/packages/synapse-core/src/auction/auction.ts
@@ -14,6 +14,15 @@ export type AuctionInfo = {
   startTime: bigint
 }
 
+/**
+ * Get the initial state of the current FilecoinPay auction
+ * This auction information can be passed to auctionPriceAt to calculate the current price
+ * Auctions are dutch, so the acceptance price decreases over time
+ * Each token has a separate auction
+ * @param client used to read the FilecoinPay contract
+ * @param token specifies which token's auction
+ * @returns the auction startPrice and startTime
+ */
 export async function auctionInfo(client: Client<Transport, Chain>, token: Address): Promise<AuctionInfo> {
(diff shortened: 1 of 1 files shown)


wjmelements:
@jennijuju suggested the auction logic could live in another package or in another repo. It's mainly making use of the wagmi ABI. Since this is a monorepo I don't think it needs to be in a different repo. @hugomrdias might have a stronger preference than me.

packages/synapse-core/src/auction/auction.ts
```diff
@@ -0,0 +1,61 @@
...
+export async function auctionFunds(client: Client<Transport, Chain>, token: Address): Promise<bigint> {
+  const chain = getChain(client.chain.id)
+  const [funds] = await readContract(client, {
+    address: chain.contracts.payments.address,
+    abi: chain.contracts.payments.abi,
+    functionName: 'accounts',
+    args: [token, chain.contracts.payments.address],
+  })
+  return funds
+}
```
hugomrdias: pls add some comments explaining the method

PR #462, opened 2025-11-24
chore: export StorageManagerUploadOptions

### LINKED ISSUE #461: `StorageManagerUploadOptions` type is not exported
https://github.com/FilOzone/synapse-sdk/blob/642b1fd947a24c423d0e0b9645c73e4ed29698c9/packages/synapse-sdk/src/storage/manager.ts#L74C11-L74C38

This makes setting up an options object verbose for consumers, see https://github.com/filecoin-project/filecoin-pin/pull/262/commits/6861385ab189b481f8ef719b75d235e6d90f7317

kaf-lamed-beyt:
@SgtPooki, i also updated the JSDoc comment, let me know what you think.

closes #461

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -68,6 +68,16 @@ If using an AI tool, you are welcome to load AGENTS.md into your context to teac
 pnpm run generate-abi # in the synapse-core package
 ```
 
+### Wireit
+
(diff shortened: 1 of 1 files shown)


SgtPooki (commented):
code change lgtm, but this was marked as `@internal` explicitly. I wonder if core maintainers want to export a wrapped type or change something before we expose this

PR #463, opened 2025-11-24
chore: set minimum acceptable release age to 7 days

galargh:
Context: https://ipdx-workspace.slack.com/archives/C06CY12V83S/p1764009233782709


rvagg (approved):
this is fine, devs really should be doing this themselves in their own global config file (and probably also using the socket.dev pnpm and npm wrappers)

PR #465, opened 2025-11-25
chore: update msw

hugomrdias:


rvagg:
https://github.com/mswjs/msw/issues/2624

someone just needs to fix it upstream

PR #471, opened 2025-11-27
fix: enable ledger signing to `endorse-sp.js`

rjan90:
Add ledger signing ability to `endorse-sp.js`

### DIFF
--- a/packages/synapse-core/tools/endorse-sp.js
+++ b/packages/synapse-core/tools/endorse-sp.js
@@ -101,7 +101,7 @@ async function main() {
       chain: getChain(Number(CHAIN_ID)),
     })
 
-    console.log('Expiry:', new Date(Number(EXPIRY) * 1000).toDateString())
+    console.log('Expiry:', new Date(Number(EXPIRY) * 1000).toLocaleString())
 
     for (const providerId of providerIds) {
       if (USE_LEDGER) console.log('\n⏳ Confirm on Ledger for provider:', providerId)


packages/synapse-core/tools/endorse-sp.js
```diff
@@ -1,19 +1,28 @@
+import EthModule from '@ledgerhq/hw-app-eth'
+import TransportNodeHidModule from '@ledgerhq/hw-transport-node-hid'
 import { createWalletClient, http } from 'viem'
-import { privateKeyToAccount } from 'viem/accounts'
+import { privateKeyToAccount, toAccount } from 'viem/accounts'
+
+const TransportNodeHid = TransportNodeHidModule.default || TransportNodeHidModule
```
rvagg: I don't think this is needed, this looks like it's either making allowance for dodgy bundlers (not a problem here) or dealing with an older version of the module. According to https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/hw-transport-node-hid, we should just be able to use the main export `import TransportNodeHid from "@ledgerhq/hw-transport-node-hid";` - i.e. get rid of `Module` on the end of the import above and remove this line.

packages/synapse-core/tools/endorse-sp.js
```diff
@@ -1,19 +1,28 @@
...
 function printUsageAndExit() {
-  console.error('Usage: PRIVATE_KEY=0x... node utils/endorse-sp.js providerId...')
+  console.error('Usage: PRIVATE_KEY=0x... node tools/endorse-sp.js providerId...')
+  console.error('   or: USE_LEDGER=true node tools/endorse-sp.js providerId...')
   process.exit(1)
 }
 
 const PRIVATE_KEY = process.env.PRIVATE_KEY
+const USE_LEDGER = process.env.USE_LEDGER === 'true'
+const LEDGER_PATH = process.env.LEDGER_PATH || "m/44'/60'/0'/0/0"
```
rvagg:
```suggestion
const LEDGER_PATH = process.env.LEDGER_PATH ?? "m/44'/60'/0'/0/0"
```

see https://github.com/FilOzone/synapse-sdk/pull/443#discussion_r2561135276 - `??` is a newer operator that avoids the vagaries of JS's "falsy" values, where various strings can evaluate to `false` which makes it easier to fall through to the default value here. But `??` gives us strictness where if someone provides a `LEDGER_PATH` that happens to be falsy too, like `"0"`, or even `""`, it should end up failing rather than silently passing and using the default path which may be a surprise.

packages/synapse-core/tools/endorse-sp.js
```diff
@@ -26,6 +35,31 @@ if (providerIds.length === 0) {
   printUsageAndExit()
 }
 
+async function createLedgerAccount() {
+  const transport = await TransportNodeHid.open('')
+  const eth = new Eth(transport)
+
+  const { address } = await eth.getAddress(LEDGER_PATH)
+
+  const account = toAccount({
```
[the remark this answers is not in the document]
hugomrdias: yeah, transport needs to be handle outside. Created a PR to support ledger in wagmi and viem https://github.com/FilOzone/synapse-sdk/pull/478

BigLep:
2025-12-17 conversation: 
- Needs to be updated to use new dependency that @hugomrdias added for getting Ledger
- If this gets merged, there will be a separate PR/issue for having an example rather than another script.

PR #478, opened 2025-12-03
feat: add support for ledger

hugomrdias:
- add ledger connector to synapse-react
- add ledgerToAccount to synpase-core
- add support for ledger connect in the demo app plus changing account
- upgrade to wagmi 3
- sync viem (we need to moved these dependencies to [catalog: protocol](https://pnpm.io/catalogs) so we dont touch all the packages just to sync them across)

**Rebase Merge** please


packages/synapse-react/src/ledger.ts
```diff
@@ -0,0 +1,280 @@
...
+          name: 'Ledger Wallet Client',
+          chain,
+          account: await ledgerToAccount({
+            transport: eth.transport,
+            accountIndex,
+            addressIndex,
+            changeIndex,
+          }),
+        })
+      },
```
rjan90: If changeAccount() is called, would not this return a wallet client for the original address, rather then the updated `path` ?

apps/synapse-playground/src/components/connect-wallet.tsx
```diff
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react'
-import { useAccount } from 'wagmi'
```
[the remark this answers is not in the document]
hugomrdias: no

rjan90 (approved):
Changes looks okay to me. One outstanding comment from Rod that would be worth aligning on before merging

docs/src/content/docs/core-concepts/fwss-overview.mdx
```diff
@@ -156,7 +156,7 @@ If your balance drops below this threshold, you risk default—at which point th
 FWSS enables the Filecoin Onchain Cloud to behave not just as a passive archive, but as a live data service:
 
 - Storage providers are continuously verified via PDP
-- Clients gain access to retrievable, service‐ready data
+- Clients gain access to retrievable, service-ready data
 - Payments are automated via onchain rails linked to proof events
 
-By linking upload, proof, retrieval, and payment, FWSS delivers cloud-style storage with blockchain‐grade guarantees.
+By linking upload, proof, retrieval, and payment, FWSS delivers cloud-style storage with blockchaingrade guarantees.
```
rvagg:
```suggestion
By linking upload, proof, retrieval, and payment, FWSS delivers cloud-style storage with blockchain-grade guarantees.
```

Issue #485, opened 2025-12-05
`resolveByDataSetId` doesn't check metadata consistency


rvagg:
Thanks to https://github.com/FilOzone/synapse-sdk/pull/438 I'm noticing that when you select a specific data set it doesn't check your metadata matches. It only checks that `withCDN` matches. We need to figure out the strategy here and it's a little bit tricky.

In `resolveByProviderId` we do a `metadataMatches` using `requestedMetadata`. But in that case you're fishing for a matching data set, in the `resolveByDataSetId` case you've already explicitly selected one, so is it the SDK's business to be checking that it matches the metadata you requested. Currently we do some checking, `withCDN` must match the data set. So we're doing a half-way job, which isn't good. I think we should either go further, or do nothing:

Options:

1. Strictly require that `requestedMetadata` matches the data set and error if not (this value is made in `resolveProviderAndDataSet` and could just be moved up to before the call to `resolveByDataSetId`).
2. Only require that meta data matches _if_ you provide some in your options - including `withCDN` being set. If `withCDN` && `metadata` are both `undefined` then perform no checks.
3. Perform no checks at all, including `withCDN`.

rvagg:
@SgtPooki: I would appreciate your thoughts in here from an API consumer perspective - should you always have to show up with matching metadata and have Synapse validate? Or should you just show up with a data set ID and that's enough. But if you do add metadata to your `options` along with a data set ID, what are we to make of that? Ignore it, error because that's not allowed, or match it?

rvagg:
In #593 I'm deciding to _not_ validate data set metadata - if you provide explicit data sets, then we'll do what we're told. I'm not even checking if/what metadata is provided because we can't distinguish between "no metadata" and "I didn't bother telling you".

```js
  /**
   * Resolve using a specific data set ID
   *
   * Note that unlike resolveByProviderId, we don't check for matching metadata,
   * we assume that if the client asked for this data set, then they know what
   * they are doing and are willing to deal with any conflicts.
   */
  private static async resolveByDataSetId(
...
```

Issue #435, opened 2025-11-13
Fix excessive call use in `resolveByDataSetId`


rvagg:
At the top of `resolveByDataSetId` in context.ts we do this:

```typescript
    // Fetch data sets to find the specific one
    const dataSets = await warmStorageService.getClientDataSetsWithDetails(signerAddress)
    const dataSet = dataSets.find((ds) => ds.pdpVerifierDataSetId === dataSetId)
```

This is not great if you have a lot of data sets, like Filecoin Pin Demo does (currently nearly 500). `getClientDataSetsWithDetails` does a whole lot of work. Per data set it has 3 call phases in series: 1 call, 3 calls in parallel, 1 call. We do that all in parallel across all data sets, but for 500 data sets that'd be 2,500 eth_calls!

So let's not do this where we don't need to.

We first need to figure out what information we _need_ for the one data set we care about in `resolveByDataSetId`, then see if we can get that information with the functions we currently have, and if not, we should make one. Just having a `getDataSetWithDetails(id)` that does the per-dataset work would be enough if we need all that information.

rvagg:
Will be closed by https://github.com/FilOzone/synapse-sdk/pull/438

PR #487, opened 2025-12-05
fix: resolveByProviderId doesnt use getClientDataSetsWithDetails

SgtPooki:
## Summary

Fix `resolveByProviderId` performance and selection logic to match existing behavior while avoiding expensive `getClientDataSetsWithDetails` calls.

## Problem

Issue #435: getClientDataSetsWithDetails makes ~5 calls per dataset (3 phases). With 639 datasets that’s ~3,195+ calls just to select one.

## Changes

- Use lightweight `getClientDataSets` instead of `getClientDataSetsWithDetails`
- Fixed sort: ascending by dataSetId (oldest first) to match existing behavior
- Sequential batched search (max(batchSize = max(50, ceil(n/3))) with early exit
- Metadata check first, bail early if no match (avoids unnecessary pieceCount/validate calls)
- Parallel pieceCount + validateDataSet only for metadata matches

## Performance

Benchmarks on a 639-dataset fixture; both paths pick the same dataSetId (84):

**Common case** (metadata matches at dataset 84):
- Old: 6,431 calls, ~13.0s
- New: 365 calls, ~9.6s
- Result: 1.36x faster, 94.3% fewer calls ✅

**Worst case** (metadata never matches, dataSetId=-1):
- Old: 6,429 calls, 27.6s
- New: 335 calls, 18s
- Result: 1.5x faster, 94.8% fewer calls ✅

## Testing

See testing results and code used at https://gist.github.com/SgtPooki/987fe31d3a8991ec73490d5df385b381

## Related

https://github.com/FilOzone/dealbot/issues/66
https://github.com/FilOzone/synapse-sdk/pull/438
https://github.com/FilOzone/synapse-sdk/issues/485 (for metadata match handling)


SgtPooki:
Ran some benchmarks with hyperfine. When there are only two datasets, the old method is usually faster for the common usecase when an account has few (2) data sets. That makes sense, because i'm trying to balance rpc call reduction and speed by not just spamming for details on everything. however, for worst-case scenarios (non-matching dataset metadata, so it checks all datasets) its roughly the same speed. 

for an account with a large number of datasets (dealbot, filecoin-pin-website, and potentially future users who don't follow our expectations), the changes in this PR are significantly faster with a serious reduction in RPC calls.

Key | Scenario | Old mean ± σ (s) | New mean ± σ (s) | Speedup (new vs old)
-- | -- | -- | -- | --
SUB | common | 2.590 ± 0.107 | 2.917 ± 0.349 | 0.89x (old ~13% faster)
SUB | worst | 2.497 ± 0.227 | 2.428 ± 0.286 | 1.03x (within noise)
MAIN | common | 15.583 ± 0.398 | 5.371 ± 1.233 | 2.90x faster
MAIN | worst | n/a (old not shown) | 6.211 ± 1.554 | 2.78x faster vs old (per hyperfine summary)

Question for @rvagg and @hugomrdias: Is rpc call reduction or latency the priority? Do we need to try to balance them? I can just `promise.all` the `getDataSetMetadata`, `getNextPieceId`, and `validateDataSet` calls to speed things up while still reducing RPC calls, but not by as much.

I think the approach I took here is worth considering, but open to feedback.

SgtPooki:
I quickly changed the code to run all dataset calls at once, and I'm thinking the RPC calls are still reduced enough to where we should probably just call them all at once.

### 2 dataSet account

#### common case
Speed: 1.59x faster
RPC calls: 48.3% reduction

#### worst case
Speed: 1.11x faster
RPC calls: 44.4% reduction

#### Avg results
Average speedup: 1.35x
Average RPC call reduction: 46.4%


and then hyperfine show the speedup is consistent (as much as 10 runs and networking can show):

```
Benchmark 1: PRIVATE_KEY=$PRIVATE_KEY_SUB RUN_OLD=true RUN_NEW=false SCENARIO=common RPC_TYPE=wss PROVIDER_ID=2 node scripts/benchmark-provider-resolve.js
  Time (mean ± σ):      2.556 s ±  0.043 s    [User: 0.747 s, System: 0.152 s]
  Range (min … max):    2.463 s …  2.612 s    10 runs

Benchmark 2: PRIVATE_KEY=$PRIVATE_KEY_SUB RUN_OLD=false RUN_NEW=true SCENARIO=common RPC_TYPE=wss PROVIDER_ID=2 node scripts/benchmark-provider-resolve.js
  Time (mean ± σ):      2.315 s ±  0.313 s    [User: 0.730 s, System: 0.151 s]
  Range (min … max):    2.082 s …  2.797 s    10 runs

Summary
  PRIVATE_KEY=$PRIVATE_KEY_SUB RUN_OLD=false RUN_NEW=true SCENARIO=common RPC_TYPE=wss PROVIDER_ID=2 node scripts/benchmark-provider-resolve.js ran
    1.10 ± 0.15 times faster than PRIVATE_KEY=$PRIVATE_KEY_SUB RUN_OLD=true RUN_NEW=false SCENARIO=common RPC_TYPE=wss PROVIDER_ID=2 node scripts/benchmark-provider-resolve.js
```

so.. it still reduces RPC calls by almost 50%, and is also faster 

### 639 dataSet account:

#### common case
Speed: 3.51x faster
RPC calls: 93.0% reduction

#### worst case
Speed: 2.39x faster
RPC calls: 79.5% reduction

#### Avg results
Average speedup: **2.95x**
Average RPC call reduction: **86.3%**

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -429,6 +430,17 @@ export class WarmStorageService {
...
 
+  /**
+   * Get the next piece ID for a dataset (effectively the current piece count)
+   * @param dataSetId - The PDPVerifier data set ID
+   * @returns The next piece ID as a number
+   */
+  async getNextPieceId(dataSetId: number): Promise<number> {
+    const pdpVerifier = this._getPDPVerifier()
+    const nextPieceId = await pdpVerifier.getNextPieceId(Number(dataSetId))
+    return Number(nextPieceId)
```
SgtPooki:
```suggestion
    return nextPieceId
```

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -325,6 +325,7 @@ export class WarmStorageService {
         clientDataSetId: ds.clientDataSetId,
         pdpEndEpoch: Number(ds.pdpEndEpoch),
         providerId: Number(ds.providerId),
+        dataSetId: Number(ds.dataSetId),
```
SgtPooki: what guarantees do we have that these are lower than MAX_SAFE_INTEGER?

SgtPooki:
looks like i need to merge master again.. on it

PR #518, opened 2025-12-18
fix: update example to deal with export changes

rvagg:
Ref: https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1765985016395079


rvagg:
* sessions: are intended to be more of an advanced feature for dapps to use to help minimise wallet interactions for users, not so much something for end-users to mess with
* typechecking - yes, although this isn't typescript code so there are limitations here, but also this directory needs a big-ol clean-up and what we want to keep probably needs to be moved into the new examples/ directory somehow

PR #527, opened 2026-01-02
feat: add devnet support

redpanda-f:
This PR adds comprehensive support for local Filecoin networks to enable easier development and testing.

### Changes
- **Add localnet support** to Synapse SDK for local Filecoin network development
- **Introduce localnet configuration utilities** with environment variable support for chain ID, RPC URLs, and contract addresses
- **Update FilecoinNetworkType** to include 'localnet' alongside mainnet and calibration
- **Enhance network detection** to automatically recognize localnet by chain ID
- **Add localnet contract address resolution** with required/optional address validation
- **Update example scripts** (storage-e2e, post-deploy-setup) to support `--network localnet` option
- **Export localnet configuration** from synapse-core and synapse-sdk packages
- **Maintain backward compatibility** with existing mainnet/calibration functionality

### Environment Variables
The following environment variables can be used to configure localnet:
- `LOCALNET_CHAIN_ID` (default: 1414)
- `LOCALNET_RPC_URL` (default: http://127.0.0.1:5700/rpc/v1)
- `LOCALNET_RPC_WS_URL` (default: ws://127.0.0.1:5700/rpc/v1)
- `LOCALNET_BLOCK_EXPLORER_URL` (default: http://localhost:8080)
- `LOCALNET_MULTICALL3_ADDRESS`, 
- `LOCALNET_USDFC_ADDRESS`, etc. for contract addresses

### DIFF
--- a/packages/synapse-sdk/README.md
+++ b/packages/synapse-sdk/README.md
@@ -38,33 +38,6 @@ pnpm install @filoz/synapse-sdk ethers
 
 Note: `ethers` v6 is a peer dependency and must be installed separately.
 
-## Quickstart (mainnet, calibration, devnet)
-
-```ts
-import { Synapse } from '@filoz/synapse-sdk'
-
-const synapse = await Synapse.create({
-	// One of privateKey, provider, or signer (privateKey shown here)
-	privateKey: process.env.PRIVATE_KEY ?? '',
-	// Defaults: mainnet and calibration auto-discover addresses; devnet requires explicit values
-	rpcURL: process.env.RPC_URL ?? 'https://api.calibration.node.glif.io/rpc/v1',
-	// Devnet-only (required): provide your deployment addresses
-	warmStorageAddress: process.env.WARM_STORAGE_ADDRESS,
-	multicall3Address: process.env.MULTICALL3_ADDRESS,
-	// Optional: override USDFC or view contract addresses when testing
-	usdfcAddress: process.env.USDFC_ADDRESS,
-	warmStorageViewAddress: process.env.WARM_STORAGE_VIEW_ADDRESS,
-})
-
-const file = new TextEncoder().encode('hello from synapse')
-const result = await synapse.storage.upload(file)
-console.log('pieceCid:', result.pieceCid)
-```
-
-- Networks supported: `mainnet`, `calibration`, `devnet`
-- Devnet requires `rpcURL`, `warmStorageAddress`, and `multicall3Address` to be provided
-- For session keys, payments, and registry access, services auto-discover addresses on mainnet/calibration
-
 ## Docs
 
 Check the documentation [website](https://synapse.filecoin.services/)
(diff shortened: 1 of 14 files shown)


packages/synapse-sdk/src/index.ts
```diff
@@ -9,6 +9,15 @@
  * ```
  */
 
+export {
```
rvagg: I think this was a mistake? we recently removed a bunch of stuff from here, are they needed for this PR?

packages/synapse-core/src/localnet-config.ts
rvagg: I don't think this is the right approach to take here. Env var reading doesn't really belong in synapse-sdk or synapse-core, that's an external concern, I'd very much prefer the approach of _injecting_ the things that need to be variable rather than implicitly discovering them by magic. We already have that with things like the ability to override `warmStorageAddress`. So a lot of the heavy-lifting of figuring out what to inject can come from outside, specifically in example-storage-e2e.js which you're using primarily to test and can demonstrate how to do it for other apps. I'll comment more inline with my thoughts on how we could do it alternatively.

packages/synapse-sdk/src/utils/constants.ts
```diff
@@ -327,6 +344,9 @@ export const CONTRACT_ADDRESSES = {
   WARM_STORAGE: {
     mainnet: '0x8408502033C418E1bbC97cE9ac48E5528F371A9f',
     calibration: '0x02925630df557F957f70E112bA06e50965417CA0',
+    get localnet() {
```
rvagg: this is not needed because the only place it matters is here where it can be overridden: https://github.com/FilOzone/synapse-sdk/blob/a1ec6dbe7ac9585b55454f4ffe84cc11ade31b16/packages/synapse-sdk/src/synapse.ts#L127

What we can do is throw if we get to that point and find we don't have an override address - the workflow I'd prefer is to force users to provide an address where we're not on calib or mainnet.

rvagg: in fact the current code should already cover this because it throws on the next line.

packages/synapse-sdk/README.md
```diff
@@ -38,6 +38,33 @@ pnpm install @filoz/synapse-sdk ethers
...
+
+const file = new TextEncoder().encode('hello from synapse')
+const result = await synapse.storage.upload(file)
+console.log('pieceCid:', result.pieceCid)
+```
+
+- Networks supported: `mainnet`, `calibration`, `devnet`
+- Devnet requires `rpcURL`, `warmStorageAddress`, and `multicall3Address` to be provided
+- For session keys, payments, and registry access, services auto-discover addresses on mainnet/calibration
+
```
hugomrdias: please update the docs folder, we will remove the sdk readme in favor of the docs

rvagg (approved):
@redpanda-f I hope you don't mind that I rebased this and pushed another commit with fixes for things I was noticing (cause I'm trying to make progress using foc-localnet for testing atm).

Approving but still untested .. I'll come back and merge if all goes well.

Issue #530, opened 2026-01-05
CI should run on dependency updates


rvagg:
https://github.com/FilOzone/synapse-sdk/blob/master/.github/workflows/ci.yml is only triggered by packages/* and ci.yml changes, so we don't see tests / lint run on dependency updates.

PR #531, opened 2026-01-05
fix(ci): trigger on dependency file changes

rvagg:
Fixes: https://github.com/FilOzone/synapse-sdk/issues/530


rvagg:
Added an additional commit:

    chore(deps): add trust policy excludes for redux ecosystem
    
    Ref: https://github.com/reduxjs/redux-toolkit/releases/tag/v2.11.1
    
    @reduxjs/toolkit@2.8.2 and reselect@5.1.1 were published manually and
    lost provenance. The team has switched to trusted publishing for new
    releases so this will resolve itself over time.

Issue #541, opened 2026-01-09
Enable debug logging of HTTP requests


BigLep:
## Done Criteria
It's possible to enable logging of all HTTP requests made by Synapse to help with debugging, specifically including:
- server
- path
- query string
- verb
- response code

## Why Important
When debugging Curio PDP issues for example, it's really useful to know what did the client send to the Curio PDP SP.  This is easy to get when Synapse is run within a browser context, but it would be nice to be able to get from the command line as well.

## User/Customer
Maintainers attempting to debug the system.

## Notes
Instances where this would have been useful:
- https://github.com/FilOzone/filecoin-services/issues/380

BigLep:
@rvagg @hugomrdias : I was wanting this today when working on https://github.com/FilOzone/filecoin-services/issues/380.  I slopped through this with https://github.com/FilOzone/synapse-sdk/pull/542 but I assume this isn't a good approach.

I'd like your input on whether there is already a good way to do this and if not, if this is worth doing...

BigLep:
> NODE_DEBUG=fetch

Thanks @rvagg .  I think that's good enough for now.  I'm documenting that in filecion-pin with https://github.com/filecoin-project/filecoin-pin/pull/300 .  Even though that doesn't add docs within Synapse itself, I think the filecoin-pin docs are good enough.  I'll resolve this issue.

PR #542, opened 2026-01-09
chore: enable HTTP logging of Curio calls

BigLep:
This was the hack I did for unblocking myself on https://github.com/FilOzone/synapse-sdk/issues/541 .  I'm sure it's not right though.  Just putting it here in case it's useful for discussion.


BigLep:
Closing because I don't expect this is the right path forward but I wanted an artifact could easily link to.

Issue #494, opened 2025-12-09
Multi-copy upload via SP-to-SP fetch


rvagg:
## Summary

Implement multi-copy storage where client uploads once to an endorsed SP, then secondary SPs fetch the piece via HTTP rather than
receiving duplicate uploads.

## Background

The current multi-copy approach requires parallel uploads of the same data, which is inefficient for large pieces, presents challenges
for streaming data and our ability to retry uploads, forces awkward API choices, and increases our chances of exposing error conditions to users over alternative options.

See https://www.notion.so/filecoindev/Multi-copy-Synapse-Upload-2bddc41950c1803d9fb6c8830c8a7a28 for background. Option 4 is what's 
being outlined here.

This flow is also a small step toward the MSP model that's currently being designed and we'll be working on early next year.

## New Flow

Depends on: https://github.com/filecoin-project/curio/issues/828 (SP-to-SP fetch endpoint)

1. Upload to endorsed SP (streaming supported)
   - `uploadPiece()` -> `findPiece()`
   - Failure here = hard fail, user must retry if they want
2. Construct retrieval URL from endorsed SP
3. Request secondary SP(s) to fetch piece
   - `POST /pdp/piece/fetch` (new Curio endpoint)
   - Poll until complete
   - On failure: retry with different SP (if auto-selected)
4. `AddPieces` on all SPs in parallel when all have the piece
5. Return result
   - ALL `AddPieces` failed = throw
   - Endorsed SP's `AddPieces` failed but secondary succeeded = throw? (Open: product requirement not met)
   - At least endorsed SP succeeded = return result object (don't throw)

As per the linked document, the benefits of this flow include:

- Streaming preserved: client sends data once, no need to buffer or replay
- Simpler retries: secondary SP failures just mean "try another SP", no re-upload, we have additional retry options that we can implement in future iterations of this
- Reduced client bandwidth: upload once instead of `N` times
- Clear failure semantics: endorsed SP failure = hard fail early, secondary failures = recoverable
- Cleaner API: easier to document and explain, no `Promise.allSettled` gymnastics, solve for *"don't make me understand your system"*

## API Changes

Simplified options and simplified internal resolver (separate issue):
- `dataSetIds` and `providerIds` mutually exclusive
- Counts must match `count` option
- No options = smart select (endorsed first)

New `PDPServer` methods:

* `fetchPieces(pieces, recordKeeper, extraData): Promise<FetchPiecesResponse>`
[...]

PR #544, opened 2026-01-12
feat(synapse-core): add SP-to-SP piece pull functionality

rvagg:
Ref: https://github.com/filecoin-project/curio/pull/864
Part of https://github.com/FilOzone/synapse-sdk/issues/494


utils/example-pull-e2e.js
rvagg: this file is on a separate commit and needs to be removed before merging this PR, but it does demonstrate how this can be tested against a devnet

rvagg:
latest fixup in here adds the ability to supply a pre-built `extraData` since we use it for auth but then use it again to addPieces and we don't want the user to have to sign twice for the same thing if we can help it

packages/synapse-core/src/warm-storage/pull.ts
hugomrdias: i think this should be inside the sp folder

packages/synapse-core/src/warm-storage/pull.ts
```diff
@@ -0,0 +1,244 @@
...
+
+export namespace waitForPullStatus {
+  /**
+   * Options for waiting for pull completion.
+   */
+  export type OptionsType = pullPieces.OptionsType & {
+    /** Callback invoked on each poll with current status. */
+    onStatus?: (response: SP.PullResponse) => void
+    /** Minimum time between poll attempts in milliseconds (default: 4000). */
+    minTimeout?: number
```
hugomrdias:
```suggestion
    /** The polling interval in milliseconds. Defaults to 4 seconds. */
    pollInterval?: number
```

PR #556, opened 2026-01-15
feat: add public calculatePieceCID() function

bajtos:
Add a unified `calculatePieceCID()` function that accepts either `Uint8Array` or `AsyncIterable<Uint8Array>` and automatically detects the input type to call the appropriate underlying function from `synapse-core`.

This allows synapse-sdk consumers to pre-calculate the PieceCID before calling the `upload` API.

Ref: https://github.com/filecoin-project/filecoin-pin/issues/282

### DIFF
--- a/packages/synapse-sdk/src/utils/piece.ts
+++ b/packages/synapse-sdk/src/utils/piece.ts
@@ -57,7 +57,7 @@ export function constructFindPieceUrl(apiEndpoint: string, pieceCid: PieceCID): 
  */
 export async function calculatePieceCID(data: Uint8Array | AsyncIterable<Uint8Array>): Promise<PieceCID> {
   // Check for Uint8Array first (before async iterable, since Uint8Array has Symbol.iterator)
-  if (isUint8Array(data)) {
+  if (data instanceof Uint8Array) {
     return calculateSync(data)
   }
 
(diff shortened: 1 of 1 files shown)


bajtos:
I am not fully convinced this is the right approach. A few alternatives that come to my mind:

- Move the `calculatePieceCID()` implementation & tests to `synapse-core` and simply re-export it in `synapse-sdk`.
- Don't add a new function, simply modify `synapse-sdk` to re-export `calculate` and `calculateFromIterable` from `synapse-core`, perhaps with more descriptive names.

Thoughts?

/cc @rvagg

hugomrdias:
Im not a fan of this the individual functions are already public from core.

From the referenced issue i dont see this as the solution, sdk and core provide piece cid as outputs everywhere.

bajtos:
@HavenCTO thank you for the detailed comment! 🙏🏻 

I see that your function `computePieceCidLocally` accepts an `Uint8Array` with the CAR bytes. That's helpful to know - it means my proposed API for `calculatePieceCID` should be directly usable in your project in case you decided to adopt it.

```ts
async function computePieceCidLocally(carBytes: Uint8Array, logger: Logger): Promise<string | undefined> {
  try {
    // TODO: carBytes wrap in another array to avoid SharedArrayBuffer?

    const pieceCid = await calculatePieceCID(carBytes)
 
    logger.info('Computed PieceCID locally from CAR', {
      pieceCid: pieceCidString,
      // NOTE: the following information is not provided by Synapse SDK API
      // pieceSize: piece.pieceSize,
      // payloadSize: piece.payloadSize,
    });
    return pieceCidString;
  } catch (error) {
    logger.warn('Local PieceCID computation failed', {
      error: error instanceof Error ? error.message : String(error),
    });
    return undefined;
  }
}
```

I noticed that you are logging `piece.pieceSize` and `piece.payloadSize` - I am wondering if this is something we need to add Synapse SDK helpers as well 🤔 Anyhow, such a new API is out of the scope of this pull request, let's open a new GH issue if anyone feels strongly about this.

PR #571, opened 2026-01-26
refactor: viem in session key and endorsements

hugomrdias:


packages/synapse-sdk/src/session/key.ts
```diff
@@ -21,103 +21,67 @@
...
+  get account(): Account {
+    return this._account
   }
 
-  getSigner(): ethers.Signer {
-    return this._signer
+  get chain(): Chain {
+    return this._chain
+  }
+  get client(): Client<Transport, Chain, Account> {
```
rvagg:
```suggestion
  }

  get client(): Client<Transport, Chain, Account> {
```

packages/synapse-sdk/src/session/key.ts
```diff
@@ -21,103 +21,67 @@
...
-        callData: registryInterface.encodeFunctionData('authorizationExpiry', [
-          ownerAddress,
-          signerAddress,
+  async fetchExpiries(
+    permissions: SK.SessionKeyPermissions[] = SK.ALL_PERMISSIONS
+  ): Promise<Record<SK.SessionKeyPermissions, bigint>> {
+    const expiries: Record<string, bigint> = {}
+    const result = await multicall(this._ownerClient, {
+      allowFailure: false,
+
```
rvagg:
```suggestion
```

packages/synapse-sdk/src/session/key.ts
```diff
@@ -21,103 +21,67 @@
...
+  private readonly _account: Account
+  private readonly _chain: Chain
+
+  public constructor(client: Client<Transport, Chain, Account>, account: Account) {
+    this._ownerClient = client
+    this._account = account
+    this._chain = asChain(client.chain)
+    this._client = createWalletClient({
+      chain: this._chain,
+      transport: http(),
```
rvagg: can we inherit the transport properly here? I _think_ this client isn't actually used for contract calls but it's exposed in client() so it could be, but we're just making our own bare http transport now which isn't what synapse was set up with

PR #573, opened 2026-01-27
refactor: use Chain type for FilBeam retriever config

bajtos:
Replace FilecoinNetworkType with Chain from synapse-core in FilBeamRetriever, reading the retrieval domain from chain configuration instead of hardcoding network-to-domain mapping.

Refs:
- This is a follow-up for #570
- Targeting #569


rvagg (approved):
lgtm, up to Hugo how it lands though cause it's on to his branch

PR #575, opened 2026-01-28
refactor: viem in sp registry

hugomrdias:
- rename contracts to filecoinPay, fwss, and fwssVier
- sp registry support in core
- sp registry service in sdk updated to use core


packages/synapse-core/src/sp-registry/types.ts
rvagg: would be a good place for a `ProductType.PDP` export

rvagg (changes requested):
just that ipnipeerid thing I think

PR #577, opened 2026-01-30
refactor: ethers gone

hugomrdias:
- context and synapse use viem
- session keys are disabled, will bring them back in GA Api change scope
- viem factory methods, public/wallet client will be improved in GA API scope

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,7 +59,7 @@ packages/synapse-sdk/src/
 **NO** semicolons at line end (`semicolons: "asNeeded"`)
 **MUST** use kebab-case filenames
 
-Run `pnpm run fix` before commits.
+Run `pnpm run lint:fix` before commits.
 
 ## Key Components
 


packages/synapse-core/src/warm-storage/get-service-price.ts
```diff
@@ -79,12 +80,12 @@ export async function getServicePrice(
...
-    /** Warm storage contract address. If not provided, the default is the storage contract address for the chain. */
-    contractAddress?: Address
-    /** The chain to use to get the service price. */
-    chain: Chain
-  }
+  export type OptionsType = Simplify<
+    getServicePrice.OptionsType & {
+      /** The chain to use to get the service price. */
+      chain: Chain
+    }
```
wjmelements: several call Options types have this `chain` option. Should `{ chain: Chain }` be a declared type?

docs/astro.config.mjs
```diff
@@ -174,10 +174,7 @@ export default defineConfig({
         plugins: [
           ecTwoSlash({
             twoslashOptions: {
-              compilerOptions: {
-                lib: ['ESNext', 'DOM'],
-                composite: true,
-              },
+              lib: ['esnext', 'dom'],
```
rvagg: having it here doesn't give us a free-pass for the inline `@lib`?

packages/synapse-sdk/src/test/epoch.test.ts
```diff
@@ -1,79 +1,63 @@
...
-  dateToEpoch,
-  epochToDate,
-  getGenesisTimestamp,
-  timeUntilEpoch,
-} from '../utils/epoch.ts'
+import { TIME_CONSTANTS } from '../utils/constants.ts'
+import { calculateLastProofDate, dateToEpoch, epochToDate, timeUntilEpoch } from '../utils/epoch.ts'
 
 describe('Epoch Utilities', () => {
   describe('epochToDate', () => {
```
rvagg: looks like some tests in here can either be moved to synapse-core or just deleted

rvagg (approved):
fine by me, sans the question about packages/synapse-sdk/src/test/epoch.test.ts extraneous tests, and William's comments too

Issue #567, opened 2026-01-22
Keep contract addresses in sync


rvagg:
We have 3 pieces where contract addresses get involved, discounting the current sdk constants.ts which is slowly being eaten away into oblivion.

* wagmi.config.ts which fetches ABIs from a pinned `GIT_REF` for ABI and an address each for FWSS on mainnet and calibnet. When run it makes a generated.ts with the addresses for the contracts it discovers via a multicall and their ABIs.
* chains.ts which imports generated.ts to build viem Chain objects
* filecoin-services [deployments.json](https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/deployments.json) which has canonical addresses updated on release of new contracts.

How do we make sure we have the right ABI and addresses here and we don't get out of sync?

One challenge is that we have been bumping that commit to pull in newer ABIs than are released, maybe we need to change that, or do it on feature branches that only get merged once we have a release. We could make that a practice post-v1.

Some options:

* Build-time validation script - `pnpm build` could fetch deployments.json and do an update for you and get everything in sync.
* CI linter in synapse-sdk - fails you when you're out of sync

Both of those options are reactive rather than proactive, they require someone to be making a change for anything to get updated, and they'd get mixed up in PRs that are entirely unrealted to the contract updates.

* Cron GHA in Synapse that opens a PR when it detects a change
* Release trigger on filecoin-services that opens a PR, or delegates to Synapse, to open a PR when a release has been cut
* Publish a new package, `@filoz/contract-addresses` from filecoin-services whenever we cut a release there and include deployments.json data in it, versioned the same as filecoin-services, depend on it in Synapse (but then what .. do we rebuild here, do we put the wagmi / generated stuff into that package so it does the building?)

hugomrdias:
1. and 2. are the same, come from the some place.

Our CI currently fails when an address changes in FWSS contract.

Feature branch sounds good!

- synapse master always in sync with filecoin-services master
- synpase next in sync with filecoin-services next

Questions:
- what is the source of truth for abi and addresses? git ref for abis, deployments.json for addresses?
- deployments.json probably does work for development, we will only have calibration
- is devnet the solution for development?

rvagg:
* Source of truth: git ref, I'd use a version tag but we can use a hash if we want to be uber-secure, it'll have ABIs and deployments.json is now _the_ place they'll be recorded.
[...]

PR #578, opened 2026-01-30
feat(docs): render contract addresses from @filoz/synapse-core/chains

nijoe1:
Replace static contract address tables with an Astro component that imports chain data directly from `@filoz/synapse-core/chains` at build time.
- Add `contract-addresses.astro` component that renders address tables from chain definitions
- Convert `contracts.md` to `contracts.mdx` to use the component
- Add `@filoz/synapse-core` as a docs workspace dependency
- Remove `deployments.json` sync script and GitHub workflow

Docs now stay in sync with the SDK source of truth automatically.

Related: FilOzone/filecoin-cloud#217

cc: @BigLep


rvagg (changes requested):
See https://github.com/FilOzone/synapse-sdk/issues/567 for a fuller set of concerns and discussion about  this topic. Some initial feedback from me:

1. Although that issue is concerned with SDK contract addresses, they should be in sync with the docs. This PR only syncs contracts.md from deployments.json. After this PR we would have two mechanisms for updating contract addresses and that's a no-go.

2. This PR also just fetches from `main` indiscriminantly, that's not necessarily the contracts we want to be publishing here, we probably want to be more intentional and either fetch from the GH releases API, or use repository_dispatch from filecoin-services to come back here. The metadata.commit field in deployments.json could help here if we go the releases route.

3. The failure mode I talked about in #567 is a concern and this PR suffers from it. Unrelated work is going to be held up just because there's a new deployments.json, creating hassle and confusion.

I think there's a few paths forward here, but whatever we do we need to integrate both code & docs. First thing might be to sync the docs _from_ generated.ts, so wagmi.config.ts is the one source of truth here. Second, we need a trigger to perform updates here in a way that doesn't hold up unrelated work. I'm partial to the release trigger on filecoin-services creating a PR to this repo to do the update, but see #567 for other options, or discuss sync in one of the team meetings with @hugomrdias || me.

nijoe1:
@rvagg what about making sure that docs are always synced with the addresses inside the latest version of the sdk ?

If that's a path to move forward this should be the same CI in filecoin.cloud which now syncs using filecoin-services deployments.json

hugomrdias:
yep docs should be in sync with `@filoz/synapse-core/chains`, you can import the package directly and render in astro from it.

keeping `@filoz/synapse-core/chains` in sync with filecoin-services can be figured out in #567

nijoe1:
Requested changes got pushed.

cc: @rvagg @hugomrdias

rvagg (approved):

PR #600, opened 2026-02-10
init: `foc-devnet-info` library in synapse-core

redpanda-f:
Introduce: `foc-devnet-info` library
- Parses and produces synapse type `Chain` from `DevnetInfo` exported by `foc-devnet`
- `e2e` test uses this library now, as an example to interface with `foc-devnet`
- mocked tests to check for functionality


packages/synapse-core/src/foc-devnet-info/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+ * // In browser
+ * import { loadDevnetInfo } from '@filoz/synapse-core/foc-devnet-info';
+ *
+ * const response = await fetch('/devnet-info.json');
+ * const data = await response.json();
+ * const devnetInfo = loadDevnetInfo(data);
+ */
+export function loadDevnetInfo(data: unknown): VersionedDevnetInfo {
+  return validateDevnetInfo(data)
+}
```
rvagg: pointless pass-through function, collapse into one

package.json
```diff
@@ -39,7 +39,7 @@
   "devEngines": {
     "runtime": {
       "name": "node",
-      "version": "^24.13.0",
+      "version": "^24.9.0",
```
rvagg: the answer here is to install a newer Node.js, not downgrade this

I kind of hate this completely, but this isn't the right place to change it

packages/synapse-core/src/index.ts
```diff
@@ -15,6 +15,7 @@ export * as chains from './chains.ts'
 export * as endorsements from './endorsements/index.ts'
 export * as erc20 from './erc20/index.ts'
 export * as errors from './errors/index.ts'
+export * as focDevnetInfo from './foc-devnet-info/src/index.ts'
```
rvagg:
```suggestion
```

just as a stand-alone `@filoz/synapse-core/foc-devnet-info` export is enough

packages/synapse-core/test/foc-devnet-info.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      const result = loadDevnetInfo(exampleDevnetInfoV1)
+      const user1 = result.info.users[0]
+
+      assert.equal(user1.name, 'USER_1')
+      assert.ok(user1.evm_addr.startsWith('0x'))
+      assert.ok(user1.native_addr.startsWith('t410f'))
+      assert.ok(user1.private_key_hex.startsWith('0x'))
+    })
+
+    it('should validate contract addresses', () => {
```
rvagg: ditto, testing zod

utils/example-storage-e2e.js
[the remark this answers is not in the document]
rvagg: Sorry, most minimal would be this since you don't need privkey: `NETWORK=devnet node example-storage-e2e.js file.txt`

packages/synapse-core/src/foc-devnet-info/src/index.ts
```diff
@@ -0,0 +1,168 @@
+/**
+ * @module @filoz/synapse-core/foc-devnet-info
+ *
+ * Library for validating and transforming FOC devnet configuration exports.
+ * Environment-agnostic - works in both Node.js and browsers.
```
rvagg:
```suggestion
 * Validates and transforms foc-devnet's devnet-info.json into Synapse-compatible configuration.
 * See https://github.com/FilOzone/foc-devnet for the export format.
```

utils/example-storage-e2e.js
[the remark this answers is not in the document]
rvagg: @redpanda-f see https://github.com/FilOzone/synapse-sdk/pull/604 for a PR against this file that works on your branch and merges in some #593 code too.

redpanda-f:
@rvagg request re-review. Especially biome.json? not opinionated, but needs fixing.

PR #601, opened 2026-02-10
feat!: change params to options object and remove withIpni

hugomrdias:
- changes most services to options object
- removes deprecated methods
- migration guide

### DIFF
--- a/docs/src/content/docs/developer-guides/payments/payment-operations.mdx
+++ b/docs/src/content/docs/developer-guides/payments/payment-operations.mdx
@@ -49,7 +49,7 @@ const synapse = Synapse.create({ account: privateKeyToAccount('0x...') });
 import { TOKENS } from "@filoz/synapse-sdk";
 
 const amount = parseUnits("100");
-const hash = await synapse.payments.depositWithPermit({amount});
+const hash = await synapse.payments.depositWithPermit({ amount });
 await synapse.client.waitForTransactionReceipt({ hash });
 ```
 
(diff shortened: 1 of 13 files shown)


docs/src/content/docs/developer-guides/payments/payment-operations.mdx
```diff
@@ -49,7 +49,7 @@ const synapse = Synapse.create({ account: privateKeyToAccount('0x...') });
 import { TOKENS } from "@filoz/synapse-sdk";
 
 const amount = parseUnits("100");
-const hash = await synapse.payments.depositWithPermit(amount);
+const hash = await synapse.payments.depositWithPermit({amount});
```
rvagg: we don't have a lint rule on mdx files? this should be spaced to be consistent with the rest of the docs and with the linter rules for the code I think, 3 instances of `{amount}` here, docs/src/content/docs/developer-guides/payments/rails-settlement.mdx has the same problem, the others look fine I think

packages/synapse-core/src/pay/payments.ts
```diff
@@ -93,7 +93,7 @@ export async function depositAndApprove(client: Client<Transport, Chain, Account
     throw new InsufficientBalanceError(balance, options.amount)
   }
 
-  const deadline = options.deadline ?? BigInt(Math.floor(Date.now() / 1000) + 3600) // 1 hour
+  const deadline = options.deadline ?? BigInt(Math.floor(Date.now() / 1000) + TIME_CONSTANTS.PERMIT_DEADLINE_DURATION)
```
rvagg: 👍

packages/synapse-sdk/src/payments/service.ts
```diff
@@ -7,21 +7,28 @@ import {
...
 /**
  * Options for deposit operation
  */
 export interface DepositOptions {
-  /** Optional recipient address (defaults to signer address if not provided) */
+  /** Optional recipient address (defaults to wallet address if not provided) */
   to?: Address
+  /** The amount to deposit */
+  amount: TokenAmount
+  /** The recipient address for the deposit. If not provided, the sender's address will be used. */
```
rvagg:
```suggestion
```
?

PR #616, opened 2026-02-18
fix: validate operator address and allowance for custom operators

### LINKED ISSUE #613: Security: Validate operator address and allowance when custom operator is provided in approve operations
### Problem
The `approve` and `depositWithPermitAndApproveOperator` operators in both `synapse-sdk` and `synapse-core` have an optional `operator` and `allowances` parameter. Currently, we default to the FWSS contract operator with `maxUint256` allowance, which is safe since this is a trusted and audited operator.

However, there's a critical edge case vulnerability:

**When a user provides a custom `operator` address but leaves `allowances` undefined, the system still defaults to `maxUint256` allowance for this arbitrary operator.**

This is problematic because:
- The custom operator could be any arbitrary address (malicious contract or EOA)
- Granting `maxUint256` allowance to an untrusted operator creates an immediate fund draining risk
- The API currently allows this dangerous combination without validation

### Affected Code

**`approve` operator:**
- [`synapse-sdk` approve service](https://github.com/FilOzone/synapse-sdk/blob/master/packages/synapse-sdk/src/payments/service.ts#L264) 
- Calls [`synapse-core` set-operator-approval](https://github.com/FilOzone/synapse-sdk/blob/master/packages/synapse-core/src/pay/set-operator-approval.ts#L206)

**`depositWithPermitAndApproveOperator`:**
- [`synapse-sdk` payments service](https://github.com/FilOzone/synapse-sdk/blob/master/packages/synapse-sdk/src/payments/service.ts#L521)
- Calls [`synapse-core` payments](https://github.com/FilOzone/synapse-sdk/blob/master/packages/synapse-core/src/pay/payments.ts#L64)

### Proposed Solution
Add validation that throws an error when:
- A custom `operator` address is provided (different from the default FWSS operator), AND
- The `allowances` parameter is left undefined

This forces users to explicitly specify allowances for custom operators, preventing accidental unlimited approvals to untrusted addresses.

### Severity
Medium - While this is an edge case, it represents a significant security risk if exploited.

cc: @hugomrdias @BigLep @rvagg @rjan90

Chaitu-Tatipamula:
Closes #613 

**Problem**
When a custom operator is provided without explicit allowances, the system defaults to maxUint256, granting unlimited spending power to a potentially untrusted address.

**Solution**
Added `ValidationError` in `setOperatorApprovalCall` and `depositAndApprove` that throws when a non-default operator is used without explicit `rateAllowance` and `lockupAllowance`. Default FWSS operator behavior is unchanged.

**Changes**
`synapse-core/src/pay/set-operator-approval.ts` _— validation on approve_
`synapse-core/src/pay/payments.ts` — _validation on deposit+approve_
`synapse-core/test/set-operator-approval.test.ts` — _3 new tests, 2 updated_

### DIFF
--- a/packages/synapse-core/src/pay/payments.ts
+++ b/packages/synapse-core/src/pay/payments.ts
@@ -1,4 +1,13 @@
-import { type Account, type Address, type Chain, type Client, maxUint256, parseSignature, type Transport } from 'viem'
+import {
+  type Account,
+  type Address,
+  type Chain,
+  type Client,
+  isAddressEqual,
+  maxUint256,
+  parseSignature,
+  type Transport,
+} from 'viem'
 import {
   type SimulateContractErrorType,
   simulateContract,
@@ -66,10 +75,15 @@ export async function depositAndApprove(client: Client<Transport, Chain, Account
   const token = options.token ?? chain.contracts.usdfc.address
   const operator = options.operator ?? chain.contracts.fwss.address
 
-  const isCustomOperator = options.operator !== undefined && options.operator !== chain.contracts.fwss.address
+  const isCustomOperator =
+    options.operator !== undefined && !isAddressEqual(options.operator, chain.contracts.fwss.address)
   if (isCustomOperator) {
-    if (options.rateAllowance === undefined || options.lockupAllowance === undefined) {
-      throw new ValidationError('Custom operator requires explicit rateAllowance and lockupAllowance')
+    if (
+      options.rateAllowance === undefined ||
+      options.lockupAllowance === undefined ||
+      options.maxLockupPeriod === undefined
+    ) {
+      throw new ValidationError('Custom operator requires explicit rateAllowance, lockupAllowance and maxLockupPeriod')
     }
   }
 
--- a/packages/synapse-core/src/pay/set-operator-approval.ts
+++ b/packages/synapse-core/src/pay/set-operator-approval.ts
@@ -12,7 +12,7 @@ import type {
   WaitForTransactionReceiptErrorType,
   WriteContractErrorType,
 } from 'viem'
-import { maxUint256, parseEventLogs } from 'viem'
+import { isAddressEqual, maxUint256, parseEventLogs } from 'viem'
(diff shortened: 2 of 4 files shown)


packages/synapse-core/src/pay/payments.ts
```diff
@@ -65,6 +65,17 @@ export async function depositAndApprove(client: Client<Transport, Chain, Account
   const chain = getChain(client.chain.id)
   const token = options.token ?? chain.contracts.usdfc.address
   const operator = options.operator ?? chain.contracts.fwss.address
+
+  const isCustomOperator = options.operator !== undefined && options.operator !== chain.contracts.fwss.address
+  if (isCustomOperator) {
+    if (options.rateAllowance === undefined || options.lockupAllowance === undefined) {
+      throw new ValidationError(
+        'Custom operator requires explicit rateAllowance and lockupAllowance. ' +
+          'Defaulting to maxUint256 for untrusted operators is not allowed.'
```
rvagg:
```suggestion
        'Custom operator requires explicit rateAllowance and lockupAllowance'
```

let's leave this unsaid

rvagg (approved):

Chaitu-Tatipamula:
> Actually, sorry @Chaitu-Tatipamula, thinking about this a bit more:
> 
> * `import { isAddressEqual } from 'viem'` is apparently a thing we should use for address comparison, we have the case sensitivity thing which is mostly not a big deal here but a hypothetical situation here is that the user supplies the fwss address but with different casing, it shouldn't be caught by this.
> * I think we should lump `maxLockupPeriod` in here as well, 30 days is FWSS-specific, let's just include it so if you set approval then it should deal with all 3
> * Your tests only touch `setOperatorApprovalCall` but you also touch `depositAndApprove` in here - either extract a common utility or test them both; since the tests are in different files, probably just do them both separately

yeah these are great to have
isAddressEqual — makes total sense, I'll switch to that to handle the case sensitivity edge case.
maxLockupPeriod — since 30 days is FWSS specific it should be explicit for custom operators. Will add it to the validation alongside the other two.
Tests for depositAndApprove — Will add a separate test file for it.

PR #618, opened 2026-02-19
feat: new session keys with viem

hugomrdias:
- [x] still needs wiring in react and sdk
- [x] proper e2e testing 

Example usage:

 ```ts
 import * as SessionKey from '@filoz/synapse-core/session-key'
 import { privateKeyToAccount } from 'viem/accounts'
 import { type Hex } from 'viem'
 import { mainnet } from '@filoz/synapse-core/chains'
 
 const rootAccount = privateKeyToAccount('0xaa14e25eaea762df1533e72394b85e56dd0c7aa61cf6df3b1f13a842ca0361e5' as Hex)
 
 const sessionKey = SessionKey.fromSecp256k1({
   privateKey: '0xaa14e25eaea762df1533e72394b85e56dd0c7aa61cf6df3b1f13a842ca0361e5' as Hex,
   root: rootAccount,
   chain: mainnet,
 })
 sessionKey.on('expirationsUpdated', (e) => {console.log(e.detail)})
 sessionKey.on('connected', (e) => {console.log(e.detail)})
 sessionKey.on('disconnected', () => {console.log('disconnected')})
 sessionKey.on('error', (e) => {console.log(e.detail)})
 
 
 const { event: loginEvent } = await SessionKey.loginSync(client, {
   address: sessionKey.address,
   onHash(hash) {
     console.log(`Waiting for tx ${hash} to be mined...`)
   },
 })
 
 console.log('event login:', loginEvent.args)
 
 await sessionKey.connect()
 
 if(sessionKey.hasPermission(SessionKey.CreateDataSetPermission)) {
   const hash = await createDataSet(sessionKey.client, {
     payee: '0x1234567890123456789012345678901234567890',
     payer: sessionKey.rootAddress,
     serviceURL: 'https://example.com',
   })
   console.log('event created dataset:', hash)
 }
 
 const { event: revokeEvent } = await SessionKey.revokeSync(client, {
   address: sessionKey.address,
   onHash(hash) {
     console.log(`Waiting for tx ${hash} to be mined...`)
   },
 })
 console.log('event revoked:', revokeEvent.args)
 sessionKey.disconnect()
 ```

### DIFF
--- a/packages/synapse-core/src/session-key/secp256k1.ts
+++ b/packages/synapse-core/src/session-key/secp256k1.ts
@@ -68,7 +68,7 @@ class Secp256k1SessionKey extends TypedEventTarget<SessionKeyEvents> implements 
     return this.#client.account
   }
 
-  async connect() {
+  async watch() {
     await this.syncExpirations()
 
     if (!this.#unsubscribe) {
@@ -77,7 +77,7 @@ class Secp256k1SessionKey extends TypedEventTarget<SessionKeyEvents> implements 
         abi: this.#client.chain.contracts.sessionKeyRegistry.abi,
         eventName: 'AuthorizationsUpdated',
         args: { identity: this.#client.account.rootAddress },
-
+        onError: this.emit.bind(this, 'error'),
         onLogs: (logs) => {
           try {
             const event = extractLoginEvent(logs)
@@ -94,9 +94,12 @@ class Secp256k1SessionKey extends TypedEventTarget<SessionKeyEvents> implements 
       })
       this.emit('connected', this.#expirations)
     }
+    return () => {
+      this.unwatch()
+    }
   }
 
-  disconnect() {
+  unwatch() {
     if (this.#unsubscribe) {
       this.#unsubscribe()
       this.#unsubscribe = undefined
--- a/packages/synapse-core/src/session-key/types.ts
+++ b/packages/synapse-core/src/session-key/types.ts
@@ -32,6 +32,14 @@ export interface SessionKey<KeyType extends SessionKeyType> extends TypedEventTa
   hasPermission: (permission: Permission) => boolean
   hasPermissions: (permissions: Permission[]) => boolean
   syncExpirations: () => Promise<void>
-  connect: () => Promise<void>
-  disconnect: () => void
+  /**
+   * Watch the session key for expirations updates.
+   *
+   * @returns A function to stop watching the session key.
+   */
+  watch: () => Promise<() => void>
+  /**
+   * Stop watching the session key for expirations updates.
+   */
+  unwatch: () => void
 }
--- a/examples/cli/src/commands/session-keys.ts
+++ b/examples/cli/src/commands/session-keys.ts
@@ -2,7 +2,7 @@ import * as p from '@clack/prompts'
 import * as SessionKey from '@filoz/synapse-core/session-key'
 import { createDataSet, waitForCreateDataSet } from '@filoz/synapse-core/sp'
 import { type Command, command } from 'cleye'
-import type { Hex } from 'viem'
+import { type Hex, stringify } from 'viem'
 import { privateKeyClient } from '../client.ts'
 import { globalFlags } from '../flags.ts'
 import { hashLink } from '../utils.ts'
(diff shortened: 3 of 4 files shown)


packages/synapse-core/src/session-key/revoke.ts
```diff
@@ -0,0 +1,143 @@
...
+  } satisfies revokeCall.OutputType
+}
+
+/**
+ * Extracts the AuthorizationsUpdated event from transaction logs.
+ *
+ * @param logs - The transaction logs.
+ * @returns The AuthorizationsUpdated event.
+ */
+export function extractRevokeEvent(logs: Log[]) {
```
rvagg: could just reuse the login one to do the same job?

packages/synapse-core/src/session-key/types.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { SessionKeyPermissions } from './permissions.ts'
+
+export type SessionKeyEvents = {
+  expirationsUpdated: CustomEvent<Record<SessionKeyPermissions, bigint>>
+  connected: CustomEvent<Record<SessionKeyPermissions, bigint>>
+  disconnected: CustomEvent<void>
+  error: CustomEvent<Error>
+}
+
+export type SessionKeyType = 'Secp256k1' | 'P-256'
```
rvagg: oh really ...?

rvagg:
This might be a good opportunity to deal with `type SessionKeyPermissions` being a closed union and `ALL_PERMISSIONS` being a bit misleading in its name. This came up this week in https://github.com/filecoin-project/curio/issues/1028, noting that (a) SessionKeyRegistry doesn't care what these strings are, it's not exclusive, and (b) `DeleteDataSet` isn't even standard anymore, it's gone, so our existing set isn't even accurate (but as per that issue we're talking about repurposing it to do Client<>Curio auth, not touching FWSS).

Perhaps:

* `ALL_PERMISSIONS` -> `FWSS_PERMISSIONS`
* `EMPTY_PERMISSIONS` has the same problem
* Make `SessionKeyPermissions` open, or rename it to `FWSSPermissions` and do a `type SessionKeyPermission = FWSSPermissions | Hex`.
* `login()`, `revoke()` should expand their `permissions?` as per above.
* `getExpirations()` and `hasPermission()` is harder, but we could clearly document that they are dealing specifically with FWSS only

Not a blocker, but the current session key code gives the impression this is all canonical and a closed set of permissions. But developers should be able to use the registry to do non-FWSS things and having an SDK that makes that easy would be nice.

hugomrdias:
> Will I be able to pass `sessionKey.client` to the SDK in the end of this? For the signing operations, will we need to update them to differentiate between a SessionKeyAccount client and a standard one so we can pick out `rootAddress` when it exists?

Yeah not sure yet, because session keys only works for those 4 mutations (maybe just 3 if delete dataset is gone), some of the other services do other mutations that need the root client.

I think for now I'm just adding a new constructor option to pass a session keys and internally use if available and has permission.

There's one subtle thing in this design, that I want your opinion.

The connect/disconnect are actually optional, they are there if you want realtime tracking of the permissions, if you don't you can just pass whatever expirations you got from login and YOLO the mutations.


I will be cleaning up with your comments and finish testing hopefully sps work today.

rvagg:
> you can just pass whatever expirations you got from login and YOLO the mutations

Pass to _where_? I don't think we care in any of the places we use it, so this isn't an internal concern. I think the answer here is to just document how expirations work and for long-lived session key instances advise that the developer should be sure to refresh expiries occasionally to be sure to perform `login()` as needed. Otherwise, not our problem (except for the fact that errors arising from expired permissions will be opaque about the cause!).

packages/synapse-core/src/session-key/secp256k1.ts
```diff
@@ -1,159 +1,217 @@
...
+        eventName: 'AuthorizationsUpdated',
+        args: { identity: this.#client.account.rootAddress },
+        onLogs: (logs) => {
+          const event = extractLoginEvent(logs)
+          if (event.args.identity === this.#client.account.rootAddress) {
+            for (const hash of event.args.permissions) {
+              const permission = getPermissionFromTypeHash(hash)
+              this.expirations[permission] = event.args.expiry
+            }
+            this.emit('expirationsUpdated', this.#expirations)
```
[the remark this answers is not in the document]
hugomrdias: What do you mean? it will throw an error. User still gets the connected event if syncExpirations doesnt fail.

rvagg (approved):
With some suggestions, and some slightly more _firm_ requests which are probably: missing `onError` handling, unless you want to defend not doing that? The `Synapse.create` could do with some attention while you're at it. I also wouldn't mind hearing more justification on `connect()` / `disconnect()` being in core and not a react specialty (I'm imagining "there are other long-running uses out there" .. but .. really, it's pretty much only going to be React right?).

I'm going to follow up with a docs PR to this.

PR #632, opened 2026-02-25
feat!: New synapse costs API

### LINKED ISSUE #360: StorageInfo from getStorageInfo() should reflect new storage pricing

lordshashank:
Resolves #360 , #437

Complete architecture and thought process:  https://hackmd.io/@lordforever/SJ6Lwv2dWx (plz read to understand better and add comments)

### DIFF
--- a/packages/synapse-sdk/src/storage/manager.ts
+++ b/packages/synapse-sdk/src/storage/manager.ts
@@ -679,7 +679,7 @@ export class StorageManager {
     // Identify existing datasets that need size lookups
     const existingDataSetIds = contexts.filter((ctx) => ctx.dataSetId != null).map((ctx) => ctx.dataSetId as bigint)
 
-    // Fetch all needed data in parallel (single RPC batch)
+    // Fetch all needed data in parallel
     const [accountInfo, pricing, approved, currentEpoch, sizes] = await Promise.all([
       payAccounts(client, { address: clientAddress }),
       getServicePrice(client),


docs/src/content/docs/developer-guides/storage/storage-costs.mdx
```diff
@@ -90,164 +90,86 @@ const totalCostFor24Months = totalCostPerMonth * 24;
...
+- **`costs`**: The full `UploadCosts` breakdown
+- **`transaction`**: A transaction object with `execute()`, or `null` if the account is already ready
 
-console.log("Account funded successfully");
-```
+The transaction automatically picks the right contract call:
+- **Needs deposit + approval**: calls `depositWithPermitAndApproveOperator`
+- **Needs approval only**: calls `approveService`
+- **Needs deposit only**: calls `depositWithPermit`
+- **Already ready**: returns `transaction: null`
```
juliangruber: Should exact API specs live outside of developer guides?

packages/synapse-core/src/pay/account-debt.ts
```diff
@@ -0,0 +1,45 @@
...
+    currentEpoch: bigint
+  }
+
+  export type OutputType = {
+    /** max(0, totalOwed - funds) */
+    debt: bigint
+    /** max(0, funds - totalOwed) */
+    availableFunds: bigint
+    /** Epoch when account runs out (maxUint256 if lockupRate == 0) */
+    fundedUntilEpoch: bigint
```
juliangruber: these two are out of scope for the function. The consumer can simply do this themselves:

```js
const [debt, info] = await Promise.all([
  calculateAccountDebt(),
  getAccountInfoIfSettled()
])
```

I would suggest to only return the `debt` from this function and not return other possibly useful but already easy to get information.

packages/synapse-core/src/warm-storage/calculate-additional-lockup.ts
```diff
@@ -0,0 +1,96 @@
...
+    /** Current total data size in the existing dataset, in bytes. 0n for new datasets. */
+    currentDataSetSize: bigint
+    /** Price per TiB per month from getServicePrice(). */
+    pricePerTiBPerMonth: bigint
+    /** Minimum monthly charge from getServicePrice(). */
+    minimumPricePerMonth: bigint
+    /** Epochs per month from getServicePrice(). */
+    epochsPerMonth: bigint
+    /** Lockup period in epochs (default: 86400 = 30 days). */
+    lockupEpochs: bigint
```
juliangruber: can we signal in the types which parameters are optional / have a default and which are required?

packages/synapse-core/src/warm-storage/calculate-additional-lockup.ts
```diff
@@ -0,0 +1,96 @@
...
+    // Floor-to-floor: if both sizes are below floor, delta is 0
+    if (rateDeltaPerEpoch < 0n) rateDeltaPerEpoch = 0n
+  } else {
+    // New dataset or unknown current size: full rate for new data
+    const newRate = calculateEffectiveRate({
+      ...rateParams,
+      sizeInBytes: dataSize,
+    })
+    rateDeltaPerEpoch = newRate.ratePerEpoch
+  }
```
juliangruber: refactor: less branches

```suggestion
  const newRate = calculateEffectiveRate({
    ...rateParams,
    sizeInBytes: currentDataSetSize + dataSize,
  })
  const currentRate = calculateEffectiveRate({
    ...rateParams,
    sizeInBytes: currentDataSetSize,
  })
  rateDeltaPerEpoch = newRate.ratePerEpoch - currentRate.ratePerEpoch
```

packages/synapse-core/src/pay/deposit-with-permit.ts
```diff
@@ -0,0 +1,210 @@
...
+      structuredSignature.s,
+    ],
+  })
+
+  return writeContract(client, request)
+}
+
+export namespace depositWithPermitSync {
+  export type OptionsType = Simplify<depositWithPermit.OptionsType & ActionSyncCallback>
+  export type OutputType = ActionSyncOutput<typeof extractDepositWithPermitEvent>
```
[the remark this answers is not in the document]
lordshashank: `extractDepositWithPermitEvent` is defined below in the same file and the `typeof`  is needed because this is a type declaration, not a value expression. wdym here?

packages/synapse-core/src/warm-storage/calculate-additional-lockup-required.ts
```diff
@@ -0,0 +1,96 @@
...
+    // Floor-to-floor: if both sizes are below floor, delta is 0
+    if (rateDeltaPerEpoch < 0n) rateDeltaPerEpoch = 0n
+  } else {
+    // New dataset or unknown current size: full rate for new data
+    const newRate = calculateEffectiveRate({
+      ...rateParams,
+      sizeInBytes: dataSize,
+    })
+    rateDeltaPerEpoch = newRate.ratePerEpoch
+  }
```
juliangruber: what about this then?

```suggestion
  const newRatePerEpoch = calculateEffectiveRate({
    ...rateParams,
    sizeInBytes: currentDataSetSize + dataSize,
  }).ratePerEpoch
  const currentRatePerEpoch = isNewDataset ? 0 : calculateEffectiveRate({
    ...rateParams,
    sizeInBytes: currentDataSetSize,
  }).ratePerEpoch
  rateDeltaPerEpoch = newRatePerEpoch - currentRatePerEpoch
```

packages/synapse-core/src/warm-storage/calculate-deposit-needed.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @returns The runway amount in token base units
+ */
+export function calculateRunwayAmount(params: calculateRunwayAmount.ParamsType): bigint {
+  return params.netRate * params.runwayEpochs
+}
+
+export namespace calculateBufferAmount {
+  export type ParamsType = {
+    /** additionalLockup + runwayAmount + debt - availableFunds (before clamping to 0). */
+    rawDepositNeeded: bigint
```
juliangruber: Is this an optional argument? And what does this argument mean? I can see it's also talked about in the doc, but I'm not finding an explanation for what it means

packages/synapse-core/src/warm-storage/calculate-deposit-needed.ts
```diff
@@ -0,0 +1,142 @@
...
+export namespace calculateDepositNeeded {
+  export type ParamsType = {
+    // Upload parameters (passed to calculateAdditionalLockupRequired)
+    dataSize: bigint
+    currentDataSetSize: bigint
+    pricePerTiBPerMonth: bigint
+    minimumPricePerMonth: bigint
+    epochsPerMonth: bigint
+    lockupEpochs: bigint
+    isNewDataset: boolean
```
juliangruber:
```suggestion
    isNewDataSet: boolean
```

juliangruber (changes requested):

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -0,0 +1,125 @@
...
+    /** Whether CDN is enabled. Default: false */
+    withCDN?: boolean
+    /** Current total data size in the existing dataset, in bytes. */
+    currentDataSetSize?: bigint
+
+    /** Size of new data to upload, in bytes. */
+    dataSize: bigint
+
+    /** Extra runway in epochs beyond the required lockup. */
+    runwayEpochs?: bigint
```
juliangruber:
```suggestion
    extraRunwayEpochs?: bigint
```

packages/synapse-core/src/pay/fund.ts
```diff
@@ -0,0 +1,166 @@
...
+        lockupAllowance: maxUint256,
+        maxLockupPeriod: LOCKUP_PERIOD,
+      })
+    }
+  } else if (options.amount > 0n) {
+    return depositWithPermit(client, { amount: options.amount })
+  }
+
+  // Already approved, no deposit needed
+  return '0x' as Hash
```
hugomrdias: it should always deposit, never return this hash.

```js
  if (options.amount <= 0n) {
    throw new DepositAmountError(options.amount)
  }
```
i think this at the top should fix this like the other deposit methods

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -0,0 +1,125 @@
...
+  })
+
+  // Calculate deposit needed
+  const depositNeeded = calculateDepositNeeded({
+    dataSize: options.dataSize,
+    currentDataSetSize,
+    pricePerTiBPerMonth: pricing.pricePerTiBPerMonthNoCDN,
+    minimumPricePerMonth: pricing.minimumPricePerMonth,
+    epochsPerMonth: pricing.epochsPerMonth,
+    lockupEpochs: LOCKUP_PERIOD,
```
hugomrdias: yep constant, just remove the option and use the constant inside. this is FWSS calculations doesnt need to be generic

packages/synapse-core/src/warm-storage/get-upload-costs.ts
juliangruber: I understand but that's how it reads

packages/synapse-core/src/warm-storage/calculate-additional-lockup-required.ts
lordshashank: still you are missing the case of  `currentDataSetSize === 0n && !isNewDataset` , i know there has been assumption that this can't happen but I would still like to handle this case as people can upload one piece and then remove it leading to same case, and there is no loss in being defensive here

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -0,0 +1,125 @@
...
+    bufferEpochs?: bigint
+  }
+
+  export type OutputType = {
+    /** Effective rate for the dataset after adding dataSize bytes. */
+    rate: {
+      /** Rate per epoch — matches on-chain PDP rail rate. */
+      perEpoch: bigint
+      /** Rate per month — full precision for display. */
+      perMonth: bigint
```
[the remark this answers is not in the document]
lordshashank: this is correct, I am adding few more comments in code to explain this better, filpay rail calcn uses per epoch but the fwss pricing is per month

packages/synapse-core/src/pay/total-account-rate.ts
```diff
@@ -0,0 +1,72 @@
...
+    address: options.address,
+    token: options.token,
+    contractAddress: options.contractAddress,
+  })
+
+  const ratePerEpoch = accountInfo.lockupRate
+
+  return {
+    ratePerEpoch,
+    ratePerMonth: ratePerEpoch * TIME_CONSTANTS.EPOCHS_PER_MONTH,
```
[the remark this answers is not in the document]
lordshashank: adding few more ocmments in code to explain this

packages/synapse-core/src/warm-storage/calculate-effective-rate.ts
```diff
@@ -0,0 +1,52 @@
...
+export namespace calculateEffectiveRate {
+  export type ParamsType = {
+    /** Total data size in the dataset (existing + new), in bytes. */
+    sizeInBytes: bigint
+    /** Price per TiB per month from getServicePrice(). */
+    pricePerTiBPerMonth: bigint
+    /** Minimum monthly charge from getServicePrice(). */
+    minimumPricePerMonth: bigint
+    /** Epochs per month from getServicePrice() (always 86400). */
+    epochsPerMonth: bigint
```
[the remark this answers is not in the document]
lordshashank: making it optional

packages/synapse-core/src/warm-storage/get-upload-costs.ts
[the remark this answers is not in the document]
hugomrdias: @juliangruber can we resolve this ?

PR #675, opened 2026-03-17
fix: support multi sp upload flow in react and demo

### LINKED ISSUE #646: Update synapse-react to support multi sp upload flow
either:
- use core only, may be useful to validate the primitives we have there but more work 
- use sdk, easy path sdk already does all the work

Demo app should also be updated to use it.

hugomrdias:
closes #646


juliangruber:
Bug discovered: https://github.com/FilOzone/synapse-sdk/issues/678
Hugo requested to file an issue so this can be solved later

Issue #678, opened 2026-03-17
Firefox upload regression


juliangruber:
The playground app's upload function used to work in Firefox, up until https://github.com/FilOzone/synapse-sdk/pull/675 was (or very soon will be) merged. The error that reliably happens in Firefox (not in Chrome for example) is this:

```
StoreError: Failed to store on primary provider 9 (https://calib.ezpdpz.net/)

Details: StorageContext store failed: Failed to store piece on service provider - Failed to upload piece.

Details: Service Provider PDP
Failed to upload piece: Failed to finalize commP calculation
    SynapseError base.ts:38
    StoreError storage.ts:18
    upload manager.ts:199
Caused by: Error: StorageContext store failed: Failed to store piece on service provider - Failed to upload piece.

Details: Service Provider PDP
Failed to upload piece: Failed to finalize commP calculation
    createError errors.ts:13
    store context.ts:683
Caused by: UploadPieceError: Failed to upload piece.

Details: Service Provider PDP
Failed to upload piece: Failed to finalize commP calculation
    SynapseError base.ts:38
    UploadPieceError pdp.ts:95
    uploadPieceStreaming sp.ts:382
Caused by: undefined
```

This has been verified on multiple SPs. It's Curio responding with a 500, after most likely the SDK uploaded invalid data.

PR #681, opened 2026-03-20
fix(core): fall back to Blob fetch body when ReadableStream not supported

rvagg:
Firefox and Safari do not support `ReadableStream` as a fetch request body, causing uploads to send `"[object ReadableStream]"` (23 bytes) instead of actual data. Feature-detect support at runtime and drain the stream pipeline into a Blob before sending when necessary.

Fixes: https://github.com/FilOzone/synapse-sdk/issues/678

Notable (non-)regression here though is that we can't take advantage of parallel commp calculation in browsers that don't support `ReadableStream`, that'll end up blocking as we slurp the stream in here and then we follow up with the actual POST. This is a regression in _intended_ behaviour, but since it never worked in those browsers we're back to where we started with them. But there are a few options here (including the wasm calculator @hugomrdias made) that mainly involve farming off the work separately so it doesn't block and joining the result back when we need it. Mainly only matters for large inputs of course and the common cases (Chrome & Node.js) are fine with the status quo.

### DIFF
--- a/packages/synapse-core/src/sp/sp.ts
+++ b/packages/synapse-core/src/sp/sp.ts
@@ -17,7 +17,7 @@ import type { PieceCID } from '../piece/piece.ts'
 import * as Piece from '../piece/piece.ts'
 import type * as TypedData from '../typed-data/index.ts'
 import { RETRY_CONSTANTS, SIZE_CONSTANTS } from '../utils/constants.ts'
-import { isUint8Array } from '../utils/streams.ts'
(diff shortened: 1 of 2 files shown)


rvagg:
I haven't tested this, Firefox won't run properly for me for some reason and Safari doesn't have metamask support so I'm not sure how to run the playgroung test but I did run the feature-detect in a console and it was good (no throw, just didn't touch `duplex`).

PR #687, opened 2026-03-23
feat(core): preserve non-standard SP registry capabilities in PDPOffering

rvagg:
We shouldn't have been stripping theses, there's potentially useful signals in here downstream from our consumption of the capabilities providers publish. One example is https://github.com/FilOzone/dealbot/issues/364 where the "serviceStatus" capability still has some use, but there's plenty more where clients have needs that align with what service providers want to dump in their capabilities field that we don't know or care about.


hugomrdias (approved):
LGTM but copilot's first suggestion seems to be good

PR #694, opened 2026-03-24
fix(permissions): remove DeleteDataSet

### LINKED ISSUE #693: Fix misnamed `DeleteDataSet` session key permission typehash
The [verify signature contract bytecode](https://filecoin-testnet.blockscout.com/address/0x3fD7BFFAcAfdf35083d1424c1feC6481c85087F6?tab=contract_bytecode) does not contain the `DefaultFwssPermissions` item `DeleteDataSet` (`0xb5d6b3fc97881f05e96958136ac09d7e0bc7cbf17ea92fce7c431d88132d2b58`). The other permissions are present. I suspect that the hash is calculated incorrectly.

wjmelements:
Reviewer @hugomrdias
Closes #693
This permission was removed in https://github.com/FilOzone/filecoin-services/pull/255
Session keys can create but cannot delete data sets :(
#### Changes
* remove removed permission from SDK

### DIFF
--- a/docs/src/content/docs/developer-guides/session-keys.mdx
+++ b/docs/src/content/docs/developer-guides/session-keys.mdx
@@ -27,7 +27,6 @@ The SessionKeyRegistry stores arbitrary `bytes32` hashes as permissions and is a
 | `CreateDataSetPermission` | Create new datasets |
 | `AddPiecesPermission` | Add pieces to datasets |
 | `SchedulePieceRemovalsPermission` | Schedule piece removals |
-| `DeleteDataSetPermission` | Delete datasets |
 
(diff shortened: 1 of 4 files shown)


docs/src/content/docs/developer-guides/session-keys.mdx
```diff
@@ -27,7 +27,6 @@ The SessionKeyRegistry stores arbitrary `bytes32` hashes as permissions and is a
 | `CreateDataSetPermission` | Create new datasets |
 | `AddPiecesPermission` | Add pieces to datasets |
 | `SchedulePieceRemovalsPermission` | Schedule piece removals |
-| `DeleteDataSetPermission` | Delete datasets |
```
juliangruber: shall we try to fix and upgrade the contracts instead? This permission is useful (if it works)

Issue #691, opened 2026-03-24
Support paginated client dataset reads from FWSS


rjan90:
Support the paginated client dataset query change that landed in [filecoin-services#444](https://github.com/FilOzone/filecoin-services/pull/444). That contract change added paginated overloads for client dataset reads plus a length helper, so SDK dataset enumeration should be updated to use that shape instead of assuming the full client dataset list can always be returned in one call.

Context:
- [filecoin-services#444](https://github.com/FilOzone/filecoin-services/pull/444)
- [filecoin-pin#362](https://github.com/filecoin-project/filecoin-pin/issues/362)

This should cover the `synapse-core` / `synapse-sdk` client dataset query paths that currently rely on unpaginated reads.

PR #698, opened 2026-03-26
feat(synapse-core): support paginated client dataset reads from FWSS

Chaitu-Tatipamula:
# Summary
Adds synapse-core actions for the paginated client dataset query overloads introduced in [filecoin-services#444](https://github.com/FilOzone/filecoin-services/pull/444), addressing [filecoin-pin#362](https://github.com/filecoin-project/filecoin-pin/issues/362). This is in support of https://github.com/FilOzone/synapse-sdk/issues/691

## New actions
`getClientDataSetsLength` - returns total count of datasets for a client (`getClientDataSetsLength(address)`)
`getClientDataSetsIds` - returns paginated dataset IDs (`clientDataSets(address, offset, limit)`)

## Updated actions
`getClientDataSets` - added optional `offset/limit` params. defaults to unpaginated.

## Other changes

- ABI regenerated with updated GIT_REF
- Mock handler and presets updated for `getClientDataSetsLength`
- Tests added for all new/updated actions (call + mocked RPC)

all new pagination params default to 0n (fetch all), so no breaking changes are introduced.

> synapse-sdk changes will follow in a separate PR. but will just have exposing these actions in `WarmStorageService` but as per [comment](https://github.com/FilOzone/synapse-sdk/issues/691#issuecomment-4133378323) there should not be any new methods so i'll just add an updated call at `getClientDataSetsWithDetails()` to use paginated version of `clientDataSets `


rvagg (approved):
preemptively giving a 👍 but it's conditional on `getClientDataSetsIds` becoming `getClientDataSetIds`; this otherwise seems fine to me and it'll be good to get the sdk part done

hugomrdias (approved):

PR #701, opened 2026-03-30
feat(storage): add 'source' and 'withCDN' getters on StorageManager

rvagg:
To help with the case where you have a `Synapse` or a `StorageManeger` instance and want to do some metadata inspection before you hit combineMetadata. Currently with these two as private, we don't know whether one will show up or not.
e.g. in filecoin-pin we don't want to override an existing 'source' parameter with the 'filecoin-pin' default if the user supplied a `Synapse` instance with a 'source' but we have no way of knowing that currently.


juliangruber (commented):
+1

PR #713, opened 2026-04-01
Normalize core sp entrypoint

hugomrdias:
Mainly a maintenance PR to cleanup and organize the sp folder in synapse-core. Each operation has its own file with all the logic in it instead of being spread around in multiple files.

There's also a few minor change to export things properly.

One breaking change waitForPullStatus rename to waitForPullPieces to match the other `waitFor*` methods.


rvagg (approved):
lgtm .. big but looks ok I think

Issue #574, opened 2026-01-28
Normalize empty, not found, null outputs


hugomrdias:
We need to normalize contract output for empty, not found states, behavior varies intra-contract and between contracts.

- do it in sdk ? core ?
- in sdk return 
- throw NotFoundError  

ref:
- https://www.soliditylang.org/blog/2021/04/21/custom-errors/
- https://viem.sh/docs/contract/simulateContract#handling-custom-errors

> Also, require(condition, "error message") should be translated to if (!condition) revert CustomError().

/cc @rvagg

BigLep:
2026-04-09 conversation:
- On synapse side: normalize "empty" and "not found" as null (https://github.com/FilOzone/synapse-sdk/pull/723 )
- This issue primarily talks about errors since there are type and string errors.
- Encouraged distinction: we can null except where the caller is confused (e.g., terminated dataset but dataset is not live and user calls "delete piece")
- Another example: "getActivePieces" on a terminated so retruning to 0
- We also need to look at the way we're using calls internally
- Other case: if contract return type is an address then not found is "0 address", but we would want that as a null.
- There will be one PR per contract that synapse-core interacts with
- Prereq for version 1
- Need to document the decision in the development guide (agents.md). Maybe turn that into a skill.

PR #723, opened 2026-04-08
fix: normalize empty outputs for pdp-verifier

hugomrdias:
Handling these portions of https://github.com/FilOzone/synapse-sdk/issues/574
- add error handling for non-live data sets and implement LimitMustBeGreaterThanZeroError
- updates in the sdk
- better multicall errors mocking


juliangruber:
Removed review request as I need to focus on pay explorer for now

packages/synapse-core/src/utils/contract-errors.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Check if the error is a viem error
+ */
+export function isViemError(error: unknown): error is BaseError {
+  return (
+    error instanceof Error &&
+    'walk' in error &&
+    'details' in error &&
+    'shortMessage' in error &&
+    error.message.includes('Version: viem@')
```
rvagg, line 24: oh, this seems fragile, could you maybe ask AI to try and craft you a unit test for this that causes viem to throw an error and we get to assert this continues to work so that when viem iterates and changes this we get obvious breakage?

rvagg: looks like people are using `instanceof BaseError` https://github.com/wevm/viem/discussions/1789

also `BaseError#walk()` instead of what you're doing below in `stringErrorEquals`

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -1150,7 +1150,14 @@ export class StorageContext {
       pieceId = pieceData.id
 
       // Calculate timing based on nextChallengeEpoch
-      if (nextChallengeEpoch > 0n) {
+      if (nextChallengeEpoch === null) {
+        // If nextChallengeEpoch is 0, it might mean:
```
rvagg: `is 0`? should this be `is null`?

packages/synapse-core/src/pdp-verifier/get-next-challenge-epoch.ts
```diff
@@ -115,3 +123,18 @@ export function getNextChallengeEpochCall(options: getNextChallengeEpochCall.Opt
...
+ *
+ * @param data - The contract output from the getNextChallengeEpoch function {@link getNextChallengeEpoch.ContractOutputType}
+ * @returns The next challenge epoch for the data set {@link getNextChallengeEpoch.OutputType}
+ */
+export function parseNextChallengeEpoch(
+  data: getNextChallengeEpoch.ContractOutputType
+): getNextChallengeEpoch.OutputType {
+  if (data <= 0n) {
+    return null
+  }
```
rvagg, line 138: I htink this can be `=== 0n` but also we're now returning `null` for two cases so we're limiting the amount of information we can present here; maybe we should just pass through `0n` in this case because it does say something, we'd just need to document what that means (no challenge epoch set, as opposed to not live)

rvagg (approved):
sgtm; I'll leave the comments up to your judgement though I reckon you could do better on the error matching in here

packages/synapse-core/src/utils/contract-errors.ts
hugomrdias: i dont like using instanceof for external Error types, multiple versions of the same package in the bundle breaks it

hugomrdias: i will add the test but i think the .walk() wont help much its basically the same thing, go down to the cause and check for `reason`

packages/synapse-core/src/pdp-verifier/get-next-challenge-epoch.ts
hugomrdias: is that information we need to exposes in this action? 
my feeling is here we just say no challenge set in both cases, user can check if dataset is live with the specific action for that.

PR #727, opened 2026-04-10
fix: capture error cause in serialisation

rvagg:
We use cause in our custom error wraps


SgtPooki (approved):

hugomrdias:
Why are we serializing errors in synapse, shouldn't we let logging libraries like `pino` handle that ?

rvagg:
yeah, downstream you should use a better library, our toJSON are just convenience at this point (this isn't required for dealbot btw, just completeness since we have toJSON already)

PR #734, opened 2026-04-14
fix: fix out of gas problems with `getClientDataSets` 

### LINKED ISSUE #362: Too many data sets make many commands unusable
`0xa108be4331296ec8b8c47c2cd2fbfddf06e27523`

This wallet was used by me before for stress testing. It has a large number of data sets. I'm not sure exactly how much right now.
I obtained it through the contract method `clientDataSets` and now there are 4214

Now it seems that this method `getClientDataSets` cannot hold so many entries?

```
➜  /tmp filecoin-pin add out2287372121
┌  Filecoin Pin Add
│
◇  ✓ File validated (102.0 B)
│
◇  ✓ Connected to Filecoin - Calibration testnet
│
◇  ✓ Minimum payment setup verified (~0.066 USDFC required)
│
◇  ✓ File packed with root CID: bafybeihpws5pknsqfpfmqufxchoww4m5ry5jidy6s3zu22ac2v63tkediq
│
◇  ✓ IPFS content loaded (297.0 B)
│
◇  ✓ Payment capacity verified for 297.0 B
│
◇  ✗ Add failed: The contract function "getClientDataSets" reverted with the following reason:
message execution failed (exit=[33], revert reason=[none], vm error=[message failed with backtrace:
00: f0179532 (method 3844450837) -- contract reverted at 943 (33)
01: f0176641 (method 3844450837) -- out of gas (7)
02: f0176641 (method 6) -- out of gas (7)
 (RetCode=33)])

Contract Call:
  address:   0x53d235D474585EC102ccaB7e0cdcE951dD00f716
  function:  getClientDataSets(address client)
  args:                       (0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523)

Docs: https://viem.sh/docs/contract/readContract
Version: viem@2.47.4



➜  /tmp filecoin-pin data-set ls --all --rpc-url http://127.0.0.1:1234/rpc/v1
┌  Filecoin Onchain Cloud Data Sets
│
◇  ✗ Failed to list data sets
│
│
│  Error: Failed to get details for data set 11184: HTTP request failed.
│
│  URL: http://127.0.0.1:1234/rpc/v1
│  Request body: {"method":"eth_call","params":[{"data":"0xbdaac0560000000000000000000000000000000000000000000000000000000000002bb0","from":"0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523","to":"0x53d235D474585EC102ccaB7e0cdcE951dD00f716"},"latest"]}
│
│  Raw Call Arguments:
│    from:  0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│    to:    0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    data:  0xbdaac0560000000000000000000000000000000000000000000000000000000000002bb0
│
│  Contract Call:
│    address:   0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    function:  getDataSet(uint256 dataSetId)
│    args:                (11184)
│
│  Docs: https://viem.sh/docs/contract/readContract
│  Details: fetch failed
│  Version: viem@2.47.4
└  Listing failed


➜  /tmp filecoin-pin data-set show 11174
┌  Filecoin Onchain Cloud Data Set Details for #11174
│
◇  ✗ Failed to inspect data set
│
│
│  Error: Failed to get details for data set 2811: The request took too long to respond.
│
│  URL: wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v1
│  Request body: {"method":"eth_call","params":[{"data":"0xbdaac0560000000000000000000000000000000000000000000000000000000000000afb","from":"0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523","to":"0x53d235D474585EC102ccaB7e0cdcE951dD00f716"},"latest"]}
│
│  Raw Call Arguments:
│    from:  0xa108Be4331296Ec8b8C47c2Cd2FbfDDF06E27523
│    to:    0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    data:  0xbdaac0560000000000000000000000000000000000000000000000000000000000000afb
│
│  Contract Call:
│    address:   0x53d235D474585EC102ccaB7e0cdcE951dD00f716
│    function:  getDataSet(uint256 dataSetId)
│    args:                (2811)
│
│  Docs: https://viem.sh/docs/contract/readContract
│  Details: The request timed out.
│  Version: viem@2.47.4
└  Inspection failed
```

hugomrdias:
Should fix https://github.com/filecoin-project/filecoin-pin/issues/362

Adds p-queue to `getPdpDataSets` its should only run into rate limits error now, `getClientDataSetsWithDetails` in the sdk many need to use something like this.

@rvagg the `hasMore` pattern from pdp-verifier would have been better for these methods

### DIFF
--- a/packages/synapse-core/src/warm-storage/get-client-data-sets.ts
+++ b/packages/synapse-core/src/warm-storage/get-client-data-sets.ts
@@ -71,17 +71,111 @@ export async function getClientDataSets(
   client: Client<Transport, Chain>,
   options: getClientDataSets.OptionsType
 ): Promise<getClientDataSets.OutputType> {
-  const data = await readContract(
-    client,
-    getClientDataSetsCall({
-      chain: client.chain,
-      address: options.address,
-      offset: options.offset,
-      limit: options.limit,
-      contractAddress: options.contractAddress,
-    })
-  )
-  return data as getClientDataSets.OutputType
+  const limit = options.limit ?? 100n
+  let offset = options.offset ?? 0n
+  let needsMore = true
+  const dataSets: getClientDataSets.OutputType = []
+
+  while (needsMore) {
+    const data = await readContract(
+      client,
+      getClientDataSetsCall({
+        chain: client.chain,
+        address: options.address,
+        offset,
+        limit,
+        contractAddress: options.contractAddress,
+      })
+    )
+
+    for (const dataSet of data) {
+      if (dataSets.length < limit) {
+        dataSets.push(dataSet)
+      } else {
+        needsMore = false
+        break
+      }
+    }
+    if (data.length < limit) {
+      needsMore = false
+    }
+    offset += limit
+  }
+  return dataSets
+}
+
+export namespace getClientDataSetsIterable {
+  export type OptionsType = {
+    /** Client address to fetch data sets for. */
+    address: Address
+    /** Starting index (0-based). Use `0` to start from the beginning. Defaults to `0n`. */
+    offset?: bigint
+    /** Batch size for each pagination call. Defaults to `100n`. */
+    batchSize?: bigint
+    /** Warm storage contract address. If not provided, the default is the storage view contract address for the chain. */
+    contractAddress?: Address
+  }
+
+  export type OutputType = AsyncGenerator<DataSetInfo>
+
+  export type ErrorType = asChain.ErrorType | ReadContractErrorType
+}
+
+/**
+ * Get client data sets iterable
+ *
+ * @param client - The client to use to get data sets for a client address.
+ * @param options - {@link getClientDataSetsIterable.OptionsType}
+ * @returns Async generator of data set info entries {@link getClientDataSetsIterable.OutputType}
+ * @throws Errors {@link getClientDataSetsIterable.ErrorType}
+ *
+ * @example
+ * ```ts
+ * import { getClientDataSetsIterable } from '@filoz/synapse-core/warm-storage'
+ * import { createPublicClient, http } from 'viem'
+ * import { calibration } from '@filoz/synapse-core/chains'
+ *
+ * const client = createPublicClient({
+ *   chain: calibration,
+ *   transport: http(),
+ * })
+ *
+ * const dataSets = await getClientDataSetsIterable(client, {
+ *   address: '0x0000000000000000000000000000000000000000',
+ * })
+ *
+ * for await (const dataSet of dataSets) {
+ *   console.log(dataSet.dataSetId)
+ * }
+ * ```
+ */
+export async function* getClientDataSetsIterable(
+  client: Client<Transport, Chain>,
+  options: getClientDataSetsIterable.OptionsType
+): getClientDataSetsIterable.OutputType {
+  const batchSize = options.batchSize ?? 100n
+  let offset = options.offset ?? 0n
+  let hasMore = true
+
+  while (hasMore) {
+    const data = await readContract(
+      client,
+      getClientDataSetsCall({
+        chain: client.chain,
+        address: options.address,
+        offset,
+        limit: batchSize,
+        contractAddress: options.contractAddress,
+      })
+    )
+    for (const dataSet of data) {
+      yield dataSet
+    }
+    if (data.length < batchSize) {
+      hasMore = false
+    }
+    offset += batchSize
+  }
 }
 
 export namespace getClientDataSetsCall {
--- a/packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
+++ b/packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
@@ -1,6 +1,8 @@
+import PQueue from 'p-queue'
 import type { Chain, Client, ReadContractErrorType, Transport } from 'viem'
 import type { asChain } from '../chains.ts'
-import { getClientDataSets } from './get-client-data-sets.ts'
+
+import { type getClientDataSets, getClientDataSetsIterable } from './get-client-data-sets.ts'
 import { readPdpDataSetInfo } from './get-pdp-data-set.ts'
 import type { PdpDataSet } from './types.ts'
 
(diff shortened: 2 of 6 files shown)


packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -71,17 +71,111 @@ export async function getClientDataSets(
...
+        chain: client.chain,
+        address: options.address,
+        offset,
+        limit: batchSize,
+        contractAddress: options.contractAddress,
+      })
+    )
+    for (const dataSet of data) {
+      yield dataSet
+    }
```
wjmelements, line 173: before yielding, you could start the next readContract.

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
```diff
@@ -43,18 +45,41 @@ export async function getPdpDataSets(
   client: Client<Transport, Chain>,
   options: getPdpDataSets.OptionsType
 ): Promise<getPdpDataSets.OutputType> {
-  const data = await getClientDataSets(client, options)
+  const queue = new PQueue({
+    concurrency: 10,
+    interval: 1000,
+    intervalCap: 20,
```
wjmelements, line 51: 20 seems low to me. In geth I could do 3k-12k calls per second. Are we enforcing someone's rate limit?

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
```diff
@@ -43,18 +45,41 @@ export async function getPdpDataSets(
...
 
-    return {
-      ...dataSet,
-      ...pdDataSetInfo,
+  try {
+    for await (const dataSet of getClientDataSetsIterable(client, {
+      address: options.address,
+      offset: options.offset,
+    })) {
+      await queue.onSizeLessThan(20)
```
wjmelements, line 62: I don't think this line is necessary, and people might think it's enforcing the concurrency limit when it's actually limiting the size of the pending task queue

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
```diff
@@ -43,18 +45,41 @@ export async function getPdpDataSets(
...
+            ...pdDataSetInfo,
+          })
+        })
+        .catch(() => {
+          /* ignore */
+        })
     }
-  })
-  return Promise.all(promises)
+    await Promise.race([queue.onError(), queue.onIdle()])
```
wjmelements, line 79: I don't think onError is possible because all of the tasks in the queue catch and ignore.

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
```diff
@@ -43,18 +45,41 @@ export async function getPdpDataSets(
...
+            dataSetInfo: dataSet,
+            providerId: dataSet.providerId,
+          })
+          dataSets.push({
+            ...dataSet,
+            ...pdDataSetInfo,
+          })
+        })
+        .catch(() => {
+          /* ignore */
```
wjmelements, line 76: should we log this error? It could help us identify why a prior client dataset isn't being reused.

wjmelements (approved):

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
rvagg: yeah, not a bad suggestion although it does complicate the code a little when you're promise juggling and potentially means more calls than necessary if you're iterating for a search but it would be quicker collection if the consumer is doing more than just appending to an array
up to you Hugo

rvagg:
so you're pushing the complexity from the sdk into core? does this create inconsistencies in the patterns we've established so far where core is lower level?
(not an objection or blocking merge, just interested in the philosophical backing or whether we're making inconsistencies for ourselves; what's the rule here that we can embed going forward?)

hugomrdias:
> so you're pushing the complexity from the sdk into core? does this create inconsistencies in the patterns we've established so far where core is lower level? (not an objection or blocking merge, just interested in the philosophical backing or whether we're making inconsistencies for ourselves; what's the rule here that we can embed going forward?)

yeah but theres already other actions depending on this method in core that would need to implement the pagination and without the `hasMore` it felt too much to keep duplicating all round.

my rule here would be if its as low level as ran "out of gas" problems we should guard against in the core action.

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
hugomrdias: yeah glif public nodes, which i just discover have a 100req/min token bucket type rate limit...

but to be honest, its mostly an experiment as this action is not used in the sdk and we want to introduce this style of handling in other places

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
hugomrdias: This is kind of a quick from the p-queue package, theres actually 2 different promises to be handled the inner promise to do the work and the outer one to add to the queue, we need to handle both even though they would throw with the same error.  

This handles the inner one and below we ignore the outer one just so we dont get unhandled exceptions.

Depends on how you use the package, here we don't manage `queue.add` directly we just care about the inner promise that does the work.

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
hugomrdias: see previous comment

hugomrdias:
@rvagg whats do you think about the p-queue in `getPdpDataSets` handling for the resolve datasets from provider id in the sdk ?

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
wjmelements: I think this would only save time if the p-queue is full so it's probably not worth it

rvagg:
no problem with the p-queue in there, IIRC we were debating adding it in at one point but punted although it's already in core so it's already in the dep tree so may as well use it and it does make a whole class of complexity problems go away so use it where it fits!

PR #744, opened 2026-04-21
feat(synapse-core): normalize outputs and harden ServiceProviderRegistry actions

### LINKED ISSUE #574: Normalize empty, not found, null outputs

hugomrdias:
## Summary

Audit-driven hardening of the `ServiceProviderRegistry` SDK wrappers in `@filoz/synapse-core/sp-registry` (plus a small propagation into `@filoz/synapse-sdk`). Each commit addresses a distinct item from the audit plan.

### 1. Active PDP product handling (`1b2cbe83`)
- `getPDPProvider` / `getPDPProviderByAddress` now return `null` when the provider exists but has no active PDP product (`product.isActive === false` or empty `capabilityKeys`) instead of producing garbage via `decodePDPOffering`.
- `getPDPProvidersByIds` filters out the same inactive/empty entries from its multicall result.
- Adds a shared `hasActivePDPProduct` helper.

### 2. Client-side input validation for writes (`44f1f8ac`)
- New `validation.ts` with `ValidationError`-throwing helpers: `validateProviderInfo`, `validatePayee`, `validateProductType`, `validateLocation`, `validateCapabilities`, `validateRegistrationFee`.
- Wired into `registerProvider`, `addProduct`, `updateProduct`, `updateProviderInfo` so callers get clear errors before any RPC simulate/write.
- Contract limits mirrored in `SERVICE_PROVIDER_REGISTRY` on `utils/constants.ts` (`MAX_NAME_LENGTH`, `MAX_DESCRIPTION_LENGTH`, `MAX_LOCATION_LENGTH`, `MAX_CAPABILITY_KEY_LENGTH`, `MAX_CAPABILITY_VALUE_LENGTH`, `MAX_CAPABILITIES`, `REGISTRATION_FEE_WEI`).

### 3. `encodePDPCapabilities` subtle bugs (`9ae4e70d`)
- Exports `RESERVED_PDP_CAPABILITY_KEYS` (all 11 PDP keys that `encodePDPCapabilities` writes itself) and throws a `ValidationError` when a user-supplied extra collides with one.
- Empty string capability values now throw a `ValidationError` instead of being silently coerced to `0x01`; callers must pass explicit hex (e.g. `0x01`) for flags.

### 4. Empty/zero-state returns → `null` / `false` (`175e447b`)
- `getProviderIdByAddress`: `0n` → `null`.
- `isProviderActive`: `providerExists` reverts → `false`.
- `getProviderWithProduct`: `providerExists` reverts → `null`.
- `getPDPProvider`: propagates the new `null` from `getProviderWithProduct`.
- Adds `SP_REGISTRY_PROVIDER_DOES_NOT_EXIST` and `SP_REGISTRY_PROVIDER_NOT_FOUND` entries to `STRING_ERRORS` plus an `isProviderExistsRevert` helper.

### 5. synapse-sdk propagation (`9132ade6`)
- `SPRegistryService.getProviderIdByAddress` now returns `Promise<bigint | null>` to match the core action.

## Breaking changes (internal SDK surface)

- `getProviderIdByAddress` return type: `bigint` → `bigint | null`.
- `getPDPProvider` / `getPDPProviderByAddress` return type already was `PDPProvider | null` but the semantics widened to include `null` for unknown providers (previously threw).
- `getProviderWithProduct` return type: `ContractOutputType` → `ContractOutputType | null` (previously threw on unknown providers).
- `encodePDPCapabilities` now throws `ValidationError` for empty string values and for reserved PDP keys (previously coerced / silently overwrote).
- Write actions (`registerProvider`, `addProduct`, `updateProduct`, `updateProviderInfo`) now throw `ValidationError` for invalid inputs before hitting the RPC.
- `SPRegistryService.getProviderIdByAddress` return type: `Promise<bigint>` (with `0n`) → `Promise<bigint | null>`.

`ErrorType` unions were updated accordingly.


docs/src/content/docs/developer-guides/synapse-core.mdx
```diff
@@ -153,6 +153,9 @@ import { getPDPProvider } from "@filoz/synapse-core/sp-registry"
 
 // 1. Select a provider
 const provider = await getPDPProvider(publicClient, { providerId: 1n })
+if(!provider) {
```
rvagg:
```suggestion
if (!provider) {
```

docs/src/content/docs/developer-guides/synapse-core.mdx
```diff
@@ -225,6 +228,7 @@ import { getPDPProvider } from "@filoz/synapse-core/sp-registry"
 import { calibration } from "@filoz/synapse-core/chains"
 const publicClient = createPublicClient({ chain: calibration, transport: http() })
 const provider = await getPDPProvider(publicClient, { providerId: 1n })
+if(!provider) throw new Error('Provider not found.')
```
rvagg:
```suggestion
if (!provider) throw new Error('Provider not found.')
```

docs/src/content/docs/developer-guides/synapse-core.mdx
```diff
@@ -295,6 +299,8 @@ const publicClient = createPublicClient({ chain: calibration, transport: http()
 import { getPDPProvider } from "@filoz/synapse-core/sp-registry"
 const primaryProvider = await getPDPProvider(publicClient, { providerId: 1n })
 const secondaryProvider = await getPDPProvider(publicClient, { providerId: 2n })
+if(!secondaryProvider) throw Error('not found')
+if(!primaryProvider) throw Error('not found')
```
rvagg:
```suggestion
if (!secondaryProvider) throw Error('not found')
if (!primaryProvider) throw Error('not found')
```

don't we have linting in these?

packages/synapse-core/src/sp-registry/get-pdp-provider.ts
```diff
@@ -21,18 +22,41 @@ export namespace getPDPProvider {
...
+ * Returns `true` when the contract response carries an active, populated PDP product.
+ *
+ * The contract's `getProviderWithProduct` view only enforces `providerExists`,
+ * so it will happily return a default-initialized `ServiceProduct` with
+ * `isActive: false` and an empty `capabilityKeys` array for providers that
+ * never registered a PDP product or had it removed. Detect that case before
+ * attempting to parse capabilities.
+ */
+export function hasActivePDPProduct(data: getPDPProvider.ContractOutputType): boolean {
+  return data.product.isActive && data.product.capabilityKeys.length > 0
```
rvagg, line 45: do we have examples of this in the contract today? I think this should be an impossible state; doesn't hurt but we only allow a PDP product currently and we enforce a minimal set of capability keys

rvagg: oh .. uninitialised pdp product I guess?

packages/synapse-core/src/utils/constants.ts
```diff
@@ -136,3 +136,41 @@ export const RETRY_CONSTANTS = {
...
+  /** Maximum number of capability key/value pairs per product. */
+  MAX_CAPABILITIES: 24,
+
+  /**
+   * Registration fee in attoFIL (5 FIL) required by `registerProvider`.
+   *
+   * Callers that pass `value` explicitly are validated against this value.
+   * Callers that omit `value` fetch the live `REGISTRATION_FEE()` from the contract.
+   */
+  REGISTRATION_FEE_WEI: 5_000_000_000_000_000_000n,
```
rvagg: can we cahnge this from `WEI` please? we've avoided using wei anywhere in Synapse to date and it would be unfortunate to put it in; it's technically incorrect although I get the purpose -- "smallest unit of", but still, it's not Ether and it's in the contract as `REGISTRATION_FEE`, we could even just express it as `5e18` to mirror the contract.

packages/synapse-core/src/utils/pdp-capabilities.ts
```diff
@@ -40,6 +40,27 @@ export const CAP_IPNI_PEER_ID = 'ipniPeerId' // Optional (not validated by Bloom
...
 export const CAP_IPNI_PEER_ID_LEGACY = 'IPNIPeerID'
 
+/**
+ * Reserved PDP capability keys that `encodePDPCapabilities` writes itself.
+ *
+ * User-supplied `capabilities` whose key appears in this set are rejected to
+ * avoid desyncing the on-chain capability array (which keeps duplicates) from
+ * the capability-value mapping (last-write-wins).
+ */
+export const RESERVED_PDP_CAPABILITY_KEYS: ReadonlySet<string> = new Set([
```
rvagg: This is a duplication of `KNOWN_CAPABILITY_KEYS` below but it's hand-curated instead of derived (from the already hand-curated schema). You should delete this and export `KNOWN_CAPABILITY_KEYS` if you need it elsewhere (rename it too and lift it up to be with its friend constants too if you like).

rvagg (approved):
Conditional approval, the things I care most about being addressed in here before merge:
* Kill "wei"
* Duplicate `RESERVED_PDP_CAPABILITY_KEYS` / `KNOWN_CAPABILITY_KEYS`; don't add another hand-curated list

docs/src/content/docs/developer-guides/synapse-core.mdx
hugomrdias: no..

packages/synapse-core/src/sp-registry/get-pdp-provider.ts
hugomrdias: yes that

PR #750, opened 2026-04-22
fix: enhance pagination logic and add validation for batch size

hugomrdias:
- Fix anyone not setting a limit would have been capping at 100 data sets
- Updated `getClientDataSets` to handle a limit of 0n, returning all datasets.
- Improved pagination logic to correctly manage offsets and remaining datasets.
- Added validation to ensure `batchSize` in `getClientDataSetsIterable` is greater than 0n.
- Expanded tests to cover new pagination scenarios and validation checks.

### DIFF
--- a/packages/synapse-core/src/warm-storage/get-client-data-sets.ts
+++ b/packages/synapse-core/src/warm-storage/get-client-data-sets.ts
@@ -76,6 +76,9 @@ export async function getClientDataSets(
 ): Promise<getClientDataSets.OutputType> {
   const dataSets: getClientDataSets.OutputType = []
   const limit = options.limit ?? 0n
+  if (limit < 0n) {
+    throw new ValidationError('`limit` must be >= 0n.')
+  }
   let offset = options.offset ?? 0n
   let remaining = limit
 
@@ -184,7 +187,6 @@ export async function* getClientDataSetsIterable(
     }
     if (data.length < batchSize) {
       hasMore = false
-      continue
     }
     offset += batchSize
   }
--- a/packages/synapse-core/src/warm-storage/get-client-data-sets.ts
+++ b/packages/synapse-core/src/warm-storage/get-client-data-sets.ts
@@ -16,7 +16,7 @@ import type { ActionCallChain } from '../types.ts'
 import type { getPdpDataSets } from './get-pdp-data-sets.ts'
 import type { DataSetInfo } from './types.ts'
 
-const GET_CLIENT_DATA_SETS_PAGE_SIZE = 100n
+const DATA_SETS_PAGE_SIZE = 100n
 
 export namespace getClientDataSets {
   export type OptionsType = {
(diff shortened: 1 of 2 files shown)


packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -71,36 +74,57 @@ export async function getClientDataSets(
   client: Client<Transport, Chain>,
   options: getClientDataSets.OptionsType
 ): Promise<getClientDataSets.OutputType> {
-  const limit = options.limit ?? 100n
```
rvagg, line 74: oh ... you should call this out as a bug in your description, maybe even your commit msg - anyone not setting a limit would have been capping at `100` data sets!

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -71,36 +74,57 @@ export async function getClientDataSets(
...
+        getClientDataSetsCall({
+          chain: client.chain,
+          address: options.address,
+          offset: options.offset,
+          limit,
+          contractAddress: options.contractAddress,
+        })
+      )),
+    ]
+  }
```
rvagg: you can remove this, your loop already does this with your `pageLimit` setting and `pageLength < pageLimit` break so it's just as fast

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -153,7 +177,10 @@ export async function* getClientDataSetsIterable(
   client: Client<Transport, Chain>,
   options: getClientDataSetsIterable.OptionsType
 ): getClientDataSetsIterable.OutputType {
-  const batchSize = options.batchSize ?? 100n
+  const batchSize = options.batchSize ?? GET_CLIENT_DATA_SETS_PAGE_SIZE
+  if (batchSize <= 0n) {
```
rvagg, line 168: we should have the same check for ``limit` in `getClientDataSets`

hugomrdias: that removes the limit = 0n to get all

rvagg: sorry, `<` is what I meant, we don't allow <0 here but allow it above

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -71,36 +74,41 @@ export async function getClientDataSets(
   client: Client<Transport, Chain>,
   options: getClientDataSets.OptionsType
 ): Promise<getClientDataSets.OutputType> {
-  const limit = options.limit ?? 100n
-  let offset = options.offset ?? 0n
-  let needsMore = true
   const dataSets: getClientDataSets.OutputType = []
+  const limit = options.limit ?? 0n
```
rvagg, line 78:
```suggestion
  const limit = options.limit ?? 0n
  if (limit < 0n) {
    throw new ValidationError('`limit` must be >= 0n.')
  }
```

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -173,6 +184,7 @@ export async function* getClientDataSetsIterable(
     }
     if (data.length < batchSize) {
       hasMore = false
+      continue
```
rvagg: this ends up being a noop btw, you're just avoiding an unused increment but `!hasMore` breaks the loop anyway

rvagg (approved):
suggestions in line, nothing blocking but they should be considered (please)

rvagg:
lgtm

PR #754, opened 2026-04-24
feat(synapse-core): account runway and epoch conversion utilities

### LINKED ISSUE #163: show service billing info
Could we displays to users:

- The total amount of data they are currently storing
- How much time is left before they need to top up (i.e., when their service will expire)

hugomrdias:
## Summary

- Surface how long an account can sustain current spending by adding `runwayInEpochs` to `resolveAccountState` and threading it through `getAccountSummary`. Computed from the existing `fundedUntilEpoch`, with `maxUint256` when `lockupRate === 0n` (infinite runway) and `0n` when the account is already insolvent.
- Consolidate the repeated raw-account + current-epoch parameter shape used by the pure helpers into a shared `AccountState` type in `pay/types.ts`, reused by `resolveAccountState` and `calculateAccountDebt`.
- Add `epochsToHours` / `epochsToDays` bigint utilities (floor division) plus `TIME_CONSTANTS.EPOCHS_PER_HOUR` so callers can convert `runwayInEpochs` to human units. Both pass `maxUint256` through unchanged to preserve the infinite-runway sentinel.

## Test plan

- [x] `resolveAccountState` unit tests cover `runwayInEpochs` across healthy, insolvent, zero-rate, fixed-lockup, and truncation cases (migrated from the removed `calculate-account-runway.test.ts`).
- [x] `getAccountSummary` tests assert `runwayInEpochs` in healthy / draining / insolvent scenarios.
- [x] New `epochsToHours` / `epochsToDays` tests cover exact multiples, floor behavior, and `maxUint256` passthrough.
- [x] `biome check` clean on the touched files.

### DIFF
--- a/packages/synapse-core/src/pay/resolve-account-state.ts
+++ b/packages/synapse-core/src/pay/resolve-account-state.ts
@@ -1,28 +1,49 @@
 import { maxUint256 } from 'viem'
+import type { AccountState } from './types.ts'
 
 export namespace resolveAccountState {
-  export type ParamsType = {
-    funds: bigint
-    lockupCurrent: bigint
-    lockupRate: bigint
-    lockupLastSettledAt: bigint
-    currentEpoch: bigint
-  }
+  export type ParamsType = AccountState
 
   export type OutputType = {
+    /**
+     * Absolute epoch at which funds run out at the current lockup rate.
+     * `maxUint256` when `lockupRate` is 0n.
+     */
     fundedUntilEpoch: bigint
+    /** Funds available after accounting for all lockup (fixed + rate) at `currentEpoch`. */
     availableFunds: bigint
+    /**
+     * Number of epochs that can pass from `currentEpoch` before the account
+     * runs out of funds at the current lockup rate — i.e. how long until the
+     * user needs to deposit more funds.
+     *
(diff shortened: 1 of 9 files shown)


packages/synapse-core/src/pay/resolve-account-state.ts
```diff
@@ -1,28 +1,49 @@
...
   export type OutputType = {
+    /**
+     * Absolute epoch at which funds run out at the current lockup rate.
+     * `maxUint256` when `lockupRate` is 0n.
+     */
     fundedUntilEpoch: bigint
+    /** Funds available after accounting for all lockup (fixed + rate) at `currentEpoch`. */
     availableFunds: bigint
+    /**
+     * Number of epochs that can pass from `currentEpoch` before the account
```
rvagg: do we need a `currentEpoch` on here if we're going to be referencing it like this? or rephrase "from the current epoch"? there will be slippage, probably not critical enough to give the user an epoch reference but maybe it helps with assurance?

rvagg: I see this comes from the input, maybe just make it clear then? "can pass from `currentEpoch` on the input account state"

rvagg (approved):

PR #758, opened 2026-04-28
chore: update deployment commands to use pnpm exec for Wrangler and add log upload on failure

hugomrdias:
Also demo ci is failing need to check the logs to be able to debug

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
       - run: pnpm run build
       - run: pnpm -r --filter docs run build
       - name: Deploy docs
-        run: pnpm exec wrangler deploy --cwd docs --env production
+        run: pnpm --dir docs exec wrangler deploy --cwd docs --env production
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - run: pnpm run build
       - run: pnpm -r --filter docs run build
       - name: Deploy docs
-        run: pnpx wrangler deploy --cwd docs --env production
+        run: pnpm exec wrangler deploy --cwd docs --env production
```
rjan90: Wrangler is installed in the docs workspace, so run pnpm from there.

```suggestion
        run: pnpm --dir docs exec wrangler deploy --env production
```

hugomrdias: pnpm resolves it anyway no need to do this

rjan90: Hmmm..  I might be missing something here though, but it is failing for me. When I´m in the repo-root:

```
pwd                                         
/Users/phi/synapse-sdk
```

Then `pnpm exec wrangler deploy --cwd docs --env production` fails with:
 ```
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "wrangler" not found
 ```
 
 While `pnpm --dir docs exec wrangler deploy --env production` succeeds:

 ```
⛅️ wrangler 4.80.0 (update available 4.86.0)
─────────────────────────────────────────────

Cloudflare collects anonymous telemetry about your usage of Wrangler. Learn more at https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/telemetry.md
Attempting to login via OAuth...
-
```

rvagg: ditto from me, I tried and I have to agree with Orjan's suggestion here

hugomrdias: mine picks up the global one in my system.....

rjan90 (approved):

PR #759, opened 2026-04-28
refactor(synapse-core): update wagmi configuration and address validation

hugomrdias:
- Refactored the wagmi configuration to utilize new deployment schemas and improved address validation using Zod.
- Replaced deprecated address constants with dynamic fetching from a remote deployments JSON.
- Enhanced error handling for address validation and added new utility schemas for address types.
- Updated the mainnet and calibration address retrieval logic to align with the new deployment structure.

note: `PDP_VERIFIER_IMPLEMENTATION_ADDRESS` is not checksummed here https://github.com/FilOzone/filecoin-services/blob/v1.2.0/service_contracts/deployments.json

### DIFF
--- a/packages/synapse-core/wagmi.config.ts
+++ b/packages/synapse-core/wagmi.config.ts
@@ -1,175 +1,119 @@
 import { defineConfig } from '@wagmi/cli'
-import { fetch } from '@wagmi/cli/plugins'
-import { type Address, type Chain, type Client, createClient, http, type Transport } from 'viem'
-import { multicall } from 'viem/actions'
-import { calibration, mainnet } from './src/chains.ts'
+import { fetch as fetchPlugin } from '@wagmi/cli/plugins'
+import { request } from 'iso-web/http'
+import { zeroAddress } from 'viem'
+import * as z from 'zod'
+import { ZodValidationError } from './src/errors/base.ts'
+import { zAddress, zAddressLoose } from './src/utils/schemas.ts'
 
 // GIT_REF can be one of: '<branch name>', '<commit>' or 'tags/<tag>'
-const GIT_REF = 'ed85348ebad54196b5bfefc5cb0dbe7e8bfd6f7c' // v1.2.0
-const BASE_URL = `https://raw.githubusercontent.com/FilOzone/filecoin-services/${GIT_REF.replace(/^(?![a-f0-9]{40}$)/, 'refs/')}/service_contracts/abi`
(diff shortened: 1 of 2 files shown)


packages/synapse-core/wagmi.config.ts
```diff
@@ -1,175 +1,119 @@
...
+  if (result.error) {
+    throw result.error
   }
-}
 
-const mainnetClient = createClient({
-  chain: mainnet,
-  transport: http(),
-})
+  const parsed = DeploymentsSchema.safeParse(await result.result.json())
```
rvagg, line 49: how about we do `strict()` here to catch changes upstream and force us to react to them?

packages/synapse-core/wagmi.config.ts
```diff
@@ -1,175 +1,119 @@
...
 
 // GIT_REF can be one of: '<branch name>', '<commit>' or 'tags/<tag>'
-const GIT_REF = 'ed85348ebad54196b5bfefc5cb0dbe7e8bfd6f7c' // v1.2.0
-const BASE_URL = `https://raw.githubusercontent.com/FilOzone/filecoin-services/${GIT_REF.replace(/^(?![a-f0-9]{40}$)/, 'refs/')}/service_contracts/abi`
-const FWSS_ADDRESS_CALIBRATION = '0x02925630df557F957f70E112bA06e50965417CA0' as Address
-const FWSS_ADDRESS_MAINNET = '0x8408502033C418E1bbC97cE9ac48E5528F371A9f' as Address
-const ENDORSEMENTS_ADDRESS_CALIBRATION = '0xAA2f7CfC7ecAc616EC9C1f6d700fAd19087FAC84' as Address
-const ENDORSEMENTS_ADDRESS_MAINNET = '0x59eFa2e8324E1551d46010d7B0B140eE2F5c726b' as Address
-const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000' as Address
+const GIT_REF = 'd08214e1b3d200e0bc80f0d4f2e5ea3e1e4d603e' // v1.2.0
```
rvagg, line 10: could rename this to `FILECOIN_SERVICES_GIT_REF` now to be clear

rvagg (approved):
suggestions inline
also can you not `Closes` #567 cause there were two parts to that and this is only the first

rvagg:
EIP-55 casing being addressed in https://github.com/FilOzone/filecoin-services/pull/464; next time we should be able to remove the lose version (although note it'll technically be a breaking change when we remove it because it's exported - we could consider not exporting it and just defining it locally as a stopgap)

PR #760, opened 2026-04-29
feat(synapse-core)!: drop fundedUntilEpoch, add grossCoverageInEpochs

rvagg:
`resolveAccountState` and `getAccountSummary` now return `runwayInEpochs` and `grossCoverageInEpochs`; `fundedUntilEpoch` is no longer exposed. Buffer and deposit-needed helpers take `runwayInEpochs` directly. Account state docs reframed around deficit semantics.

Ref: https://github.com/filecoin-project/filecoin-pin/pull/412

---

I spent some time noodling this problem because in https://github.com/filecoin-project/filecoin-pin/pull/412 we're proposing "runway" as something different to what we mean by "runway" here; we'd have two divergent calculations. _But_ both are valid, just with a different framing, and 412 has a real user complaint behind it. So,  this proposal is to introduce both concepts here, and while we're doing it drop the absolute "Until" number because both of these runway variants could have an "Until" and we don't really need it here anyway.

* `runwayInEpochs` tells you how much you have at current rate _discounting_ your lockup
* `grossCoverageInEpochs` tells you how much you have at the current rate _including_ your lockup

The latter would solve for the complaint in 412 but what we should be doing is displaying both, unfortunately, we just need to express it clearly.

Lots of doc effort going on in here to try and express the concepts as clearly as possible, but it's not easy. Suggested here as:

> "your deposit covers ~X days of storage in total; you have ~Y days before you need to top up to keep paying"

### DIFF
--- a/docs/src/content/docs/developer-guides/payments/payment-operations.mdx
+++ b/docs/src/content/docs/developer-guides/payments/payment-operations.mdx
@@ -80,7 +80,7 @@ This requires two transactions and higher gas costs. Use `depositWithPermit` ins
 
 ### Account Summary
 
-Get a comprehensive snapshot of your account's payment state in a single call. This is the recommended way to check account health — it returns derived values like debt, lockup breakdown, and funded-until epoch.
+Get a comprehensive snapshot of your account's payment state in a single call. This is the recommended way to check account health: it returns derived values like debt, lockup breakdown, and runway.
 
 ```ts twoslash
 // @lib: esnext,dom
@@ -98,8 +98,16 @@ console.log("Rate per month:", formatUnits(summary.lockupRatePerMonth));
 console.log("Total lockup:", formatUnits(summary.totalLockup));
 console.log("  Fixed lockup:", formatUnits(summary.totalFixedLockup));
 console.log("  Rate-based lockup:", formatUnits(summary.totalRateBasedLockup));
-console.log("Funded until epoch:", summary.fundedUntilEpoch);
+console.log("Runway (epochs):", summary.runwayInEpochs);
+console.log("Gross coverage (epochs):", summary.grossCoverageInEpochs);
 ```
+
+`runwayInEpochs` and `grossCoverageInEpochs` answer different questions and generally diverge by roughly the size of the safety reserve:
+
+- **`runwayInEpochs`**: how long until your account enters deficit and the standard payment flow to providers halts. The operational "when must I act?" number. After this point the safety reserve is not automatically spent: it can only be claimed by a provider terminating their rail (which gives them up to one lockup period of guaranteed payment from the deficit point). Providers may keep serving for a while in deficit, but the relationship is at risk.
+- **`grossCoverageInEpochs`**: total spend your deposit covers at the current rate (`funds / lockupRate`). The explanatory "how much have I prepaid?" number. Treats `funds` as a single bucket: it uses the *whole* deposit, including the portion held in reserve, without modeling whether the reserve is actually flowing as ongoing payment.
+
+In a healthy account the gap between them is roughly the lockup period of spend; when in deficit `runwayInEpochs` clamps to `0n` while `grossCoverageInEpochs` still reflects the funds sitting in the account.
 
 ### Account Health Monitoring
 
@@ -111,16 +119,24 @@ import { Synapse } from "@filoz/synapse-sdk";
 import { privateKeyToAccount } from 'viem/accounts'
 const synapse = Synapse.create({ account: privateKeyToAccount('0x...'), source: 'my-app' });
 // ---cut---
-import { TIME_CONSTANTS } from "@filoz/synapse-sdk";
+import { epochsToDays } from "@filoz/synapse-core/utils";
 
 const summary = await synapse.payments.accountSummary();
-const epochsRemaining = summary.availableFunds / summary.lockupRatePerEpoch;
-const daysRemaining =
-  Number(epochsRemaining) / Number(TIME_CONSTANTS.EPOCHS_PER_DAY);
 
-console.log(`Days remaining: ${daysRemaining.toFixed(1)}`);
-if (daysRemaining < 7) console.warn("Low balance!");
+const runwayDays = epochsToDays(summary.runwayInEpochs);
+const totalDays = epochsToDays(summary.grossCoverageInEpochs);
+
+console.log(`Runway: ${runwayDays} days`);
+console.log(`Gross coverage: ${totalDays} days`);
+
+if (runwayDays === 0n && summary.funds > 0n) {
+  console.warn("Account in deficit: standard settlement to providers has halted. Deposit to resume.");
+} else if (runwayDays < 7n) {
+  console.warn("Low runway, top up soon to avoid service interruption.");
+}
 ```
+
+`epochsToDays` floors to whole days and passes `maxUint256` through unchanged (use that as the "infinite" sentinel when the account has no spend rate).
 
 ### Withdrawing Unlocked Funds
 
--- a/docs/src/content/docs/developer-guides/payments/payment-operations.mdx
+++ b/docs/src/content/docs/developer-guides/payments/payment-operations.mdx
@@ -102,12 +102,12 @@ console.log("Runway (epochs):", summary.runwayInEpochs);
 console.log("Gross coverage (epochs):", summary.grossCoverageInEpochs);
 ```
 
-`runwayInEpochs` and `grossCoverageInEpochs` answer different questions and generally diverge by roughly the size of the safety reserve:
+Two runway numbers, differing by the safety reserve:
 
-- **`runwayInEpochs`**: how long until your account enters deficit and the standard payment flow to providers halts. The operational "when must I act?" number. After this point the safety reserve is not automatically spent: it can only be claimed by a provider terminating their rail (which gives them up to one lockup period of guaranteed payment from the deficit point). Providers may keep serving for a while in deficit, but the relationship is at risk.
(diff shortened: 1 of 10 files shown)


rvagg:
Alternative names considered for the new variable: `fundsExhaustionInEpochs`, `burnDownEpochs`, `epochsAtCurrentRate`, `balanceRunwayEpochs`. All a bit crap in their own way. I initially stuck with the first of them but decided that `grossCoverageInEpochs` would be more accurate and give a better impression of what it might be without having to look at the docs.

packages/synapse-core/src/pay/get-account-summary.ts
```diff
@@ -19,38 +19,91 @@ export namespace getAccountSummary {
...
     lockupRatePerMonth: bigint
 
-    /** Total effective lockup at the given epoch (fixed + rate-based) */
+    /** Total effective lockup at `epoch` (fixed + rate-based). */
     totalLockup: bigint
-    /** Sum of lockupFixed across all rails (CDN deposits, etc.) */
+    /**
+     * Sum of `lockupFixed` across all rails. Reserved for one-time payments
+     * rather than streaming rate (e.g. FWSS CDN egress and cache miss
+     * credits).
```
Kubuxu: The `(e.g. FWSS CDN ....)` can read as being attached to the `streaming rate`. That is how I read it on first skim.

Kubuxu:
```suggestion
     * Sum of `lockupFixed` across all rails. Reserved for one-time payments,
     * for example, FWSS CDN egress and cache miss credits,
     * rather than streaming rate.
```

packages/synapse-core/src/pay/get-account-summary.ts
```diff
@@ -19,38 +19,91 @@ export namespace getAccountSummary {
...
+     *
+     * Always >= {@link runwayInEpochs}, typically by roughly the size of
+     * the reserve held in `totalLockup`. The two answer different
+     * questions: `runwayInEpochs` is the operational "act-by" number,
+     * accounting for the reserve as a floor that halts settlement;
+     * `grossCoverageInEpochs` treats `funds` as a single bucket without
+     * modeling whether the reserve is actually flowing as ongoing payment.
+     * They converge in degenerate cases (e.g. `totalLockup == 0` at
+     * `lockupLastSettledAt == epoch`, or trivially at zero `funds`), but
+     * generally diverge.
```
Kubuxu: This part comes across more as curiosity than something useful to the API's user.

packages/synapse-core/src/pay/resolve-account-state.ts
```diff
@@ -6,41 +6,101 @@ export namespace resolveAccountState {
...
+ * - `grossCoverageInEpochs`: `funds / lockupRate`, the total horizon the
+ *   deposit covers at the current rate. Use for "how much coverage has the
+ *   user prepaid in total?".
+ *
+ * `runwayInEpochs` and `grossCoverageInEpochs` answer different questions
+ * and generally diverge by roughly the reserve held in `lockupCurrent`,
+ * even for a healthy account. They only converge in degenerate cases
+ * (`lockupCurrent == 0` at `currentEpoch == lockupLastSettledAt`, or
+ * trivially at zero `funds`). Worked examples (in token units, with
+ * `lockupRate = 1 token / day`):
```
Kubuxu: This repeats what is stated in the `OutputType` documentation. It's up to you whether you'd like to keep it.

Kubuxu (commented):
I left some comments in the docs. It feels to me that there are quite a few repetitions in them.

rvagg:
> quite a few repetitions in them

There are because there's 3 main places: the docs themselves, and the jsdocs for two different APIs, so they'll all show up differently depending on how you approach this.

rvagg:
kk, economised a bit, dropped the curiosity bits (you're right) and dropped some of the details from the function-level jsdocs in deference to the type-level jsdocs

Kubuxu (approved):
SGTM but might be worth a second look from someone knowing Synpase better

packages/synapse-core/src/pay/get-account-summary.ts
```diff
@@ -19,38 +19,85 @@ export namespace getAccountSummary {
...
+     *
+     * Always >= {@link runwayInEpochs}, typically by roughly the size of
+     * the reserve held in `totalLockup`. `runwayInEpochs` accounts for the
+     * reserve as a floor that halts settlement; `grossCoverageInEpochs`
+     * treats `funds` as a single bucket without modeling whether the
+     * reserve is actually flowing as ongoing payment.
+     *
+     * Useful as a complement to `runwayInEpochs` in user-facing displays,
+     * e.g. "your deposit covers ~X days of storage in total; you have ~Y
+     * days before you need to top up to keep paying".
```
hugomrdias: i get the "your deposit covers ~X days of storage in total;" part but not the "you have ~Y days before you need to top up to keep paying".

If the SP terminates the rail because the user is in deficit it doesnt matter if the user top up at this point the rail is gone.

docs/src/content/docs/developer-guides/payments/payment-operations.mdx
```diff
@@ -98,9 +98,17 @@ console.log("Rate per month:", formatUnits(summary.lockupRatePerMonth));
...
-console.log("Funded until epoch:", summary.fundedUntilEpoch);
+console.log("Runway (epochs):", summary.runwayInEpochs);
+console.log("Gross coverage (epochs):", summary.grossCoverageInEpochs);
 ```
 
+`runwayInEpochs` and `grossCoverageInEpochs` answer different questions and generally diverge by roughly the size of the safety reserve:
+
+- **`runwayInEpochs`**: how long until your account enters deficit and the standard payment flow to providers halts. The operational "when must I act?" number. After this point the safety reserve is not automatically spent: it can only be claimed by a provider terminating their rail (which gives them up to one lockup period of guaranteed payment from the deficit point). Providers may keep serving for a while in deficit, but the relationship is at risk.
+- **`grossCoverageInEpochs`**: total spend your deposit covers at the current rate (`funds / lockupRate`). The explanatory "how much have I prepaid?" number. Treats `funds` as a single bucket: it uses the *whole* deposit, including the portion held in reserve, without modeling whether the reserve is actually flowing as ongoing payment.
```
hugomrdias: i think we should be more terse and explicit here and explain the "safety reserve" with that `runwayInEpochs` becomes runway without safety reserve and `grossCoverageInEpochs` becomes runway with safety reserve.

hugomrdias (approved):
LGTM just one issue about how to explain this to the user and usage in user facing interfaces.

> "your deposit covers ~X days of storage in total" 

Can be displayed with `runwayInEpochs` but with the catch that there is a safety buffer locked away for one extra "lookup period" (30 days for fwss)

Or it can be displayed with `grossCoverageInEpoch` for the absolute total that storage is paid for but with the catch that SP can terminates the rail because its already inside the "lookup period" and account is in deficit, at this point if user tops up the rail is gone.

> "you have ~Y days before you need to top up to keep paying"

Similar situation for displaying this one. 

So IMO we should always use `runwayInEpochs` for "you have ~Y days before you need to top up to keep paying"

rvagg:
Did another pass to nuance it even more. There's no "guarantee", no clear "safety", it's a one-way street once you enter it, etc. I think this properly expresses it now in a way the reader _may_ be able to grok enough of and the distinction between the numbers is a bit more stark now I hope.

docs/src/content/docs/developer-guides/payments/payment-operations.mdx
```diff
@@ -111,17 +119,25 @@ import { Synapse } from "@filoz/synapse-sdk";
...
+console.log(`Gross coverage: ${totalDays} days`);
+
+if (runwayDays === 0n && summary.funds > 0n) {
+  console.warn("Account in deficit: standard settlement to providers has halted. Deposit to resume.");
+} else if (runwayDays < 7n) {
+  console.warn("Low runway, top up soon to avoid service interruption.");
+}
 ```
 
+`epochsToDays` floors to whole days and passes `maxUint256` through unchanged (use that as the "infinite" sentinel when the account has no spend rate).
```
SgtPooki, line 139: > use that as the "infinite" sentinel when the account has no spend rate

what is the suggested way to check that the account has no spend rate? It seems like we're trying to say we check against maxUint256, but, check what against it?
which of these is right?:

* `epochsToDays(Infinity) === epochsToDays(summary.runwayInEpochs)`
* `epochsToDays(Infinity) === epochsToDays(summary.grossCoverageInEpochs)`
* `(1n << 256n) - 1n === summary.runwayInEpochs && (1n << 256n) - 1n === summary.grossCoverageInEpochs`
* other?

if this is clear elsewhere, can we link to it?

packages/synapse-core/test/account-debt.test.ts
```diff
@@ -80,18 +79,15 @@ describe('resolveAccountState', () => {
...
-
-    // simulatedSettledAt = min(maxUint256, 100) = 100
-    // simulatedLockupCurrent = 100 + 0 * (100 - 0) = 100
     // availableFunds = max(0, 1000 - 100) = 900
     assert.equal(result.availableFunds, 900n)
-
-    // zero rate → infinite runway
+    // zero rate: nothing draining, infinite for both numbers
     assert.equal(result.runwayInEpochs, maxUint256)
+    assert.equal(result.grossCoverageInEpochs, maxUint256)
```
SgtPooki, line 86: this answers my other question partially, but will a check of either imply both are maxUint256?

SgtPooki (approved):
makes sense to me. one minor nit of a question that might help me and others.

hugomrdias (approved):
LGTM, great changes!!! It feels way for clear now.

PR #772, opened 2026-05-05
chore: update lint scripts to remove --fix option in package.json files

hugomrdias:
This is currently blocking releases because its make the repo dirty before publishing to npm


wjmelements (approved):
I would like a separate `lint:fix` that runs biome check fix, but not blocking this PR

PR #785, opened 2026-05-12
ci: update filecoin-services ref on release

### LINKED ISSUE #567: Keep contract addresses in sync

rjan90:
## Why

Completes the synapse-sdk side of https://github.com/FilOzone/synapse-sdk/issues/567.

Companion PR in filecoin-services: https://github.com/FilOzone/filecoin-services/pull/475 to make filecoin-services repo dispatches release tag events to synapse-sdk. The workflow here in synapse-sdk then turn those events into SDK PRs that keep the pinned contract ref and generated bindings in sync.

## Summary

- add a `repository_dispatch` workflow for `filecoin-services-release`
- update `FILECOIN_SERVICES_GIT_REF` from the dispatched release SHA
- regenerate synapse-core ABIs and open/update an SDK PR automatically
- include a manual `workflow_dispatch` path for test runs
- include this workflow/helper script in CI path filters

### DIFF
--- a/.github/scripts/update-filecoin-services-ref.mjs
+++ b/.github/scripts/update-filecoin-services-ref.mjs
@@ -1,0 +1,38 @@
+import { readFile, writeFile } from 'node:fs/promises'
+import { resolve } from 'node:path'
+
+const usage = 'Usage: node .github/scripts/update-filecoin-services-ref.mjs <commit-sha> <tag>'
+const [shaInput, tag] = process.argv.slice(2)
+
+if (shaInput === undefined || shaInput === '' || tag === undefined || tag === '') {
+  throw new Error(usage)
+}
+
+const sha = shaInput.toLowerCase()
+
+if (/^[a-f0-9]{40}$/.test(sha) === false) {
+  throw new Error(`Expected a 40-character filecoin-services commit SHA, got '${shaInput}'`)
+}
+
+if (/^v[0-9][0-9A-Za-z._-]*$/.test(tag) === false) {
+  throw new Error(`Expected a filecoin-services release tag like 'v1.2.3', got '${tag}'`)
+}
+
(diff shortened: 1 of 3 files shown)


rjan90:
Closing, since this will all be done in the filecoin-services repo: https://github.com/FilOzone/filecoin-services/pull/475#issuecomment-4458910183

PR #793, opened 2026-05-18
docs(filbeam): add disclaimer about limited support

juliangruber:
For https://github.com/filbeam/roadmap/issues/105#issuecomment-4464181349

## Summary
- Adds a Disclaimer section to the FilBeam module JSDoc noting that FilBeam isn't currently fully supported and should be used with caution.
- Links users to the [filbeam/roadmap issue tracker](https://github.com/filbeam/roadmap/issues) so they can submit their use cases.

## Test plan
- [ ] JSDoc renders correctly in the generated docs
- [ ] No lint or build regressions (`pnpm run lint:fix`, `pnpm run build`)

### DIFF
--- a/packages/synapse-sdk/src/filbeam/index.ts
+++ b/packages/synapse-sdk/src/filbeam/index.ts
@@ -5,7 +5,7 @@
  *
  * ## Disclaimer
  *
- * FilBeam is not currently fully supported -- use with caution. If you have a
+ * FilBeam is not currently a priority for active support and development, use with caution. If you have a
  * use case for FilBeam, please open an issue at
  * {@link https://github.com/filbeam/roadmap/issues} describing it, so we can
  * prioritize the work accordingly.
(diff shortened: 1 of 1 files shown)


packages/synapse-sdk/src/filbeam/index.ts
```diff
@@ -3,6 +3,13 @@
  *
  * Client for the FilBeam stats API.
  *
+ * ## Disclaimer
+ *
+ * FilBeam is not currently fully supported -- use with caution. If you have a
```
rvagg:
```suggestion
 * FilBeam is not currently a priority for active support and development, use with caution. If you have a
```

just trying to find a way to make it a bit more helpful to users, what does "fully supported" actually mean? in what sense?

rvagg (approved):
approved pending something that expands on "fully supported"

PR #795, opened 2026-05-18
feat(piece)!: vendor PieceCID, redesign API

rvagg:
Triggered by https://github.com/FilOzone/synapse-sdk/pull/787, we have to coordinate to get a multiformats update but upstream is likely going to get less attention while synapse is going to be a source of truth for things like this.

So this does a targeted ingest of PieceCIDv2 functionality from `@web3-storage/data-segment` to `@filoz/synapse-core/piece`. It saves us a couple of unnecessary dependencies and we drop around half of the data-segment code (the code that deals with _data-segment_ functionality; PieceCIDv2 functionality just happens to be in a PoDSI library).

What this does:

- Vendor the upstream FR32, merkle, digest, hasher, and size primitives into packages/synapse-core/src/piece/internal/
- `PieceCID` class extends multiformats `CID` with eager `.root` / `.size` / `.height` / `.padding` / `.paddedSize` / `.toHex()` accessors
- New top-level API: `calculate`, `from`, `tryFrom`, `is`, `equals`, `hasher`, `transformStream`, plus sizing helpers (`paddedSizeFor`, `heightFor`)
- New public namespaces: `Piece.fr32.{expand,reduce}`, `Piece.merkle.{computeNode,zeroRoot,truncate}`
- Migrate all internal synapse-core / synapse-sdk call sites

Differences from `@web3-storage/data-segment`:
- `PieceCID` is a class extending `CID` with eager getters; upstream returns plain `CID` and callers manually parse digest fields
- `calculate()` accepts `Uint8Array | Blob | ReadableStream | AsyncIterable`; upstream's `Hasher` takes `Uint8Array` writes only
- Universal `from()` / `tryFrom()` accept `string` / `CID` / hex / bytes; upstream has separate parse paths
- Cross-realm and cross-bundle safe `Piece.is()` via `Symbol.for` brand on the prototype
- `fr32.{expand,reduce}` instead of `{pad,unpad}` (we've since all agreed to use different language for these operations to avoid confusion).

Also fixed a couple of bugs that turned up from adding more tests, they just were never exercised upstream.


hugomrdias (approved):
JUST WOW, this is great!

PR #796, opened 2026-05-19
fix(typehash)!: rename DeleteDataSet auth to TerminateService

rvagg:
Replace the DeleteDataSet EIP-712 type with TerminateService and update the default session-key permission set accordingly.

This matches the contract fixture generator: termination auth is consumed at terminateService, while deleteDataSet cleanup is no longer the user-auth action.

Also update the FRC-0069 PieceCID fixture to use the corrected padding/height encoding from filecoin-project/FIPs#1263.

BREAKING CHANGE: DeleteDataSetPermission is replaced by TerminateServicePermission. Existing session keys must be re-minted; passing a stale session key to Synapse.create() fails because the default permission set no longer matches.

_(sorry for the churn)_


SgtPooki (commented):
lgtm

hugomrdias (approved):

hugomrdias:
Probably needs a section in the migration guide https://github.com/FilOzone/synapse-sdk/blob/master/docs/src/content/docs/developer-guides/migration-guide.md

rvagg:
Added a 0.42.0 migration guide section with this.

PR #797, opened 2026-05-19
chore: properly cleanup archive

rvagg:
I accidentally comitted an _archive copy of the files I intended to remove and this went unnoticed in https://github.com/FilOzone/synapse-sdk/pull/777


rjan90 (approved):
Oooooops

PR #805, opened 2026-05-28
fix(session-key): name missing permissions in Synapse.create error

### LINKED ISSUE #695: Session key usage should not require every permission
Currently the synapse SDK crashes if your session key does not have 4 permissions (and one of them is fictional, see #693).

I should be able to authorize an app to add pieces but not to remove them. I should be able to authorize an app to add and remove pieces but not to create a new dataset. This is not currently possible with the SynapseSDK.

The error is
> Session key does not have the required permissions.

Furthermore, the error is not helpful or informative because it does not list the required permissions or whether they have expired.

The app will know which permissions they need. So, I think the correct time to fail is when attempting the unauthorized operation. If the app initialized the SDK with a session key that has no unexpired permissions, you can fail earlier.

juliangruber:
## Summary

Closes [#695](https://github.com/FilOzone/synapse-sdk/issues/695). Implements the plan from BigLep's [comment](https://github.com/FilOzone/synapse-sdk/issues/695#issuecomment-4215679419) on that issue: update the docs to capture the session-key philosophy, and replace the generic error with a more informative one.

### What changed

- **Informative error in `Synapse.create`** ([packages/synapse-sdk/src/synapse.ts:61](packages/synapse-sdk/src/synapse.ts)). The previous message was a single line that didn't tell the caller which permission was missing. The new message:
  - Names each missing permission by its EIP-712 type (`CreateDataSet`, `AddPieces`, `SchedulePieceRemovals`, `TerminateService`)
  - Distinguishes "not authorized" (`expiry === 0n`) from "expired at `<ISO timestamp>`"
  - Suggests calling `SessionKey.login` + `syncExpirations` to extend, or dropping to `@filoz/synapse-core` for a narrower scope
  - Links to https://docs.filecoin.cloud/developer-guides/session-keys/

- **New `PermissionNames` export** ([packages/synapse-core/src/session-key/permissions.ts](packages/synapse-core/src/session-key/permissions.ts)). A `Record<Hex, string>` mapping each FWSS permission hash to its human-readable EIP-712 type name, reusable in error messages and debug output.

- **Docs: philosophy section** ([docs/src/content/docs/developer-guides/session-keys.mdx](docs/src/content/docs/developer-guides/session-keys.mdx)). New "All-or-nothing in `@filoz/synapse-sdk`" subsection under "Use with the Synapse class" documenting that `Synapse.create` requires all four FWSS permissions by design, and pointing users who want a narrower scope at `@filoz/synapse-core`.

### Out of scope

[SgtPooki's later comment](https://github.com/FilOzone/synapse-sdk/issues/695#issuecomment-4421842036) proposes a `requiredPermissions?: Permission[]` option on `SynapseOptions` to let callers opt into a narrower required set without leaving `Synapse.create`. Not included here — happy to do that as a follow-up if you want to keep this PR focused on the agreed-upon error + docs change.

## Test plan

- [x] `pnpm run build` clean
- [x] `pnpm run lint` clean
- [x] All 254 node tests in `@filoz/synapse-sdk` pass, including two new tests in [packages/synapse-sdk/src/test/session-keys.test.ts](packages/synapse-sdk/src/test/session-keys.test.ts):
  - asserts the error names "AddPieces (not authorized)" and "SchedulePieceRemovals (expired at ...)" while omitting the two valid permissions, and includes the synapse-core suggestion + docs link
  - asserts `Synapse.create` succeeds when all four permissions are valid

### DIFF
--- a/docs/src/content/docs/developer-guides/session-keys.mdx
+++ b/docs/src/content/docs/developer-guides/session-keys.mdx
@@ -198,6 +198,14 @@ const synapse = Synapse.create({
 
 `Synapse.create()` validates that the session key has all four FWSS permissions (`DefaultFwssPermissions`) and that none are expired. This means the session key's expirations must be populated before construction, either by passing `expirations` to `fromSecp256k1()`, or by calling `sessionKey.syncExpirations()` after login.
 
+#### All-or-nothing in `@filoz/synapse-sdk`
+
+`@filoz/synapse-sdk` is the high-level "golden path" entry point and deliberately takes an all-or-nothing stance on session key permissions: `Synapse.create()` will throw if any of `CreateDataSet`, `AddPieces`, `SchedulePieceRemovals`, or `TerminateService` is missing or expired. This keeps the API surface predictable — every operation the high-level SDK exposes will work once construction succeeds.
+
+If you want a narrower scope (for example, only `CreateDataSet` + `AddPieces` for an upload-only client), drop down to [`@filoz/synapse-core`](/developer-guides/synapse-core/) directly. The core package's `SessionKey` and SP-client functions check permissions per operation, so you can authorize the minimum set you need and call those operations directly without going through `Synapse.create()`.
+
+The error thrown by `Synapse.create()` names the specific permissions that are missing or expired so you can decide whether to extend the session key's authorizations or switch to `@filoz/synapse-core` for that flow.
+
 ### Revoke the session key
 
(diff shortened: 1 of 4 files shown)


rvagg (approved):
that there's a verbose error message; but I suppose it's not bad in an AI world where agents read these things

juliangruber:
Yeah, it might be a bit too nice for an error string, but that's also a good problem to have. Let's revisit if there's additional feedback

PR #806, opened 2026-05-28
feat(synapse): add requiredPermissions option to SynapseOptions

juliangruber:
## Summary

Implements [SgtPooki's suggestion](https://github.com/FilOzone/synapse-sdk/issues/695#issuecomment-4421842036) on [#695](https://github.com/FilOzone/synapse-sdk/issues/695): add a `requiredPermissions?: Permission[]` option on `SynapseOptions`, defaulting to `SessionKey.DefaultFwssPermissions`, so least-privilege session keys can stay on the `Synapse.create` happy path.

### Why

Apps like [filecoin-pin-website](https://github.com/filecoin-project/filecoin-pin-website/pull/178) intentionally scope their session key to a subset of FWSS permissions (e.g. `CreateDataSet` + `AddPieces`). Today `Synapse.create` always checks the full `DefaultFwssPermissions` set, so those apps have to bypass `Synapse.create` and use the public `Synapse` constructor through `@filoz/synapse-core` directly, which means re-implementing the viem `createClient` + custom-transport setup that `Synapse.create` provides for free.

### What changed

- **[packages/synapse-sdk/src/types.ts](packages/synapse-sdk/src/types.ts)** — new `requiredPermissions?: Permission[]` field on `SynapseOptions`, with a docstring explaining that it gates construction (not per-operation), defaults to `SessionKey.DefaultFwssPermissions`, and that operations whose permissions are not listed will still revert on-chain if attempted.
- **[packages/synapse-sdk/src/synapse.ts:61](packages/synapse-sdk/src/synapse.ts:61)** — `Synapse.create` reads `options.requiredPermissions ?? SessionKey.DefaultFwssPermissions` instead of the hard-coded `DefaultFwssPermissions`. No other behavior change.
- **[docs/src/content/docs/developer-guides/session-keys.mdx](docs/src/content/docs/developer-guides/session-keys.mdx)** — new "Narrowing the required scope" subsection under "Use with the Synapse class" with example usage and a callout that the SDK still does not enforce per-operation checks.
- **[packages/synapse-sdk/src/test/session-keys.test.ts](packages/synapse-sdk/src/test/session-keys.test.ts)** — three new tests:
  - narrower required scope succeeds when only those permissions are authorized
  - still throws when one of the *required* permissions is missing even if other permissions are valid
  - defaults to `DefaultFwssPermissions` when `requiredPermissions` is omitted (regression guard)

### Backwards compatibility

`requiredPermissions` is optional and defaults to the previous behavior, so existing callers are unaffected.

### Per-operation checks

Deliberately not adding per-operation permission enforcement here — that was discussed previously as out of scope for GA, and SgtPooki's suggestion was specifically a construction-time knob. Callers must still only invoke SDK methods whose permissions they have authorized; otherwise the call will revert on-chain. The docs call this out explicitly.

### Relation to #805

[#805](https://github.com/FilOzone/synapse-sdk/pull/805) handles the other half of #695 (better error message + docs philosophy). This PR is independent: it only changes the validation input, not the message. Both PRs touch `Synapse.create` and will produce a small merge conflict whichever lands second; resolution is trivial (apply both: read `requiredPermissions` from options *and* itemize the missing ones in the error).

## Test plan

- [x] `pnpm run build` clean
- [x] `pnpm run lint` clean
- [x] All 255 node tests in `@filoz/synapse-sdk` pass, including the three new ones above


rvagg (approved):
sgtm but you're going to have to reconcile impl and docs between this and #805 before merging them

rjan90:
@hugomrdias will review this today

hugomrdias (approved):

PR #813, opened 2026-06-02
fix: improve retry and poll logic

### LINKED ISSUE #738: Retry sp http requests on network error
We should retry SP request on network error.

Currently some already do retries for other reasons and should also retry for network errors since the latest iso-web release, but we should verify the current retry logic doesnt override the default retry on network error.

Some other requests do not retry and can benefit from simple retry on network error.

related:
- https://github.com/FilOzone/synapse-sdk/pull/725

hugomrdias:
iso-web 3.0.0 added support for retry and poll logic and this PR updates all SP functions for it.

- updates to iso-web 3.0.0 with separate retry and poll logic
- sp functions can now retry on errors and also poll on 2XXs 
- new Curio 429 retry-after header is respected and properly handled
- most SP endpoints retry on error
- findOnProviders uses a HEAD request on the retrieval url
- response schema validation is now handled by iso-web
- retry and poll options exposed in all methods

```js
{
    /** The number of retries. Defaults to 2. */
    retryCount?: number
    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
    retryDelay?: number
    /** Whether to poll the request. Defaults to false. */
    poll?: boolean
    /** The poll interval in milliseconds. Defaults to 4 second. */
    pollInterval?: number
}
```

closes #738

### DIFF
--- a/packages/synapse-core/src/piece/download.ts
+++ b/packages/synapse-core/src/piece/download.ts
@@ -11,7 +11,7 @@ export namespace download {
     url: string
     /** The number of retries. Defaults to 2. */
     retryCount?: number
-    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to {@link RETRY_CONSTANTS.RETRY_DELAY}. */
     retryDelay?: number
     /** The signal to abort the request. */
     signal?: AbortSignal
@@ -51,7 +51,7 @@ export namespace downloadAndValidate {
     expectedPieceCid: string | PieceCID
     /** The number of retries. Defaults to 2. */
     retryCount?: number
-    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to {@link RETRY_CONSTANTS.RETRY_DELAY}. */
     retryDelay?: number
     /** The signal to abort the request. */
     signal?: AbortSignal
--- a/packages/synapse-core/src/piece/download.ts
+++ b/packages/synapse-core/src/piece/download.ts
@@ -1,6 +1,7 @@
-import { type AbortError, HttpError, type NetworkError, request, type TimeoutError } from 'iso-web/http'
+import { HttpError, type Errors as HttpErrors, request } from 'iso-web/http'
 import { DownloadPieceError } from '../errors/pdp.ts'
 import { InvalidPieceCIDError } from '../errors/piece.ts'
+import { RETRY_CONSTANTS } from '../utils/constants.ts'
 import { transformStream } from './calculate.ts'
 import { tryFrom } from './parse.ts'
 import type { PieceCID } from './piece-cid.ts'
@@ -8,9 +9,15 @@ import type { PieceCID } from './piece-cid.ts'
 export namespace download {
   export type OptionsType = {
     url: string
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
+    retryDelay?: number
+    /** The signal to abort the request. */
+    signal?: AbortSignal
   }
   export type ReturnType = Uint8Array
-  export type ErrorType = DownloadPieceError | TimeoutError | NetworkError | AbortError
+  export type ErrorType = DownloadPieceError | HttpErrors
 }
 
 /**
@@ -21,7 +28,14 @@ export namespace download {
  * @throws Errors {@link download.ErrorType}
  */
 export async function download(options: download.OptionsType): Promise<download.ReturnType> {
-  const response = await request.get(options.url)
+  const response = await request.get(options.url, {
+    timeout: false,
+    retry: {
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+    },
+    signal: options.signal,
+  })
   if (response.error) {
     if (HttpError.is(response.error)) {
       throw new DownloadPieceError(await response.error.response.text())
@@ -35,9 +49,15 @@ export namespace downloadAndValidate {
   export type OptionsType = {
     url: string
     expectedPieceCid: string | PieceCID
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
+    retryDelay?: number
+    /** The signal to abort the request. */
+    signal?: AbortSignal
   }
   export type ReturnType = Uint8Array
-  export type ErrorType = DownloadPieceError | TimeoutError | NetworkError | AbortError | InvalidPieceCIDError
+  export type ErrorType = DownloadPieceError | HttpErrors | InvalidPieceCIDError
 }
 
 /**
@@ -71,7 +91,15 @@ export async function downloadAndValidate(options: downloadAndValidate.OptionsTy
     throw new InvalidPieceCIDError(expectedPieceCid)
   }
 
-  const rsp = await request.get(url)
+  const rsp = await request.get(url, {
+    timeout: false,
+    retry: {
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+    },
+    signal: options.signal,
+  })
+
   if (rsp.error) {
     if (HttpError.is(rsp.error)) {
       throw new DownloadPieceError(await rsp.error.response.text())
--- a/packages/synapse-core/src/piece/download.ts
+++ b/packages/synapse-core/src/piece/download.ts
@@ -1,4 +1,4 @@
-import { HttpError, type Errors as HttpErrors, request } from 'iso-web/http'
+import { HttpError, type RequestErrors, request } from 'iso-web/http'
 import { DownloadPieceError } from '../errors/pdp.ts'
 import { InvalidPieceCIDError } from '../errors/piece.ts'
 import { RETRY_CONSTANTS } from '../utils/constants.ts'
@@ -17,7 +17,7 @@ export namespace download {
     signal?: AbortSignal
   }
   export type ReturnType = Uint8Array
-  export type ErrorType = DownloadPieceError | HttpErrors
+  export type ErrorType = DownloadPieceError | RequestErrors
 }
 
 /**
@@ -57,7 +57,7 @@ export namespace downloadAndValidate {
     signal?: AbortSignal
   }
   export type ReturnType = Uint8Array
-  export type ErrorType = DownloadPieceError | HttpErrors | InvalidPieceCIDError
+  export type ErrorType = DownloadPieceError | RequestErrors | InvalidPieceCIDError
 }
 
 /**
--- a/packages/synapse-core/src/piece/resolve-piece-url.ts
+++ b/packages/synapse-core/src/piece/resolve-piece-url.ts
@@ -3,7 +3,6 @@ import pLocate from 'p-locate'
 import pSome from 'p-some'
 import type { Address, Chain, Client, Transport } from 'viem'
 import { asChain } from '../chains.ts'
-import { findPiece } from '../sp/find-piece.ts'
 import type { PDPProvider } from '../sp-registry/types.ts'
 import { createPieceUrlPDP } from '../utils/piece-url.ts'
 import { getPdpDataSets } from '../warm-storage/get-pdp-data-sets.ts'
@@ -132,13 +131,17 @@ export async function chainResolver(options: resolvePieceUrl.ResolverFnOptionsTy
   }, new Map<bigint, (typeof dataSets)[number]['provider']>())
   const providers = [...providersById.values()]
 
-  const result = await findPieceOnProviders(providers, pieceCid, signal)
+  const result = await findPieceOnProviders(
+    providers.map((p) => p.pdp.serviceURL),
+    pieceCid,
+    signal
+  )
   if (result == null) {
     throw new Error('No provider found')
   }
   return createPieceUrlPDP({
     cid: pieceCid.toString(),
-    serviceURL: result.pdp.serviceURL,
+    serviceURL: result,
   })
 }
 
@@ -160,14 +163,18 @@ export async function chainResolver(options: resolvePieceUrl.ResolverFnOptionsTy
 export function providersResolver(providers: PDPProvider[]): resolvePieceUrl.ResolverFnType {
   return async (options: resolvePieceUrl.ResolverFnOptionsType) => {
     const { pieceCid, signal } = options
-    const result = await findPieceOnProviders(providers, pieceCid, signal)
+    const result = await findPieceOnProviders(
+      providers.map((p) => p.pdp.serviceURL),
+      pieceCid,
+      signal
+    )
     if (result == null) {
       throw new Error('No provider found')
     }
 
     return createPieceUrlPDP({
       cid: pieceCid.toString(),
-      serviceURL: result.pdp.serviceURL,
+      serviceURL: result,
     })
   }
 }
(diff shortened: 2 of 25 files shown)


packages/synapse-core/src/piece/resolve-piece-url.ts
```diff
@@ -160,43 +163,54 @@ export async function chainResolver(options: resolvePieceUrl.ResolverFnOptionsTy
...
-        serviceURL: p.pdp.serviceURL,
-        pieceCid,
-        signal: _signal,
-      }).then(
+    serviceURLs.map((p) =>
+      headPiece(p).then(
         () => p,
-        () => null
+        () => undefined
       )
```
rvagg:
```suggestion
      headPiece(p).catch(() => undefined)
```

`headPiece` already returns `p`

packages/synapse-core/src/sp/add-pieces.ts
```diff
@@ -184,7 +196,13 @@ export namespace waitForAddPieces {
...
     timeout?: number
-    /** The polling interval in milliseconds. Defaults to 4 seconds. */
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
+    retryDelay?: number
+    /** Whether to poll the request. Defaults to false. */
+    poll?: boolean
+    /** The poll interval in milliseconds. Defaults to 4 second. */
     pollInterval?: number
```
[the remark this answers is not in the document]
rvagg: genuine, all the waitFor's need resolution on this point - either accept and let an override, or don't accept at all

packages/synapse-core/src/sp/get-data-set.ts
```diff
@@ -27,9 +28,13 @@ export namespace getDataSet {
...
     /** The ID of the data set. */
     dataSetId: bigint
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
+    retryDelay?: number
   }
   export type OutputType = DataSet
-  export type ErrorType = GetDataSetError | TimeoutError | NetworkError | AbortError
+  export type ErrorType = GetDataSetError | HttpErrors
```
rvagg: looks like your new `schema` option is doing some work that makes its own kind of errors that may not be covered by `HttpErrors`? `SchemaError` and `JsonError` may be needed here, or augment `HttpErrors` upstream? 

the other places you use `schema:` would have the same problem

packages/synapse-core/src/sp/ping.ts
```diff
@@ -10,7 +10,10 @@ import { request } from 'iso-web/http'
  * @throws Errors {@link Error}
  */
 export async function ping(serviceURL: string) {
-  const response = await request.get(new URL(`pdp/ping`, serviceURL))
+  const response = await request.get(new URL(`pdp/ping`, serviceURL), {
+    retry: true,
+    timeout: 1000,
```
rvagg: I think `retry` isn't doing any work here, you have p-retry given a `minTimeout` of `1000` in iso-web, so we're going to hit our `timeout` before that; maybe you should/could unify the two in iso-web to catch things like this, if timeout<=1000, ignore `retry`

packages/synapse-core/src/sp/schedule-piece-deletion.ts
```diff
@@ -32,7 +36,11 @@ export async function deletePiece(options: deletePiece.OptionsType): Promise<del
     new URL(`pdp/data-sets/${dataSetId}/pieces/${pieceId}`, serviceURL),
     {
       body: { extraData },
-      timeout: RETRY_CONSTANTS.MAX_RETRY_TIME,
+      timeout: RETRY_CONSTANTS.TIMEOUT,
+      retry: {
+        retries: options.retryCount,
+        minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
```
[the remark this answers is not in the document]
rvagg: I think this is wrong and it will serialise won't it?

packages/synapse-core/src/sp/upload.ts
```diff
@@ -76,6 +83,10 @@ export async function uploadPiece(options: uploadPiece.OptionsType): Promise<voi
       'Content-Length': options.data.length.toString(),
     },
     timeout: false,
+    retry: {
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+    },
```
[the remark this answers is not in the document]
rvagg: I think it's on the list of defaults isn't it? so this is wrong

rvagg (approved):
approving assuming all comments/suggestions are resolved, Copilot has a good point about `poll` on the `waitFor` msgs, and there's some minor cleanup that's possible

rvagg:
Lifting up something I was writing inline at createDataSet:

Cases like this call are interesting given that it'll retry on a 5xx. If we get a createDataSet or addPieces error from some intermediary piece or perhaps a Curio db problem even though it might have resulted in a tx on the chain will retry, and then the user will get an error because of nonce reuse, so now we're in a crappy situation of having submitted somethign to the chain but the user thinks there's a strange error that's disconnected from the real error. Since iso-web also does the `isNetworkError` for retries too it's even worse - bad wifi but got it submitted? well when we do retry and submit it you're going to be told "can't reuse that nonce" or whatever our errors are for that.

I'm not sure what the answer is here tbh. Retrying on a 429 is an easy choice, but maybe the user should be handed errors in the case of 5xx's and network errors?

packages/synapse-core/src/piece/download.ts
```diff
@@ -1,16 +1,23 @@
...
 import { transformStream } from './calculate.ts'
 import { tryFrom } from './parse.ts'
 import type { PieceCID } from './piece-cid.ts'
 
 export namespace download {
   export type OptionsType = {
     url: string
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
```
juliangruber:
```suggestion
    /** The delay with exponential backoff between retries in milliseconds. Defaults to `RETRY_CONSTANTS.RETRY_DELAY`. */
```

this will otherwise easily get out of sync

packages/synapse-core/src/piece/resolve-piece-url.ts
```diff
@@ -160,43 +163,54 @@ export async function chainResolver(options: resolvePieceUrl.ResolverFnOptionsTy
...
  * Find the piece on the providers
  *
- * @param providers - {@link PDPProvider[]}
+ * @param serviceURLs - {@link string[]}
  * @param pieceCid - {@link PieceCID}
  * @param signal - {@link AbortSignal}
  * @returns The piece URL
  */
-export async function findPieceOnProviders(providers: PDPProvider[], pieceCid: PieceCID, signal?: AbortSignal) {
+export async function findPieceOnProviders(serviceURLs: string[], pieceCid: PieceCID, signal?: AbortSignal) {
```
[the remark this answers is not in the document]
juliangruber: +1

Would be nice to add a return type in TS, to get this with tooling

packages/synapse-core/src/sp/add-pieces.ts
```diff
@@ -101,6 +107,10 @@ export namespace addPieces {
     nonce?: bigint
     /** Pre-built signed extraData. When provided, skips internal EIP-712 signing. */
     extraData?: Hex
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
```
juliangruber:
```suggestion
    /** The delay with exponential backoff between retries in milliseconds. Defaults to `RETRY_CONSTANTS.RETRY_DELAY`. */
```

packages/synapse-core/src/sp/add-pieces.ts
```diff
@@ -184,7 +196,13 @@ export namespace waitForAddPieces {
     statusUrl: string
     /** The timeout in milliseconds. Defaults to 5 minutes. */
     timeout?: number
-    /** The polling interval in milliseconds. Defaults to 4 seconds. */
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
```
juliangruber:
```suggestion
    /** The delay with exponential backoff between retries in milliseconds. Defaults to `RETRY_CONSTANTS.RETRY_DELAY`. */
```

packages/synapse-core/src/sp/add-pieces.ts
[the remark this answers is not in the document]
juliangruber: +1

packages/synapse-core/src/sp/add-pieces.ts
```diff
@@ -184,7 +196,13 @@ export namespace waitForAddPieces {
...
+    retryDelay?: number
+    /** Whether to poll the request. Defaults to false. */
+    poll?: boolean
+    /** The poll interval in milliseconds. Defaults to 4 second. */
```
juliangruber: these will also easily get out of sync

packages/synapse-core/src/sp/add-pieces.ts
```diff
@@ -206,32 +224,31 @@ export namespace waitForAddPieces {
...
+    },
+    poll: {
+      limit: RETRY_CONSTANTS.POLL_LIMIT,
+      interval: options.pollInterval ?? RETRY_CONSTANTS.POLL_INTERVAL,
+      statusCodes: [202, 200], // 202 is processing, 200 is success
+      shouldPoll: async (ctx) => {
+        const data = (await ctx.response.clone().json()) as AddPiecesResponse
+        return data.piecesAdded === false
+      },
     },
```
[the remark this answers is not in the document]
juliangruber: +1

packages/synapse-core/src/sp/create-dataset-add-pieces.ts
```diff
@@ -26,14 +26,18 @@ export namespace createDataSetAndAddPiecesApiRequest {
     extraData: Hex
     /** The pieces to add. */
     pieces: PieceCID[]
+    /** The number of retries. Defaults to 2. */
+    retryCount?: number
+    /** The delay with exponential backoff between retries in milliseconds. Defaults to 250ms. */
```
juliangruber: as above..

juliangruber: there's more occurrences, might be able to use string replace

packages/synapse-core/src/sp/ping.ts
juliangruber: +1

packages/synapse-core/src/sp/schedule-piece-deletion.ts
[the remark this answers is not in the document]
juliangruber: yes, wrong

packages/synapse-core/src/sp/upload.ts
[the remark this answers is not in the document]
juliangruber: yes, wrong

juliangruber:
> Lifting up something I was writing inline at createDataSet:
> 
> Cases like this call are interesting given that it'll retry on a 5xx. If we get a createDataSet or addPieces error from some intermediary piece or perhaps a Curio db problem even though it might have resulted in a tx on the chain will retry, and then the user will get an error because of nonce reuse, so now we're in a crappy situation of having submitted somethign to the chain but the user thinks there's a strange error that's disconnected from the real error. Since iso-web also does the `isNetworkError` for retries too it's even worse - bad wifi but got it submitted? well when we do retry and submit it you're going to be told "can't reuse that nonce" or whatever our errors are for that.
> 
> I'm not sure what the answer is here tbh. Retrying on a 429 is an easy choice, but maybe the user should be handed errors in the case of 5xx's and network errors?

Agreed. The mutating POSTs (createDataSet, createDataSetAndAddPieces, addPieces, pullPieces) now set `retry: { methods: ['post'], ... }`. iso-web's default retry status codes are `[408, 413, 429, 500, 502, 503, 504]` and it also retries all network errors. So if Curio submits an on-chain tx and then returns a 500 (or the connection drops), the SDK silently re-POSTs → second tx submission → nonce-reuse error surfaced to the user, disconnected from the real cause. For these non-idempotent, tx-producing endpoints I'd restrict retry to `statusCodes: [429]` (keeping afterStatusCodes for Retry-After) and not retry on 5xx/network. Worth resolving before merge.

juliangruber (changes requested):

packages/synapse-core/src/sp/schedule-piece-deletion.ts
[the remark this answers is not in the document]
hugomrdias: yes its wrong, its the request.json variant.

hugomrdias:
@rvagg @juliangruber agree with your comments but i would keep the retry on `NetworkError` this error means the network failed the request never got to the other side so its safe to retry for this.

juliangruber:
> @rvagg @juliangruber agree with your comments but i would keep the retry on `NetworkError` this error means the network failed the request never got to the other side so its safe to retry for this.

A `NetworkError` does not guarantee the request never arrived. iso-web's `NetworkError` comes from `is-network-error`, whose triggers include `'terminated'`, `'Network connection lost'`, `'fetch failed'`, and `'Failed to fetch'` — all of which can fire after the request was sent and the server acted on it, when only the response was lost.

packages/synapse-core/src/sp/create-dataset.ts
```diff
@@ -60,14 +63,18 @@ export async function createDataSetApiRequest(
...
-      'Content-Type': 'application/json',
     },
-    timeout: RETRY_CONSTANTS.MAX_RETRY_TIME,
+    timeout: RETRY_CONSTANTS.TIMEOUT,
+    retry: {
+      methods: ['post'],
+      statusCodes: [429],
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+      shouldRetry: (ctx) => !NetworkError.is(ctx.error),
```
juliangruber: Still retries on 5xx (this predicate overrides the `statusCodes: [429]` gate). Should work:
```ts
shouldRetry: (ctx) => HttpError.is(ctx.error) && ctx.error.code === 429
```

packages/synapse-core/src/sp/create-dataset-add-pieces.ts
```diff
@@ -58,18 +62,22 @@ export async function createDataSetAndAddPiecesApiRequest(
...
-      'Content-Type': 'application/json',
     },
-    timeout: RETRY_CONSTANTS.MAX_RETRY_TIME,
+    timeout: RETRY_CONSTANTS.TIMEOUT,
+    retry: {
+      methods: ['post'],
+      statusCodes: [429],
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+      shouldRetry: (ctx) => !NetworkError.is(ctx.error),
```
juliangruber: here as well

packages/synapse-core/src/sp/add-pieces.ts
```diff
@@ -52,17 +56,21 @@ export async function addPiecesApiRequest(
...
-    }),
-    timeout: RETRY_CONSTANTS.MAX_RETRY_TIME,
+    },
+    timeout: RETRY_CONSTANTS.TIMEOUT,
+    retry: {
+      methods: ['post'],
+      statusCodes: [429],
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+      shouldRetry: (ctx) => !NetworkError.is(ctx.error),
```
juliangruber: here as well

packages/synapse-core/src/sp/schedule-piece-deletion.ts
```diff
@@ -32,7 +36,14 @@ export async function deletePiece(options: deletePiece.OptionsType): Promise<del
...
     {
       body: { extraData },
-      timeout: RETRY_CONSTANTS.MAX_RETRY_TIME,
+      timeout: RETRY_CONSTANTS.TIMEOUT,
+      retry: {
+        methods: ['delete'],
+        statusCodes: [429],
+        retries: options.retryCount,
+        minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+        shouldRetry: (ctx) => !NetworkError.is(ctx.error),
```
juliangruber: here as well

juliangruber (dismissed):
There's an issue with 429 handling: iso-web's `shouldRetry` replaces the built-in decision rather than filtering it further, so the `statusCodes` option is ignored

packages/synapse-core/src/sp/create-dataset.ts
juliangruber: This also suggests that tests are missing

juliangruber: Once these are in place, we can probably remove `statusCodes: [429]`, as that's not used

rvagg (approved):
approving again, nice find by @juliangruber and it looks like its fixed precisely as prescribed.

only possible remaining change is that methods: ['post']/['delete']` are now redundant where you have a `shouldRetry` based on what Julian pointed out, but not a blocking concern, just possibly causing confusion

pnpm-workspace.yaml
```diff
@@ -44,6 +44,7 @@ minimumReleaseAgeExclude:
   # Remove after 2026-06-08 when both age out.
   - viem@2.52.0
   - ox@0.14.27
+  - astro
```
rvagg: can you version this pls, and add a comment

rvagg: I'm doing it in commit 1287666 in https://github.com/FilOzone/synapse-sdk/pull/821

PR #822, opened 2026-06-03
fix: remove left-over withIPNI metadata

hugomrdias:


rvagg (approved):
lint doesn't like you tho

Issue #763, opened 2026-04-29
Support new per-operation fee structure


timfong888:
<details>
<summary>Original issue description, replaced by @rvagg comment below</summary>

## Context

From the [Pricing Proposal review](https://docs.google.com/spreadsheets/d/17tnkZgdPYB4D0YurbcmFBeUsCnHyWbvanWQmjH4lBL0/edit) (Apr 28–29, 2026):

> **Hugo:** "all of this will change the current Costs api"
>
> **Rod:** "In all of this there will be SDK work; which may end up being the hardest part of implementation (floor-price-while-zero-data contract problem is the second most complicated bit). Costs API was hard enough with the simple pricing, this will take a lot of care and thought and we might need to add some fuzz to our language and the things we return and give access to pricing schedules (which we should surface from the contracts)."

The pricing model is moving from a single floor + creation fee to a layered structure that the SDK / Costs API needs to expose cleanly — including the new per-operation gas fee paid by the user.

## Scope

- Surface the on-contract **pricing schedule** so SDK consumers can read upcoming pricing (depends on [filecoin-services#466](https://github.com/FilOzone/filecoin-services/issues/466))
- Differentiate fee components in the Costs API response:
  - Sybil fee (client-paid, currently 0.1 USDFC)
  - SP gas-cover fee on `createDataSet` (~$0.00112) — the per-operation fee originally scoped here
  - Recurring per-data-set floor ($0.06/mo today)
  - Per-piece add / delete fees
- Handle **floor-price-while-zero-data** semantics (drawing from escrow against an empty data set)
- **Token-decimal aware floor** — 6-decimal ERC20s like axlUSDC truncate the current $0.06/mo floor to 0 (separate pricing decision ticket coming)
- Add fuzz/flexibility to the API language and returned shapes so future pricing tweaks don't require breaking changes

## Related

- [filecoin-services#466](https://github.com/FilOzone/filecoin-services/issues/466) — contract-side pricing schedule (prerequisite)
- [filecoin-services#458](https://github.com/FilOzone/filecoin-services/issues/458) — empty data set / floor-from-escrow
- https://github.com/FilOzone/filecoin-services/issues/469

## Owner

Rod (assigned) — Hugo on SDK side
</details>

rvagg:
Building on the description and in https://github.com/FilOzone/filecoin-services/issues/469 (note the details in there, particularly the precision note) and assuming that shape continues to deployment, this is the design proposal for Synapse APIs:

[...]

PR #823, opened 2026-06-03
feat(costs)!: per-operation pricing price list

SgtPooki:
## What changed

Reshapes the Costs API to the GA per-operation pricing model. `getPriceList()` reads the full `token` / `rates` / `fees` / `lockups` catalogue from `FilecoinWarmStorageServiceStateView.getPriceList()` ([filecoin-services#501](https://github.com/FilOzone/filecoin-services/pull/501)); the effective-rate, lockup, deposit, and upload-cost paths, `StorageManager`, and the React `usePriceList` hook all read from it. Operator-approval `maxLockupPeriod` now reads `getPriceList().lockups.defaultLockupPeriod` ([filecoin-services#503](https://github.com/FilOzone/filecoin-services/issues/503)).

## How to verify

```sh
pnpm -F @filoz/synapse-core build && pnpm -F @filoz/synapse-sdk build
pnpm -F @filoz/synapse-core test:node
pnpm -F @filoz/synapse-sdk test:node
```

## Notes / risks

- **Do not publish until calibration and mainnet redeploy with [filecoin-services#501](https://github.com/FilOzone/filecoin-services/pull/501).** `getPriceList()` reverts on the current live contracts and there is no fallback (mocked tests pass).
- The ABI is wired via a contained `getPriceList` fragment (`src/abis/price-list.ts`) instead of bumping the generated ref; in-code TODOs flag it for removal once `FILECOIN_SERVICES_GIT_REF` includes `getPriceList`.
- Relates to #763.

BREAKING CHANGE: removes `getServicePrice()` (core + SDK) and `useServicePrice()` in favor of `getPriceList()` / `usePriceList()`; drops the `rate` alias on upload costs (use `rates`), the `LOCKUP_PERIOD` export, and the `addPiecesOperationCount` cost option.


rvagg:
@SgtPooki you should be able to change the generated.ts to point to a commit hash and get new ABI pulled in. I think this is fine, we're nearing the last of  the changes pre-GA and it wouldn't be terrible to merge in a specific commit that's not on a tag, we can always follow-up later (and we'll get the auto PR now that there's a workflow in place).

packages/synapse-core/src/utils/constants.ts
rvagg:
```suggestion
```

especially helpful to defer to the chain considering https://github.com/FilOzone/filecoin-services/issues/503

packages/synapse-core/src/warm-storage/calculate-upload-fees.ts
```diff
@@ -0,0 +1,38 @@
...
+    total: bigint
+  }
+}
+
+/**
+ * Compute the one-time fees an upload incurs.
+ *
+ * Scope is intentionally limited to upload-time fees: create-data-set (new
+ * datasets only) and add-pieces. Schedule-removals, terminate, and delete are
+ * post-upload lifecycle operations and are not part of an upload cost preview.
```
rvagg, line 24: so maybe this file and this function is poorly named, `upload-operation-fees` perhaps?

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -23,6 +24,10 @@ export namespace getUploadCosts {
 
     /** Size of new data to upload, in bytes. */
     dataSize: bigint
+    /** Number of pieces added by this operation. Default: 1 */
+    pieceCount?: bigint
+    /** Number of addPieces operations. Default: 1 */
+    addPiecesOperationCount?: bigint
```
rvagg: this is an interesting DX hurdle - how to express this well? why would I want multiple operations? should there be a relationship between this API and  the known max batching limit? we should at least mention that limit as a reason that it would span beyond the max, maybe we should just calculate the number of operations for them if they don't provide it and `pieceCount` is more than the max (40)?

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -32,12 +37,27 @@ export namespace getUploadCosts {
...
+      /** Rate per month — full precision for display. */
+      perMonth: bigint
+    }
+    /** @deprecated Use rates. */
     rate: {
       /** Rate per epoch — matches on-chain PDP rail rate. */
       perEpoch: bigint
       /** Rate per month — full precision for display. */
       perMonth: bigint
     }
```
rvagg: just kill it I reckon, we're going to v1 this so lets not `@deprecated` anything in here, maybe we should even do a pass and clean it all out before v1

packages/synapse-core/src/warm-storage/index.ts
```diff
@@ -13,6 +13,7 @@ export * from './add-approved-provider.ts'
 export * from './calculate-additional-lockup-required.ts'
 export * from './calculate-deposit-needed.ts'
 export * from './calculate-effective-rate.ts'
+export * from './calculate-operation-fees.ts'
```
rvagg: I can see why the name now it's in context with the otherss, but it's still not the right name, the others are ok because they are obviously about specific things, but "operation fees" could be about deletions or terminations too

packages/synapse-core/src/warm-storage/price-list.ts
```diff
@@ -0,0 +1,88 @@
+import type { Address, Chain, Client, Transport } from 'viem'
+import { getServicePrice } from './get-service-price.ts'
+
+export namespace getPriceList {
+  export type OptionsType = getServicePrice.OptionsType
+
+  /**
+   * The canonical warm storage price list. Matches the on-chain `PriceList`
+   * struct from `FilecoinWarmStorageServiceStateView.getPriceList()`
+   * ([filecoin-services#501](https://github.com/FilOzone/filecoin-services/issues/501)).
```
rvagg:
```suggestion
   * struct from `FilecoinWarmStorageServiceStateView.getPriceList()`.
```

packages/synapse-core/src/warm-storage/price-list.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Storage and egress rates and the token come from the contract via
+ * {@link getServicePrice}. The proving (dataset) rate, operation fees, and
+ * lockup amounts come from {@link expectedFees} / {@link expectedLockups}.
+ */
+export async function getPriceList(
+  client: Client<Transport, Chain>,
+  options: getPriceList.OptionsType = {}
+): Promise<getPriceList.OutputType> {
+  const servicePrice = await getServicePrice(client, options)
```
rvagg:
```suggestion
  // TODO: replace with actual call and actual data and remove expected* above
  const servicePrice = await getServicePrice(client, options)
```

you don't need to commit this, I just really don't want to lose track of this

packages/synapse-react/src/warm-storage/use-service-price.ts
```diff
@@ -17,6 +17,9 @@ export interface UseServicePriceProps {
 /**
  * Get the service price for the warm storage.
  *
+ * @deprecated Use {@link usePriceList} for the latest pricing. This returns only
```
rvagg: yeet

packages/synapse-core/src/warm-storage/calculate-effective-rate.ts
```diff
@@ -4,23 +4,23 @@ export namespace calculateEffectiveRate {
...
     sizeInBytes: bigint
-    /** Price per TiB per month from getServicePrice(). */
-    pricePerTiBPerMonth: bigint
-    /** Minimum monthly charge from getServicePrice(). */
-    minimumPricePerMonth: bigint
-    /** Epochs per month from getServicePrice() (always 86400). */
+    /** Storage price per TiB per month. */
+    storagePerTibPerMonth: bigint
+    /** Monthly proving service rate for non-empty datasets. */
+    provingServicePerMonth: bigint
```
rvagg: probably should just align with contract language on this one, it mutated into `datasetMonthlyFee`, we can explain it in docs as a "proving service fee"

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -357,8 +358,20 @@ export class WarmStorageService {
...
+   * @returns Recurring rates, operation fees, and lockups.
+   */
+  async getPriceList(): Promise<getPriceList.OutputType> {
+    return getPriceList(this._client)
+  }
+
+  /**
+   * Get the current service price from the current FWSS ABI.
+   *
+   * @deprecated Use {@link WarmStorageService.getPriceList} for the latest
```
rvagg: remove

packages/synapse-core/src/warm-storage/get-upload-costs.ts
SgtPooki: ok yea, we can derive this automatically and remove this from surface area for consumers..

packages/synapse-core/src/warm-storage/index.ts
SgtPooki: changed to calculate-upload-fees.ts & calculateUploadFees

packages/synapse-core/src/warm-storage/calculate-effective-rate.ts
SgtPooki: updated, but the name merged was `datasetFeePerMonth` so i used that

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -64,13 +75,14 @@ export async function getUploadCosts(
...
   const extraRunwayEpochs = options.extraRunwayEpochs ?? DEFAULT_RUNWAY_EPOCHS
   const bufferEpochs = options.bufferEpochs ?? DEFAULT_BUFFER_EPOCHS
 
   // Fetch all needed data in parallel
-  const [accountInfo, pricing, approved, currentEpoch] = await Promise.all([
+  const [accountInfo, priceList, approved, currentEpoch] = await Promise.all([
     accounts(client, { address: options.clientAddress }),
-    getServicePrice(client),
+    getPriceList(client),
     isFwssMaxApproved(client, { clientAddress: options.clientAddress }),
```
rvagg: `isFwssMaxApproved` has a new call to `getPriceList` itself when you don't provide a lockup option; a bit of a tangle to remove the extra call without adding a sequential step. The obvious thing to do would be to pull `isFwssMaxApproved` out of this batch, then send it the lockup explicitly so it doesn't need to fetch that and then do it separately. Maybe this is OK since that call already has a sequential set of calls and we'd be removing one of them so possibly we don't lose anything by stacking them up here cause we're stacking calls in `isFwssMaxApproved` anyway

packages/synapse-core/src/pay/set-operator-approval.ts
```diff
@@ -222,10 +231,19 @@ export function setOperatorApprovalCall(
     }
   }
 
+  // maxLockupPeriod comes from the chain price list
+  // (getPriceList().lockups.defaultLockupPeriod), which this synchronous builder
+  // cannot read, so approving callers must resolve and pass it.
+  if (options.approve && options.maxLockupPeriod === undefined) {
```
rvagg, line 248: Could mutate the type that comes in to `setOperatorApprovalCall` to signal this flip behaviour, when `approve` is true then require `maxLockupPeriod`. Not critical, it would muddy the call argument (&'s and |'s) just might make it nicer for TS consumers.

packages/synapse-core/src/warm-storage/price-list.ts
rvagg, line 1: @hugomrdias @SgtPooki I don't love that we don't cache the price list between calls, it's going to rarely change and it'll be almost impossible for it to change within a single JS runtime instance, although there's a small chance for a very long running Node process. We've comitted to making synapse-core stateless, but could we consider a call shape here that's cached? Ideally the default, so maybe a separate call so you can always go direct (we could rename `getPriceList` to `getPriceListUncached` or something obvious) and store a local state cached version of it in this module, keyed by chainid (could throw in view contract address as well just to be super safe), perhaps also with a TTL of an hour.

rvagg (approved):
Nothing blocking in here, some suggestions made, I'd love caching but it's not critical, we could discuss as a separate change. The ABI import is fine, we can move to a tag for v1. There's just the one duplicate `getPriceList()` in a batch that it would be nice to minimise if we can do it cleanly.

packages/synapse-core/src/abis/price-list.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * `getPriceList()` view fragment for `FilecoinWarmStorageServiceStateView`,
+ * mirroring the on-chain `PriceList` struct from
+ * [FilOzone/filecoin-services#501](https://github.com/FilOzone/filecoin-services/pull/501).
+ *
+ * TODO: remove this file and the `fwssView` merge in `abis/index.ts` once
+ * `FILECOIN_SERVICES_GIT_REF` (`wagmi.config.ts`) points at a release whose
```
rvagg: you should just do this btw, for main, I'll put a PR up to make it happen but you can bundle it here. Update to latest `main`, we'll need to do it once more for https://github.com/FilOzone/filecoin-services/pull/506 which adds an event, but that's fine

rvagg:
@SgtPooki I pushed a commit here that pulls in the latest fwss `main` ABI and fixes some fallout from that. It's not going to be the last abi update, we still have https://github.com/FilOzone/filecoin-services/pull/506 to go at least, but it's good enough and we'll just not release from `master` until we have an FWSS release.

packages/synapse-core/src/warm-storage/get-upload-costs.ts
SgtPooki: Done. Pulled `isFwssMaxApproved` out of the `Promise.all` batch and passed it `requiredMaxLockupPeriod: priceList.lockups.defaultLockupPeriod` from the list we already fetched, so it doesn't re-read `getPriceList`. Costs one sequential step after the batch resolves, which is fine since that call already stacks its own reads. Same change in `getUploadCosts` and `StorageManager.calculateMultiContextCosts`.

packages/synapse-core/src/pay/set-operator-approval.ts
SgtPooki: Did it. `setOperatorApprovalCall.OptionsType` is a discriminated union now: `approve: true` requires `maxLockupPeriod`, `approve: false` makes it optional. Direct builder callers get a compile error instead of the runtime throw.

I kept the runtime guard as the JS-consumer safety net. Left the custom-operator allowance check runtime-only, since encoding that one in the type would muddy it more than it's worth.

packages/synapse-core/src/warm-storage/price-list.ts
SgtPooki: Agreed it rarely changes, caching makes sense. My hesitation: it'd be the first stateful thing in synapse-core, against the stateless commitment, so i'd rather not land it without @hugomrdias signoff.

React consumers already get this via `usePriceList` (react-query, keyed on `chain.id`). Direct core callers don't.

Options:
1. memoize on the `WarmStorageService` instance, caching at the stateful layer
2. your module-level default + `getPriceListUncached`, most convenient but the biggest departure

I think we should track in a follow-up issue, but open to suggestions.. @hugomrdias what do you think?

SgtPooki:
FYI this is fine to merge, but just has the [do not merge] tag because we're waiting on @hugomrdias to align the v1 releases (so we don't block the release pipeline with contract upgrade required changes)

BigLep:
Per 2026-06-09 FOC wg sync: @hugomrdias is going to take this over and merge it.

packages/synapse-core/src/warm-storage/price-list.ts
hugomrdias: https://github.com/FilOzone/synapse-sdk/issues/837

PR #827, opened 2026-06-04
feat!: provider-relayed service termination as the default path

### LINKED ISSUE #792: Synapse support for simplified DataSet terminations and cleanup for GA
https://www.notion.so/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7f#360dc41950c180ab9ee4f7ff8cd96e12

rvagg:
- synapse-core/sp: terminateService + waitForTerminateService against the SP's terminate endpoints (sign 712, POST, poll; typed conflict/failure errors)
- sdk: StorageManager.terminateService and StorageContext.terminate share one flow; relay is immediate with settlement pre-check, onChain: true submits directly and winds down over the lockup period
- renames terminateDataSet to terminateService; result carries endEpoch

Closes: https://github.com/FilOzone/synapse-sdk/issues/792


docs/src/content/docs/developer-guides/storage/storage-operations.mdx
```diff
@@ -289,23 +289,54 @@ console.log("PDP URL:", providerInfo.pdp.serviceURL);
...
+Only terminate when you're certain you no longer need the data.
 :::
 
-To delete an entire data set and discontinue payments for the service, call `context.terminate()`.
-This method submits an on-chain transaction to initiate the termination process. Following a defined termination period, payments will cease, and the service provider will be able to delete the data set.
-
-You can also terminate a data set using `synapse.storage.terminateDataSet({ dataSetId })`, when the data set ID is known and creating a context is not necessary.
+To end the storage service for a data set and discontinue its payments, call
+`context.terminate()`, or `synapse.storage.terminateService({ dataSetId })`
+when the data set ID is known and a context is not necessary.
```
SgtPooki:
```suggestion
when the data set ID is known (a context is not necessary).
```

packages/synapse-sdk/src/storage/terminate.ts
```diff
@@ -0,0 +1,105 @@
...
+    lockupRate: accountInfo.lockupRate,
+    lockupLastSettledAt: accountInfo.lockupLastSettledAt,
+    currentEpoch,
+  })
+  if (debt > 0n) {
+    throw createError(
+      'StorageManager',
+      'terminateService',
+      `Account cannot settle its lockup in full (shortfall: ${debt} of the payment token's base units); deposit funds, or terminate on-chain (onChain: true) to wind down over the lockup period instead`
+    )
```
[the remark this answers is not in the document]
SgtPooki: +1 here.. doesn't seem like a big deal but maybe a more generic name would be better?

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -974,14 +977,44 @@ export class StorageManager {
...
+
+  /**
+   * Resolve the PDP endpoint of the provider holding a data set.
+   * Validates existence, liveness and ownership along the way.
    */
-  async terminateDataSet(options: { dataSetId: bigint }): Promise<Hash> {
-    return this._warmStorageService.terminateDataSet(options)
+  private async _resolveServiceURL(dataSetId: bigint): Promise<string> {
+    const spRegistry = new SPRegistryService({ client: this._synapse.client })
+    const { provider } = await StorageContext.resolveByDataSetId(
```
SgtPooki, line 1016: resolveByDataSetId checks dataSet is live, if its not, it doesn't look like the `	DataSetAlreadyTerminatedError` error will be thrown if this fails. maybe add a test for this?

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -367,12 +367,14 @@ export class WarmStorageService {
   // ========== Data Set Operations ==========
 
   /**
-   * Terminate a data set with given ID
-   * @param options - Options for the data set termination
+   * Terminate a data set service by submitting the transaction on-chain directly.
```
SgtPooki:
```suggestion
   * Terminate the storage service by submitting the transaction on-chain directly.
```

SgtPooki (dismissed):
in general looks good, but a few things that should probably be nailed down

packages/synapse-core/src/sp/terminate-service.ts
```diff
@@ -0,0 +1,355 @@
...
+
+export type TerminateServiceStatusPending = z.infer<typeof TerminateServiceStatusPendingSchema>
+export type TerminateServiceStatusRejected = z.infer<typeof TerminateServiceStatusRejectedSchema>
+export type TerminateServiceStatusSuccess = z.infer<typeof TerminateServiceStatusSuccessSchema>
+export type TerminateServiceStatusResponse =
+  | TerminateServiceStatusPending
+  | TerminateServiceStatusRejected
+  | TerminateServiceStatusSuccess
+
+const schema = z.union([TerminateServiceStatusRejectedSchema, TerminateServiceStatusSuccessSchema])
```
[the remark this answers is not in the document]
rvagg: no, this is the _final_ schema, we don't want pending mixed up in there, `shouldPoll` does something different

added a comment

packages/synapse-sdk/src/storage/terminate.ts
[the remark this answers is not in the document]
rvagg: fixed by supplying the source down into the call so it can be accurate

packages/synapse-sdk/src/storage/manager.ts
rvagg: ok, cool, I've pulled up `getServiceURL` earlier so it'll fail on liveness check before it gets here; added test

BigLep:
Requesting @juliangruber so we can hopefully get this merged before Rod signs off for travel.

rvagg:
It seems that this will need reconciliation with https://github.com/FilOzone/synapse-sdk/pull/823 because of the updated ABIs, there's a minor change that impacts a test case here, easy fix.

docs/src/content/docs/developer-guides/storage/storage-operations.mdx
```diff
@@ -289,23 +289,54 @@ console.log("PDP URL:", providerInfo.pdp.serviceURL);
...
-To delete an entire data set and discontinue payments for the service, call `context.terminate()`.
-This method submits an on-chain transaction to initiate the termination process. Following a defined termination period, payments will cease, and the service provider will be able to delete the data set.
-
-You can also terminate a data set using `synapse.storage.terminateDataSet({ dataSetId })`, when the data set ID is known and creating a context is not necessary.
+To end the storage service for a data set and discontinue its payments, call
+`context.terminate()`, or `synapse.storage.terminateService({ dataSetId })`
+when the data set ID is known (a context is not necessary).
+
+There are two ways to terminate, chosen with the `onChain` option. The
+provider-mediated path is the recommended default: you sign an authorization,
```
juliangruber, line 310:
```suggestion
provider-mediated path (`onChain: false`) is the recommended default: you sign an authorization,
```

juliangruber: It's a bit confusing to mention that it's `onChain: false` but then say directly afterwards that the provider relays it on chain. The option is slightly confusing, as this is always on chain, the change is who submits it there. In the spirit of yesterday's Slack suggestion I don't consider this a merge blocker, but would recommend to find a better name. Suggestion: `skipProvider`.

docs/src/content/docs/developer-guides/storage/storage-operations.mdx
```diff
@@ -289,23 +289,54 @@ console.log("PDP URL:", providerInfo.pdp.serviceURL);
...
+the provider relays it on chain, and if your account can settle in full the
+service can terminate immediately so provider cleanup can begin right away.
+
+The direct on-chain path is an independent fallback. It is useful if the
+provider-mediated request cannot be completed, for example because the provider
+endpoint is temporarily unavailable, relay support is not active yet, or you
+need to submit the transaction yourself. It does not require provider
+cooperation, but it uses the standard lockup wind-down: the service continues
+until `endEpoch`, typically about 30 days, while the provider settles and winds
+down the data set before cleanup.
```
juliangruber, line 319: It's not very clear from the docs why the default is preferred and the other is a fallback. It's obvious that the fallback is required when the SP doesn't cooperate. It's less clear that the fallback will lead to more spending. Docs mention that with the default the `service can terminate immediately` but that doesn't clearly state that payments end sooner. Can we mention more prominently the most important user impact, how these options affect one's funds?

juliangruber: The table also doesn't clearly resolve this.

docs/src/content/docs/developer-guides/migration-guide.md
```diff
@@ -11,6 +11,26 @@ If you are coming from an earlier version of any of the Synapse packages, you wi
 
 ## 0.42.0
 
+### Action: Replace `terminateDataSet` with `terminateService`
+
+Data set termination is now service termination. `terminateDataSet` was removed.
```
juliangruber, line 16: It's not obvious to me why the new method is called `terminateService` and not `terminateDataSet`. Is that rename because there's a semantic change and thus the method is less confusing when having a new name?

packages/synapse-core/src/errors/pdp.ts
```diff
@@ -178,6 +179,113 @@ export class DeletePieceError extends SynapseError {
...
+      details: decodedError,
+    })
+  }
+
+  static override is(value: unknown): value is TerminateServiceError {
+    return isSynapseError(value) && value.name === 'TerminateServiceError'
+  }
+}
+
+export class DataSetAlreadyTerminatedError extends SynapseError {
```
juliangruber: confusing: Why is there a `TerminateServiceError` but not `ServiceAlreadyTerminatedError`?

packages/synapse-core/src/errors/pdp.ts
```diff
@@ -178,6 +179,113 @@ export class DeletePieceError extends SynapseError {
...
+  static override is(value: unknown): value is WaitForTerminateServiceError {
+    return isSynapseError(value) && value.name === 'WaitForTerminateServiceError'
+  }
+}
+
+export class WaitForTerminateServiceNotFoundError extends SynapseError {
+  override name: 'WaitForTerminateServiceNotFoundError' = 'WaitForTerminateServiceNotFoundError'
+
+  constructor() {
+    super(`No client-requested termination found for this data set.`, {
```
juliangruber: why not "..for this service."?

packages/synapse-core/src/sp/terminate-service.ts
```diff
@@ -0,0 +1,357 @@
...
+  WaitForTerminateServiceRejectedError,
+} from '../errors/pdp.ts'
+import { signTerminateService } from '../typed-data/sign-terminate-service.ts'
+import { RETRY_CONSTANTS } from '../utils/constants.ts'
+import { zHex, zNumberToBigInt } from '../utils/schemas.ts'
+
+/*
+SP-side termination protocol, as observed through the HTTP API.
+
+POST /pdp/data-sets/{id}/terminate
```
juliangruber, line 30: This is also in favor of calling it `terminateDataSet`

packages/synapse-sdk/src/storage/terminate.ts
```diff
@@ -0,0 +1,110 @@
...
+    lockupCurrent: accountInfo.lockupCurrent,
+    lockupRate: accountInfo.lockupRate,
+    lockupLastSettledAt: accountInfo.lockupLastSettledAt,
+    currentEpoch,
+  })
+  if (debt > 0n) {
+    throw createError(
+      source,
+      'terminateService',
+      `Account cannot settle its lockup in full (shortfall: ${debt} of the payment token's base units); deposit funds, or terminate on-chain (onChain: true) to wind down over the lockup period instead`
```
juliangruber: This case isn't mentioned in docs, that you'd want to go on-chain because of lacking funds

packages/synapse-core/src/sp/terminate-service.ts
```diff
@@ -0,0 +1,357 @@
...
+            response for SP-initiated terminations (only client-requested ones
+            are visible) and, eventually, for fully cleaned-up data sets.
+
+  Reverted and 404 are two observations of the same outcome (the SP discards a
+  failed relay shortly after it lands): retry, or terminate on-chain. Success
+  may carry an empty hash (the SP found the service already terminated and
+  sent no tx), and fwssTerminated: true wins over txSuccess: false (a
+  competing terminate landed first; the goal state holds). When no terminate
+  tx ever lands, ours or anyone's (e.g. the SP is unable to send), there is no
+  terminal signal: the status stays queued and the poller runs to its timeout.
```
juliangruber, line 56: how can the caller recover from the timeout?

packages/synapse-sdk/src/storage/terminate.ts
```diff
@@ -0,0 +1,110 @@
...
+  const serviceURL = await getServiceURL()
+
+  // Immediate termination settles the payer's account in full. Best-effort
+  // pre-check: catches a clear shortfall before signing, but lockup keeps
+  // accruing until the provider's tx lands, so a marginal account can still
+  // revert SP-side (surfacing as the rejected/404 path).
+  const payerAddress = synapse.client.account.address
+  const [accountInfo, currentEpoch] = await Promise.all([
+    payAccounts(synapse.client, { address: payerAddress }),
+    getBlockNumber(synapse.client, { cacheTime: 0 }),
```
juliangruber: `payAccounts` internally also fetches the block number. Could refactor this later on.

packages/synapse-core/src/warm-storage/terminate-service.ts
juliangruber, line 1: Claude noted that `WarmStorageSerivce.terminateService` is now unused internally, and wants to confirm it should stay part of the public api.

packages/synapse-core/src/sp/terminate-service.ts
```diff
@@ -0,0 +1,357 @@
...
+    retry: {
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+    },
+    poll: {
+      limit: RETRY_CONSTANTS.POLL_LIMIT,
+      interval: options.pollInterval ?? RETRY_CONSTANTS.POLL_INTERVAL,
+      statusCodes: [200],
+      shouldPoll: async (ctx) => {
+        const data = (await ctx.response.clone().json()) as TerminateServiceStatusResponse
```
juliangruber, line 341: potentially unsafe cast without schema validation

juliangruber (dismissed):
Mostly looking good to me, there's smaller things I'd like to see changed. As discussed yesterday, feel free to dismiss the change request if you want to move on

SgtPooki:
I addressed some of @juliangruber's comments in https://github.com/FilOzone/synapse-sdk/pull/830, I didn't want to push directly up to this PR knowing @rvagg was likely using it in his foc-devnet testing.

I do think the `onChain` name is odd, and think we should pivot to `skipProvider` or similar

rvagg:
@juliangruber @SgtPooki if you wouldn't mind taking this over and just pushing changes instead of doing blocking reviews that would help move this forward. I'm now travelling and have to focus on FDS so I'm going to slow things down if resolving red boxes relies on me. I trust your judgement.

packages/synapse-core/src/sp/terminate-service.ts
rvagg: ... try again? I had enough trouble trying to map the SP-side state machine here and exposing the gory details was part of the price. I think that if we end up with problems on this flow then we need to revisit how this was implemented in Curio and make the state transitions more clear and failure cases more obvious

docs/src/content/docs/developer-guides/storage/storage-operations.mdx
rvagg: Fair, the TL;DR is: doing it via the SP means you don't need to spend FIL, you spend USDFC _and_ you get immediate termination. There's other benefits, but they mostly fall on the SP, not the client, so not worth mentioning here.

docs/src/content/docs/developer-guides/migration-guide.md
rvagg: Aligning with what's in the contracts, it's "terminateService" in there; I think the reasoning is something like there being a whole service _associated_ with the data set, not just the data set itself; CDN for example expands it.
Mostly I've just been trying to align the terminology more through the stack so we don't have too many things using different names at different levels because it makes debugging, documentation and communication difficult.

juliangruber:
converted to draft to prevent accidental merge

hugomrdias:
i can take this an finish it

packages/synapse-core/src/warm-storage/terminate-service.ts
hugomrdias: eventually all methods from the WarmStorageService will be just core action but for now its ok to keep it

packages/synapse-core/src/sp/terminate-service.ts
hugomrdias: the user gets the same TimeoutError as in the other endpoints

packages/synapse-core/src/sp/terminate-service.ts
hugomrdias: for now this works the same as in the other endpoints with can improve this in a follow up PR

hugomrdias (approved):

PR #873, opened 2026-07-08
docs: add RPC endpoint best practices

### LINKED ISSUE #866: provide best practices for using external RPC endpoints
### Problem

Public RPC endpoints such as glif are limited to 100 RPM, but the SDK's RPC usage grows with the amount of data in an account, so ordinary usage quickly exceeds the limit:

| Scenario | Cause | Rough threshold |
|---|---|---|
| Listing/inspecting data sets (`findDataSets`, etc.) | ~2 RPC calls per data set, fired concurrently with no concurrency cap | ~50 data sets |
| Viewing a single data set's details (per-piece metadata reads) | 1 `eth_call` per piece, serially, N times | ~100 pieces in one data set |
| Resolving existing data sets during upload | Iterates all of the account's data sets, enriching each with on-chain reads | Triggered on every upload once data sets accumulate |

Both piece count and data set count grow monotonically with usage (every upload adds a piece), so **a user doing a bulk upload/migration hits 429s on day one** — this is not an edge case.

### Request

Add a **"Using an external RPC endpoint / avoiding rate limits" best-practices doc**, covering:

- How to pass a custom transport and enable batching: `http(url, { batch: true })`, client-level `batch: { multicall: true }`;
- How to use token-authenticated endpoints with higher limits, e.g. `https://api.node.glif.io/rpc/v1?token=<TOKEN>` or an `Authorization: Bearer` header.

beck-8:
## Summary

Adds a developer guide for using external RPC endpoints with Synapse SDK and avoiding public endpoint rate limits.

The guide covers:
- creating a viem client and passing it to `Synapse`
- enabling JSON-RPC HTTP batching with `http(url, { batch: true })`
- enabling viem multicall batching with `batch: { multicall: true }`
- using GLIF authenticated RPC tokens through URL query parameters
- using GLIF authenticated RPC tokens through `Authorization: Bearer` headers
- linking to the Filecoin Docs list of commonly used RPC endpoints
- practical tips for high-volume storage workflows

Also adds the page to the Developer Guides overview and adjusts sidebar ordering.

Closes #866.

## Validation

- Verified the documented viem client configuration type-checks locally.
- Verified GLIF Calibration RPC token authentication at runtime:
  - URL query token returned `eth_chainId = 0x4cb2f`
  - `Authorization: Bearer` token returned `eth_chainId = 0x4cb2f`
  - invalid token returned `403`
- Ran markdown lint for the new/updated docs pages:
  - `pnpm exec markdownlint-cli2 --config .github/.markdownlint-cli2.jsonc docs/src/content/docs/developer-guides/rpc-endpoints.md docs/src/content/docs/developer-guides/index.md`
- Confirmed local docs page returns `200 OK`:
  - `http://localhost:4321/developer-guides/rpc-endpoints/`

### DIFF
--- a/docs/src/content/docs/developer-guides/index.md
+++ b/docs/src/content/docs/developer-guides/index.md
@@ -12,6 +12,7 @@ This page describes the available Synapse packages and how to choose the one tha
 | Package | Best for | Builds on |
 | --------- | ---------- | ----------- |
 | **[Synapse SDK](/developer-guides/synapse/)** | Most applications | Core |
+| **[RPC Endpoints](/developer-guides/rpc-endpoints/)** | Production apps, bulk uploads, dashboards | Viem |
 | **[Synapse React](/developer-guides/synapse-react/)** | React applications | Core + Wagmi |
 | **[Synapse Core](/developer-guides/synapse-core/)** | Fine-grained control, custom integrations | Viem |
 
@@ -27,6 +28,12 @@ This page describes the available Synapse packages and how to choose the one tha
 [**Synapse SDK Guide →**](/developer-guides/synapse/)
 
 [**Synapse SDK API Reference →**](/reference/filoz/synapse-sdk/toc/)
+
+## RPC Endpoints
+
+Use authenticated RPC endpoints and viem batching for production apps, bulk uploads, migrations, and dashboards that inspect many data sets or pieces.
+
+[**RPC Endpoints Guide →**](/developer-guides/rpc-endpoints/)
 
 ## Synapse Core
 
--- a/docs/src/content/docs/developer-guides/rpc-endpoints.md
+++ b/docs/src/content/docs/developer-guides/rpc-endpoints.md
@@ -1,0 +1,124 @@
+---
+title: RPC Endpoints
+description: Best practices for using external RPC endpoints and avoiding rate limits.
+sidebar:
+  order: 3
+---
+
+Synapse SDK reads on-chain state while it prepares payments, finds or creates data sets, resolves providers, and inspects piece metadata. For accounts with many data sets or pieces, those reads can exceed the rate limits of public RPC endpoints.
+
+For production apps, bulk uploads, migrations, and dashboards, use a dedicated RPC endpoint with higher limits and enable batching in the viem client that you pass to Synapse.
+
+## Recommended setup
+
+Create a viem client yourself, configure its transport, and pass that client to `Synapse`. This gives you access to viem's HTTP batching, request headers, and multicall batching options.
+
+```ts twoslash
+import { Synapse, calibration } from '@filoz/synapse-sdk'
+import { createClient, http, type Hex } from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+
+// GLIF accepts authenticated RPC tokens in the URL query string.
+const rpcUrl = 'https://api.calibration.node.glif.io/rpc/v1?token=YOUR_TOKEN'
+const privateKey = '0x0000000000000000000000000000000000000000000000000000000000000001'
+
+const client = createClient({
+  account: privateKeyToAccount(privateKey as Hex),
+  chain: calibration,
+  transport: http(rpcUrl, {
+    batch: true,
+  }),
+  batch: {
+    multicall: true,
+  },
+})
+
+const synapse = new Synapse({
+  client,
+  source: 'my-app',
+})
+```
+
+This setup enables two different batching layers:
+
+- `transport: http(url, { batch: true })` batches multiple JSON-RPC requests into one HTTP request when the RPC provider supports JSON-RPC batch requests.
+- `batch: { multicall: true }` lets viem aggregate compatible `eth_call` contract reads through Multicall3.
+
+These settings reduce request count, but they do not remove the need for an endpoint with enough capacity for high-volume accounts.
+
+## Authenticated endpoints
+
+Most RPC providers offer authenticated endpoints with higher limits than public URLs. Prefer a token-authenticated endpoint for production and migration workloads. The Filecoin Docs maintain a list of commonly used [Filecoin RPC endpoints](https://docs.filecoin.io/networks-and-tools/networks/mainnet/rpcs).
+
+GLIF accepts RPC tokens in the URL query string. Pass the authenticated URL directly to `http()`:
+
+```ts twoslash
+import { Synapse, calibration } from '@filoz/synapse-sdk'
+import { createClient, http, type Hex } from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+
+const privateKey = '0x0000000000000000000000000000000000000000000000000000000000000001'
+
+const client = createClient({
+  account: privateKeyToAccount(privateKey as Hex),
+  chain: calibration,
+  transport: http('https://api.calibration.node.glif.io/rpc/v1?token=YOUR_TOKEN', {
+    batch: true,
+  }),
+  batch: { multicall: true },
+})
+
+const synapse = new Synapse({ client, source: 'my-app' })
+```
+
+GLIF also accepts the same token in an `Authorization: Bearer` header. Configure `fetchOptions.headers` on the viem HTTP transport:
+
+```ts twoslash
+import { Synapse, calibration } from '@filoz/synapse-sdk'
+import { createClient, http, type Hex } from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+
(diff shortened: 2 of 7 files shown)


beck-8:
image

rvagg:
why are we not seeing the full CI run on this ..

docs/src/content/docs/developer-guides/index.md
```diff
@@ -12,6 +12,7 @@ This page describes the available Synapse packages and how to choose the one tha
 | Package | Best for | Builds on |
 | --------- | ---------- | ----------- |
 | **[Synapse SDK](/developer-guides/synapse/)** | Most applications | Core |
+| **[RPC Endpoints](/developer-guides/rpc-endpoints/)** | Production apps, bulk uploads, dashboards | Viem |
```
rvagg, line 15: it's not a published "package" like the others so doesn't belong here


```suggestion
```

docs/src/content/docs/developer-guides/rpc-endpoints.md
```diff
@@ -0,0 +1,124 @@
...
+Public endpoints are useful for examples and small tests, but ordinary production usage can outgrow them quickly:
+
+| Workflow | Why it can create many RPC reads |
+| --------- | -------------------------------- |
+| Listing data sets | The SDK reads data set IDs and enriches them with on-chain metadata. |
+| Inspecting a large data set | Piece metadata reads scale with the number of pieces in the data set. |
+| Uploading to an account with existing data sets | The SDK resolves reusable data sets before creating new ones. |
+| Dashboards and migrations | Repeated listing, filtering, and inspection can multiply reads across many data sets. |
+
+For these workflows, use an authenticated endpoint, avoid tight polling loops, and cache known data set IDs or piece information when your application can tolerate cached state.
```
rvagg, line 113: do we dare suggest that for heavy, large-scale workloads, that they considering installing Forest, or Lotus nodes to have their own trusted, local, un-capped endpoint?

beck-8: I considered it yesterday, but it felt a bit heavy on them, especially for developers who are just getting to know them. However, adding this part here is not a problem; it just gives them another option, which I will add later.

docs/src/content/docs/developer-guides/rpc-endpoints.md
```diff
@@ -0,0 +1,124 @@
+---
+title: RPC Endpoints
+description: Best practices for using external RPC endpoints and avoiding rate limits.
+sidebar:
+  order: 3
```
rvagg, line 5: I think you should put this in position 5, after the packages themselves

rvagg:
Good content, but there's a fundamental limitation that I think we need to figure out a path through before we adopt docs like this. The "recommended" path is only suitable for server-side apps and other situations where you have a local private key you can feed in, it's not helpful for wallet connected clients, like MetaMask, which has its own RPC, which you _can't_ just patch to point to your authenticated RPC endpoint (a dapp developer might be able to document all of the ways users of their dapp could configure their wallets to talk to an authenticated endpoint and how to sign up for an endpoint, but that's pretty tedious and not likely to help cause average users just don't grok this stuff).

The batching and multicall advice is still useful. `batch: { multicall: true }` can reduce compatible contract reads, including for wallet-backed clients if the client is constructed that way. But it does not solve the root problem that triggered this issue: browser wallet flows still use the wallet provider's RPC endpoint, so glif public rate-limit reductions can still hurt users.

Comes back to https://github.com/FilOzone/synapse-sdk/issues/504 and I don't think we really fully resolved that to solve for this problem in the sdk at least. In synapse-core at least you can choose what client you pass in, but the sdk takes one `client` option.

There seem to be two possibilities here we can pursue:

1. Build or expose a hybrid viem client/transport that routes read-like RPC methods to an app-controlled authenticated RPC endpoint and wallet methods to the wallet provider. This might work, but it is gnarly and I would prefer not to roll our own if viem/wagmi already has a blessed pattern for this.
2. Add first-class split client support to Synapse, probably as `publicClient` and `walletClient`. IIRC @hugomrdias suggested this at some point? Then reads, discovery, metadata, pricing, simulation, and receipt polling use `publicClient`, while signing and transaction submission use `walletClient`. This would let browser apps use an authenticated/proxied RPC endpoint for read-heavy FOC work while keeping wallet UX for signatures and transactions.

 Until we have that, I think this page should be scoped more narrowly:

 - Present the current `createClient + privateKeyToAccount + authenticated http()` setup as the server/CLI/migration/service-account path (clearly stated)
 - Keep the batching and multicall advice, with caveats.
 - Add an explicit browser-wallet limitation section.
 - Avoid implying that this solves authenticated RPC for MetaMask-connected Synapse apps today.

I'll defer to @hugomrdias for more input, he's thought more about this than me. I think I'd be happy adding option 2 and I don't think it'd be too hard with what we have now.

rvagg:
> why are we not seeing the full CI run on this ..

PR incoming

rvagg:
Even a split client still doesn't fully solve the fundamental problem because I don't think we can practically advise dapp developers to push their glif RPC token out to their clients. Clients are either going to have to configure their own RPC (which I don't think scales well as advice), dapps may have to proxy RPC and expose that (kind of gross), or we have some kind of graceful rate limit handling that slows down calls when you hit limits (a bit gross and prone to breakage, but we could pull something basic off).

OR we could limit scope of this even further and be clear that caring about this is for the cases of high volume usage, particularly where data sets get very large or you have lots of data sets. The common case _should_ be fine, but "if you encounter 429 rate limits then ..." might be the right framing for this doc instead. Then our job becomes focusing on how we can limit the number of RPC calls to get it under threshold as much as possible.

beck-8:
Yes, my starting point is that I am worried that glif will be more likely to have this problem after reducing the RPM.

After waiting for more ideas to enter, I will to update this section.

hugomrdias:
After seeing this PR https://github.com/FilOzone/synapse-sdk/pull/876, i agree with @rvagg we need to go back to https://github.com/FilOzone/synapse-sdk/issues/504 and figure this out better.

I will do further research on this.

docs/src/content/docs/developer-guides/rpc-endpoints.md
```diff
@@ -0,0 +1,124 @@
...
+// GLIF accepts authenticated RPC tokens in the URL query string.
+const rpcUrl = 'https://api.calibration.node.glif.io/rpc/v1?token=YOUR_TOKEN'
+const privateKey = '0x0000000000000000000000000000000000000000000000000000000000000001'
+
+const client = createClient({
+  account: privateKeyToAccount(privateKey as Hex),
+  chain: calibration,
+  transport: http(rpcUrl, {
+    batch: true,
+  }),
```
hugomrdias, line 30: json rpc batching doesnt help here, it still counts to limits on most rpcs and doesn't really make things faster or better.

docs/src/content/docs/developer-guides/rpc-endpoints.md
```diff
@@ -0,0 +1,124 @@
...
+  batch: {
+    multicall: true,
+  },
```
hugomrdias, line 33: enable multicall for viem is fine but for wagmi its already enabled by default

hugomrdias: on the sdk we can probably enable this automatically

docs/src/content/docs/developer-guides/rpc-endpoints.md
```diff
@@ -0,0 +1,124 @@
...
+      headers: {
+        Authorization: 'Bearer YOUR_TOKEN',
+      },
+    },
+  }),
+  batch: { multicall: true },
+})
+
+const synapse = new Synapse({ client, source: 'my-app' })
+```
```
hugomrdias, line 98: We should point users to the viem docs https://viem.sh/docs/clients/transports/http#usage

hugomrdias (commented):
There's two very different use cases that we need to consider Dapps and backend services.

Dapps need public nodes and should use the [fallback](https://viem.sh/docs/clients/transports/fallback) transport

Backend services can use privates node and probably also the fallback transport or something like ERPC

We need to split these out and also take into considerations that Dapps use Wagmi and not viem directly.


The bigger issue here is how we handle the "clients" in the SDK we should not make reads go through the wallets, we can do that by default and provide a config if the developer wants to configure the "read client" manually

hugomrdias:
ok i created an issue to take care of this for good https://github.com/FilOzone/synapse-sdk/issues/895 we can close this PR as the issue includes using these changes as the base for a future documentation improvement.

PR #875, opened 2026-07-09
ci: add docs PR checks and trusted preview deploys

rvagg:
Trying to do two things here:

1. Run in-repo docs validation on every PR, including fork PRs.
  - `docs-check` always runs the markdown lint for docs-relevant changes.
  - It also builds the docs when full CI is not already going to do that.
  - This gives us GitHub-visible lint/build failures, which the existing Cloudflare preview integration does not provide.
2. Allow manual preview publishing for a reviewed/trusted PR.
  - Same-repo branches already get Cloudflare Workers preview URLs from the existing Cloudflare Git integration.
  - Fork PRs do not get those previews, and we should not expose Cloudflare secrets to automatic fork PR workflows.
  - The manual `documentation` workflow now accepts a PR number and reviewed commit SHA, checks that the PR head still matches that SHA, restores the trusted Wrangler config from `master`, then uploads a non-production preview version.

### DIFF
--- a/.github/workflows/docs-check.yml
+++ b/.github/workflows/docs-check.yml
@@ -1,11 +1,13 @@
+# Cloudflare Workers Git integration already builds same-repo PR previews outside this repo.
+# This workflow is the in-repo docs lint/build check, covers fork PRs without Cloudflare secrets,
+# and skips docs builds when full CI already covers them.
 name: docs-check
 
 on:
   pull_request:
-    paths:
-      - "docs/**"
-      - .github/workflows/docs-check.yml
-      - .github/workflows/docs.yml
+concurrency:
+  group: docs-check-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
 
 permissions:
   contents: read
@@ -13,10 +15,11 @@ permissions:
 
 jobs:
   changes:
-    name: Detect docs check requirement
+    name: Detect docs changes
     runs-on: ubuntu-latest
     outputs:
-      docs_check_required: ${{ steps.detect.outputs.docs_check_required }}
+      docs_relevant_changed: ${{ steps.detect.outputs.docs_relevant_changed }}
+      docs_build_required: ${{ steps.detect.outputs.docs_build_required }}
     steps:
       - name: Detect changed paths
         id: detect
@@ -50,16 +53,18 @@ jobs:
           $files
           EOF
 
+          echo "docs_relevant_changed=${docs_relevant_changed}" >> "$GITHUB_OUTPUT"
+
           if [ "$docs_relevant_changed" = true ] && [ "$full_ci_changed" = false ]; then
-            echo "docs_check_required=true" >> "$GITHUB_OUTPUT"
+            echo "docs_build_required=true" >> "$GITHUB_OUTPUT"
           else
-            echo "docs_check_required=false" >> "$GITHUB_OUTPUT"
+            echo "docs_build_required=false" >> "$GITHUB_OUTPUT"
           fi
 
-  build-docs:
-    name: Build docs
+  check-docs:
+    name: Check docs
     needs: changes
-    if: needs.changes.outputs.docs_check_required == 'true'
+    if: needs.changes.outputs.docs_relevant_changed == 'true'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -12,19 +23,60 @@ jobs:
...
+              ;;
+          esac
+      - name: Checkout Repository
+        if: inputs.pr_number == ''
+        uses: actions/checkout@v7
+      - name: Checkout trusted PR
+        if: inputs.pr_number != ''
+        uses: actions/checkout@v7
+        with:
+          ref: refs/pull/${{ inputs.pr_number }}/head
```
BigLep, line 67: Is this a security concern we need to worry about? It's definitely not as convenient...

`refs/pull/N/head` is a moving ref, so there is a review-then-dispatch race: the PR author can push new commits after a maintainer reviews but before (or while) this runs, and it is the new, unreviewed code that executes with the Cloudflare secrets in scope. Since the whole safety model here is "reviewed and trusted", consider taking a full commit SHA as the input (validated as 40 hex chars, like the numeric check above) and checking that out instead, so what deploys is exactly what was reviewed.

.github/workflows/docs.yml
```diff
@@ -12,19 +23,60 @@ jobs:
...
       - run: pnpm run build
       - run: pnpm -r --filter docs run build
-      - name: Deploy docs
+      - name: Deploy production docs
+        if: inputs.pr_number == ''
         run: pnpm --dir docs exec wrangler deploy --env production
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+      - name: Deploy preview docs
```
BigLep, line 97: Other AI input I got:

Note that the target worker for this upload comes from the PR's own `docs/wrangler.json` (`name: synapse-dev`). A PR that edits that file could point `versions upload` at the production `synapse` worker instead. Uploaded versions do not serve traffic until promoted, so the blast radius is limited, but if you want belt-and-braces you could restore the trusted config first (e.g. `git checkout origin/master -- docs/wrangler.json` before deploying).

.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,78 @@
+name: docs-check
```
BigLep, line 4: Nit: consider adding a `concurrency` group keyed on the PR (e.g. `docs-check-${{ github.event.pull_request.number }}` with `cancel-in-progress: true`) so rapid successive pushes do not stack redundant builds.

.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,78 @@
...
+
+on:
+  pull_request:
+    paths:
+      - "docs/**"
```
BigLep: AI feedback I got. Take it or leave it: If the plan is to eventually make `Build docs` a required check for merging, the workflow-level `paths:` filter will be a problem: on PRs that touch nothing docs-related the workflow never runs at all, so branch protection would hold them at "Expected - waiting for status" forever. If it stays advisory this is fine as-is. If it becomes required, the usual fix is to drop the `paths:` filter and let the `changes` job (which you already have) decide, since a skipped `Build docs` job satisfies a required check while a never-reported one does not.

BigLep (commented):
Thanks for this. Could you expand the PR description (or a comment in the workflows) on the motivation? In particular the relationship to the existing Cloudflare Workers Builds Git integration: IIUC today that is what builds docs PRs and posts preview URLs (e.g. what I got with #874), but it is configured outside the repo and does not cover fork PRs. Is the intent that we stop relying on that integration, and if so should it be disabled once this lands so we do not run two parallel build pipelines? Having that rationale written down will help the next person understand why both docs-check.yml and the Workers Builds checks exist.

BigLep (approved):

.github/workflows/docs.yml
rvagg: yes and no, I didn't add it cause it's friction and very low probability of exploitability; but I guess we'll have people reporting security bugs to us because this is a standard security heuristic, so I've added the unfortunate friction

rvagg:
Updated the original post with more details. What I really want is actual checking / linting of docs on docs PRs, it's extremely annoying when a doc change lands that breaks lint and we end up having to fix it in an unrelated code PR. Plus (the thing that actually triggered this), the ability to see the docs for fork PRs like for #873 - in that case I have no trust issues with the author, I just want to be able to click something to see the preview.

PR #876, opened 2026-07-10
fix: omit sender from read-only contract calls

### LINKED ISSUE #868: Read-only payment/storage calls fail for Safe owner addresses on Filecoin when eth_call includes from
While working through https://github.com/FilOzone/dealbot/issues/627, we hit an issue using a Safe multisig as the wallet owner with a registered session key.

Related: https://github.com/FilOzone/synapse-sdk/issues/714 mentions the same underlying multisig/FWSS friction from a documentation angle. This issue is the narrower SDK/runtime bug we ran into while trying to use `filecoin-pin` against the Safe.

## Setup

- Safe owner address: `0x37c5835DBd07c5f1E9525aA2852c960BCfD1177f`
- Session key registered via Safe
- Safe has FIL
- Safe has USDFC in the wallet and deposited in Filecoin Pay
- FWSS operator approval is max-approved

## What happened

Running payment status through `filecoin-pin` with the Safe owner and session key reported that the Safe had no USDFC:

```sh
pnpm exec filecoin-pin payments status \
  --network mainnet \
  --wallet-address "$WALLET_ADDRESS" \
  --session-key "$SESSION_KEY"
```

Result:

```text
No USDFC tokens found
```

But direct accountless contract reads showed the expected balances and approvals:

- USDFC wallet balance: `0.04`
- Filecoin Pay balance: `1`
- FWSS operator approval: max-approved

## Diagnosis

The failing reads appear to happen because SDK read-only calls use a client with the Safe address as the account/from address. On Filecoin, some `eth_call`s with the Safe contract as sender fail with:

```text
SysErrSenderInvalid(1)
pre-validation failed: Send not from valid sender
```

We reproduced this locally: direct accountless viem reads returned the expected balances, while the SDK/filecoin-pin payment read path returned zero or failed. This caused `filecoin-pin` preflight to block uploads incorrectly.

A local workaround was to make read-only payment/storage calls use an accountless read client, similar in spirit to the existing workaround in `session-key/authorization-expiry.ts`, while leaving signing/write clients unchanged.

Affected areas observed locally:

- `PaymentsService.walletBalance`
- `PaymentsService.accountInfo`
- `PaymentsService.serviceApproval`
- `WarmStorageService.getServicePrice`
- storage-manager cost/preflight reads

## Expected behavior

Read-only SDK calls should not require a valid transaction sender and should work for Safe/contract-wallet owner addresses.

## Actual behavior

Read calls with a Safe owner can fail or be interpreted as zero balances, causing `filecoin-pin` preflight to report `No USDFC tokens found` and block uploads even when balances and approvals are present.

rjan90:
Closes #868, see https://github.com/FilOzone/synapse-sdk/issues/868#issuecomment-4875349965 for what I worked towards with this PR.

## Summary

- add a shared accountless read client that preserves transport retries, batching, method filters, and client configuration
- route synapse-core and direct synapse-sdk contract reads through accountless clients
- keep transaction simulations and writes on the signed client
- add regression coverage for contract-wallet reads and read/write client boundaries

## Root cause

Viem uses a client's account as the default `from` address for `eth_call`. Lotus currently applies sender pre-validation to these calls and rejects Safe/contract-wallet addresses with `SysErrSenderInvalid`, even when the contract call is read-only. This caused valid balances, approvals, provider data, and storage state to appear unavailable.

## Impact

Read-only SDK operations no longer require the wallet address to be a valid Filecoin transaction sender. Safe and other contract-wallet owners can use payment, storage, PDP, and provider-registry reads without changing signing or write behavior.


packages/synapse-sdk/src/synapse.ts
```diff
@@ -115,6 +118,16 @@ export class Synapse {
...
+  /**
+   * Accountless client for read-only RPC calls.
+   *
+   * Unlike {@link client}, this client does not add the wallet address as the
+   * default `from` value for contract calls.
+   */
+  get readClient(): Client<Transport, Chain, undefined, PublicRpcSchema, PublicActions<Transport, Chain>> {
+    return this._readClient
+  }
+
```
rvagg: Let's remove this unless you think there's a really good reason for it. More external API surface area means more to care about going forward and they're hard to remove once exposed. `import { toReadClient } from '@filoz/synapse-core/utils'` is available for anyone that wants to use it, including filecoin-pin.

rvagg (approved):
very nice, only two items:
* unnecessary (as far as I can tell) new public API `Synapse#readClient()`, let's prefer not to add more external APIs unless necessary
* rebase on master, you'll probably have to deal with a conflict because Russel merged a fix for a lint issue that I think you've bundled in here

otherwise g2g

packages/synapse-sdk/src/synapse.ts
rjan90: Done in https://github.com/FilOzone/synapse-sdk/pull/876/changes/d609d43e7fabdc67c3f0bd6483f726184049efa7

PR #883, opened 2026-07-16
chore(deps): pin non-core dependencies by version, slow down dependabot

rvagg:
Attempt to deal with dependency churn and conflicts we experience in the repo
while resolving tensions we have in what we ship: As libraries, synapse-core
and synapse-sdk have a floating dependency tree where they are used. We could
choose to have a lockfile here for everything but that would only test _our_
use of these packages, not how they get used upstream. Any breakage, even via
transitive dependencies, don't get noticed. It's through us experiencing
the breakage that we get to deal with it.
Unfortunately we also have some very noisy sub-packages in this repo with a
significant dependency trees of their own which cause most of our transitive
breakage. So in an attempt to slow that problem down, we pin our direct
dependencies in the packages other than synapse-sdk and synapse-core, so it's
only breakage via transitives that we experience outside of dependabot updates.

The alternative is to separate out the core libary packages from the extraneous
stuff, particularly docs, which are the source of most of our grief.

This strategy is already working with biome, which is _very_ break-happy.

---

Also pulls in the outstanding dependabot commits.


docs/src/content/docs/core-concepts/storage-providers.mdx
```diff
@@ -26,6 +26,7 @@ A non-approved provider is registered in the [ServiceProviderRegistry](/core-con
 - Can still be used explicitly if you pass their `providerId` to lower-level APIs, though this bypasses the quality vetting that approval represents.
 
 It is possible for a provider to be removed from the approved list. Removal does not deactivate existing storage deals and does not prevent further interaction with that provider via the SDK but removal may be an indication of a lowering of service quality.
+
 ## Approved Providers
```
SgtPooki, line 30: change should be in separate PR?

SgtPooki (approved):
this is a decent enough solution for now.

docs/src/content/docs/core-concepts/storage-providers.mdx
rvagg: I pulled in the 3 outstanding dependabot PRs and this is a necessary lint failure fix

rvagg:
pulled in all the astro stuff here including the v7 bump which had a few behaviour and config changes, nothing major

PR #886, opened 2026-07-17
fix(sdk): harden provider health checks

### LINKED ISSUE #849: SmartSelect: intermittent ping failures cause "No endorsed provider available" even when providers are reachable
## Summary

`StorageContext.smartSelect` intermittently marks endorsed providers as unreachable due to failed pings, even when the same providers respond successfully on retry. This causes uploads to fail with a hard error rather than retrying or falling back gracefully.

## Error

```
Provider 0xCb9e86945cA31E6C3120725BF0385CBAD684040c (ID: 4) failed ping: Ping failed
Provider 0xa3971A7234a3379A1813d9867B531e7EeB20ae07 (ID: 9) failed ping: Ping failed
Error: StorageContext smartSelect failed: No endorsed provider available — all endorsed provider(s) failed health check
    at createError (file:///Users/nick-w3/foc-cli/cli/node_modules/@filoz/synapse-sdk/dist/src/utils/errors.js:8:22)
    at StorageContext.smartSelect (file:///Users/nick-w3/foc-cli/cli/node_modules/@filoz/synapse-sdk/dist/src/storage/context.js:290:23)
    at async StorageContext.createContexts (file:///Users/nick-w3/foc-cli/cli/node_modules/@filoz/synapse-sdk/dist/src/storage/context.js:109:27)
    at async StorageManager.createContexts (file:///Users/nick-w3/foc-cli/cli/node_modules/@filoz/synapse-sdk/dist/src/storage/manager.js:546:26)
```

Image

## Behavior

- Pings to endorsed providers fail intermittently, not consistently.
- Retrying the same operation (e.g. re-running the upload command) succeeds most of the time.
- This points to the ping retry policy being too tight (timeout too short, retry count too low, or no backoff) rather than providers genuinely being down.

## Workaround

A custom provider selection implementation was introduced in `foc-cli` and `foc-upload-dapp` as a workaround:

- [`[cli/src/provider-selection.ts](https://github.com/FIL-Builders/foc-cli/blob/main/cli/src/provider-selection.ts)`](https://github.com/FIL-Builders/foc-cli/blob/main/cli/src/provider-selection.ts)
- [`[cli/src/commands/multi-upload.ts#L134-L155](https://github.com/FIL-Builders/foc-cli/blob/main/cli/src/commands/multi-upload.ts#L134C2-L155C9)`](https://github.com/FIL-Builders/foc-cli/blob/main/cli/src/commands/multi-upload.ts#L134C2-L155C9)

The key fix was passing `providerIds` directly into `createContexts` instead of passing copies, which appears to avoid the condition that triggers false ping failures.

## Suggested Fix

Review the ping retry policy inside `smartSelect` (timeout, retry count, backoff). Consider:

- Increasing the ping timeout threshold
- Adding at least one retry with a short delay before marking a provider as failed

### LINKED ISSUE #850: console.warn on provider ping failure corrupts caller terminal UI during selection
### Description

During provider selection in `StorageContext`, a failed `SP.ping` is logged with a raw `console.warn` before the provider is excluded. This writes to stderr at an arbitrary moment, which corrupts callers that render their own terminal UI (spinners, progress lines).

Source: packages/synapse-sdk/src/storage/context.ts:610-617

```js
} catch (error) {
  console.warn(
    `Provider ${candidate.provider.serviceProvider} (ID: ${candidate.provider.id}) failed ping:`,
    error instanceof Error ? error.message : String(error)
  )
  excludeProviderIds.push(candidate.provider.id)
  pingFailures++
}
```

### Impact

Any consumer drawing interactive output during `executeUpload` gets its rendering punched through whenever a candidate provider is unreachable. Reported downstream in filecoin-project/filecoin-pin#429, where a ping failure interleaves with clack spinner output mid-upload.

### Proposed approach

Stop writing selection diagnostics to `console` directly:

- route the message through the SDK's logger so consumers can configure or silence it, or
- surface ping failures as a provider-selection progress event the caller can render itself.

Either keeps selection logic intact (still excludes the failed provider, still counts `pingFailures`) while leaving terminal output to the consumer. Worth auditing for the same `console.warn` pattern elsewhere in selection.

rjan90:
## Summary

- raise the PDP ping deadline from 1s to 8s and retry fast failures twice with 500ms exponential backoff
- let low-level callers override the total ping timeout while preserving the existing `ping(serviceURL)` call shape
- surface the original HTTP/network/timeout error instead of replacing it with a generic `Ping failed`
- ping the ranked endorsed-provider pool concurrently while still choosing the highest-ranked healthy provider
- remove the direct `console.warn` from provider selection
- include provider-specific causes when every endorsed health check fails

## Root cause

The ping helper used a one-second total deadline around the request and retries. On a small global provider pool this made transient latency look like provider unavailability. `smartSelect` then checked endorsed candidates serially and wrote recovered failures directly to stderr, which could corrupt caller-owned terminal UIs.

## Validation

- `pnpm -C packages/synapse-core lint`
- `pnpm -C packages/synapse-sdk lint`
- `pnpm -C packages/synapse-core test` (Node and browser)
- `pnpm -C packages/synapse-sdk test` (Node and browser)

Fixes #849
Fixes #850

### DIFF
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -648,7 +648,7 @@ export class StorageContext {
         const candidates = selectProviders({
           ...input,
           endorsedIds: endorsedSlot ? input.endorsedIds : [],
-          count: endorsedSlot ? Math.max(1, input.endorsedIds.length) : 1,
+          count: endorsedSlot ? input.endorsedIds.length : 1,
           excludeProviderIds,
           metadata,
         })
@@ -667,8 +667,7 @@ export class StorageContext {
         // Await in selection order so a faster, lower-ranked provider cannot
         // displace a healthy higher-ranked provider. The checks themselves are
         // already in flight, so a failed candidate does not delay starting the next.
-        for (const healthCheck of healthChecks) {
-          const outcome = await healthCheck
+        for await (const outcome of healthChecks) {
           if (outcome.ok) {
             results.push(StorageContext.toProviderSelectionResult(outcome.candidate))
             excludeProviderIds.push(outcome.candidate.provider.id)
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -641,6 +641,8 @@ export class StorageContext {
       // Keep selecting and pinging until a healthy provider is found
       // or all candidates are exhausted
       for (;;) {
+        if (endorsedSlot && input.endorsedIds.length === 0) break
+
         // There are few endorsed providers, so check the complete ranked pool
         // concurrently. Other slots keep the existing one-at-a-time behavior.
         const candidates = selectProviders({
--- a/packages/synapse-sdk/src/storage/context.ts
+++ b/packages/synapse-sdk/src/storage/context.ts
@@ -628,53 +628,78 @@ export class StorageContext {
       address: clientAddress,
     })
 
-    // Inline ping-retry loop: select a candidate from core, ping it,
-    // exclude on failure, re-select. One outer iteration per copy needed.
+    // Select a healthy candidate for each copy. Endorsed primary candidates
+    // are pinged concurrently, while results are consumed in ranked order.
     const results: ProviderSelectionResult[] = []
     const excludeProviderIds = [...options.excludeProviderIds]
 
     for (let i = 0; i < count; i++) {
       const endorsedSlot = requireEndorsedPrimary && i === 0
       let found = false
-      let pingFailures = 0
+      const pingFailures: Array<{ candidate: ResolvedLocation; error: unknown }> = []
 
       // Keep selecting and pinging until a healthy provider is found
       // or all candidates are exhausted
       for (;;) {
+        // There are few endorsed providers, so check the complete ranked pool
+        // concurrently. Other slots keep the existing one-at-a-time behavior.
         const candidates = selectProviders({
           ...input,
           endorsedIds: endorsedSlot ? input.endorsedIds : [],
-          count: 1,
+          count: endorsedSlot ? Math.max(1, input.endorsedIds.length) : 1,
           excludeProviderIds,
           metadata,
         })
 
         if (candidates.length === 0) break
 
-        const candidate = candidates[0]
-        try {
-          await SP.ping(candidate.provider.pdp.serviceURL)
-          results.push(StorageContext.toProviderSelectionResult(candidate))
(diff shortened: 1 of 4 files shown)


packages/synapse-sdk/src/storage/context.ts
```diff
@@ -628,53 +628,80 @@ export class StorageContext {
...
       for (;;) {
+        if (endorsedSlot && input.endorsedIds.length === 0) break
+
+        // There are few endorsed providers, so check the complete ranked pool
+        // concurrently. Other slots keep the existing one-at-a-time behavior.
         const candidates = selectProviders({
           ...input,
           endorsedIds: endorsedSlot ? input.endorsedIds : [],
-          count: 1,
+          count: endorsedSlot ? Math.max(1, input.endorsedIds.length) : 1,
```
hugomrdias: replace Math.max(1, input.endorsedIds.length) with input.endorsedIds.length; zero is already impossible there.

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -628,53 +628,80 @@ export class StorageContext {
...
+            return { ok: true as const, candidate }
+          } catch (error) {
+            return { ok: false as const, candidate, error }
+          }
+        })
+
+        // Await in selection order so a faster, lower-ranked provider cannot
+        // displace a healthy higher-ranked provider. The checks themselves are
+        // already in flight, so a failed candidate does not delay starting the next.
+        for (const healthCheck of healthChecks) {
```
hugomrdias: Use for await (const outcome of healthChecks) instead of iterating promises and awaiting each inside the loop. It preserves concurrent startup and ranked consumption.

hugomrdias (changes requested):

packages/synapse-sdk/src/storage/context.ts
rjan90: Addressed in [`215c6db`](https://github.com/FilOzone/synapse-sdk/commit/215c6dba5bc2ad93bc0550002a48ed2be6754234): removed the redundant `Math.max(1, ...)` now that the empty endorsed-provider case is guarded.

packages/synapse-sdk/src/storage/context.ts
rjan90: Addressed in [`215c6db`](https://github.com/FilOzone/synapse-sdk/commit/215c6dba5bc2ad93bc0550002a48ed2be6754234): switched ranked health-check consumption to `for await...of` while retaining eager concurrent startup.

hugomrdias (approved):

PR #897, opened 2026-07-22
fix(synapse-core): validate PDP verifier address checksum

### LINKED ISSUE #761: Remove loose address validate for pdp verifier in wagmi.config.ts
https://github.com/FilOzone/synapse-sdk/blob/fada7cdcaf67addd1edd83aa483c28ff9e4ef5c4/packages/synapse-core/wagmi.config.ts#L22

EIP-55 casing being addressed in https://github.com/FilOzone/filecoin-services/pull/464; next time we should be able to remove the lose version (although note it'll technically be a breaking change when we remove it because it's exported - we could consider not exporting it and just defining it locally as a stopgap)

hugomrdias:
## What changed

- validate `PDP_VERIFIER_IMPLEMENTATION_ADDRESS` with the strict EIP-55 address schema
- stop using the loose address validator in the Wagmi deployment configuration
- retain the public `zAddressLoose` export to avoid an unrelated breaking API change

## Why

The PDP verifier implementation addresses previously lost their checksum casing in the upstream deployment manifest, requiring a temporary loose validator. The upstream manifest now enforces and publishes valid EIP-55 checksums, so the exception is no longer necessary.

This makes ABI generation fail early if a future deployment manifest contains an invalid mixed-case PDP verifier implementation address.

Closes #761

## Validation

- `pnpm --filter @filoz/synapse-core run generate-abi`
- `pnpm exec biome check packages/synapse-core/wagmi.config.ts`
- `git diff --check`

### DIFF
--- a/packages/synapse-core/wagmi.config.ts
+++ b/packages/synapse-core/wagmi.config.ts
@@ -4,7 +4,7 @@ import { request } from 'iso-web/http'
 import { zeroAddress } from 'viem'
 import * as z from 'zod'
 import { ZodValidationError } from './src/errors/base.ts'
-import { zAddress, zAddressLoose } from './src/utils/schemas.ts'
+import { zAddress } from './src/utils/schemas.ts'
 
 // GIT_REF can be one of: '<branch name>', '<commit>' or 'tags/<tag>'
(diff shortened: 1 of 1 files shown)


rvagg (approved):
👌

PR #898, opened 2026-07-22
feat: add batch piece deletion

### LINKED ISSUE #551: Add delete multiple pieces in one go
our eip712 `SchedulePieceRemovals` supports deleting multiple pieces but curio pdp api only support deleting one by one

hugomrdias:
## Summary

- add batched `deletePieces` and `schedulePieceDeletions` operations to synapse-core
- expose `StorageContext.deletePieces()` while preserving the existing singular API as a compatibility wrapper
- deduplicate IDs before signing and submission, enforce Curio batch/range limits, and preserve bigint precision in JSON requests
- document PDPVerifier’s cumulative pending-removal limit and add core/SDK coverage

## Why

The EIP-712 `SchedulePieceRemovals` authorization already supports multiple piece IDs, but the SDK only submitted one piece at a time. Curio PR filecoin-project/curio#1309 added a compatible `pieceIds` request field, allowing one authorization and one on-chain transaction to schedule up to 500 removals.

Closes #551.

## Developer impact

Consumers can now call:

```ts
await context.deletePieces({ pieces: [1n, 2n, pieceCid] })
```

The existing `deletePiece()` and `schedulePieceDeletion()` APIs remain available and delegate to the plural implementations.

## Validation

- `pnpm run build`
- `pnpm test`
- `pnpm run lint:fix` in `packages/synapse-core` and `packages/synapse-sdk`
- core: 836 passing
- SDK: 277 passing, 7 pending

The root lint command still reports pre-existing accessibility errors in documentation SVG assets; no documentation assets are changed by this PR.

### DIFF
--- a/packages/synapse-core/src/sp/schedule-piece-deletion.ts
+++ b/packages/synapse-core/src/sp/schedule-piece-deletion.ts
@@ -1,14 +1,17 @@
-import { HttpError, type RequestJsonErrors, request } from 'iso-web/http'
+import { HttpError, type RequestErrors, request } from 'iso-web/http'
 import type { Account, Chain, Client, Hex, Transport } from 'viem'
 import { DeletePieceError } from '../errors/pdp.ts'
+import { AtLeastOnePieceRequiredError, TooManyPiecesError } from '../errors/warm-storage.ts'
 import { signSchedulePieceRemovals } from '../typed-data/sign-schedule-piece-removals.ts'
-import { RETRY_CONSTANTS } from '../utils/constants.ts'
+import { RETRY_CONSTANTS, SIZE_CONSTANTS } from '../utils/constants.ts'
 
-export namespace deletePiece {
+const MAX_CURIO_PIECE_ID = (1n << 63n) - 1n
+
+export namespace deletePieces {
   export type OptionsType = {
     serviceURL: string
     dataSetId: bigint
-    pieceId: bigint
+    pieceIds: bigint[]
     extraData: Hex
     /** The number of retries. Defaults to 2. */
     retryCount?: number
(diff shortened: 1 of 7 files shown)


packages/synapse-core/src/sp/schedule-piece-deletion.ts
```diff
@@ -108,19 +148,52 @@ export namespace schedulePieceDeletion {
...
       clientDataSetId: options.clientDataSetId,
-      pieceIds: [options.pieceId],
+      pieceIds,
     }),
     retryCount: options.retryCount,
     retryDelay: options.retryDelay,
   })
 }
+
+export namespace deletePiece {
```
rvagg, line 172: you could mark all of the items below here as `@deprecated`

packages/synapse-core/src/utils/constants.ts
```diff
@@ -99,6 +99,14 @@ export const SIZE_CONSTANTS = {
    */
   MAX_ADD_PIECES_BATCH_SIZE: 40,
 
+  /**
+   * Maximum pieces per schedulePieceDeletions call accepted by the Curio PDP API.
+   *
+   * PDPVerifier also limits a data set to 2,000 cumulative queued removals until
+   * its next proving period, so a valid batch can still fail when that queue is full.
+   */
+  MAX_DELETE_PIECES_BATCH_SIZE: 500,
```
rvagg:
```suggestion
   * Curio also rejects requests (429) when the data set already has 200 or more
   * removals queued on-chain; the queue only drains at the next proving period.
   */
  MAX_DELETE_PIECES_BATCH_SIZE: 200,
```

see https://github.com/filecoin-project/curio/pull/1372, it's slightly more nuanced than this but no need to overcomplicate it here

packages/synapse-sdk/src/storage/context.ts
```diff
@@ -1148,31 +1148,50 @@ export class StorageContext {
...
+   * Delete pieces with the given CIDs or piece IDs from this data set in one transaction.
    *
    * @param options - Options for the delete operation
-   * @param options.piece - The PieceCID identifier or a piece number to delete by pieceID
+   * @param options.pieces - PieceCID identifiers or piece numbers to delete by piece ID
    * @returns Transaction hash of the delete operation
+   *
+   * @remarks
+   * Curio accepts at most 500 pieces per request. PDPVerifier also allows at
+   * most 2,000 cumulative queued removals until the next proving period.
```
rvagg:
```suggestion
   * Curio accepts at most 200 pieces per request and rejects requests (429) when
   * the data set already has 200 or more removals queued on-chain; the queue only
   * drains at the next proving period.
```

also suggesting in these edits to leave out mention of PDPVerifier, it's a theoretical max and in practice it's been shown to not even be usable, so it "allows" but that doesn't work and will brick your data set

packages/synapse-core/src/sp/schedule-piece-deletion.ts
```diff
@@ -18,32 +21,43 @@ export namespace deletePiece {
...
+  // strings so IDs above Number.MAX_SAFE_INTEGER are not rounded by JSON.stringify.
+  const body = `{"extraData":${JSON.stringify(extraData)},"pieceIds":[${pieceIds.join(',')}]}`
+  const response = await request.delete(new URL(`pdp/data-sets/${dataSetId}/pieces/${pieceIds[0]}`, serviceURL), {
+    body,
+    headers: { 'content-type': 'application/json' },
+    timeout: RETRY_CONSTANTS.TIMEOUT,
+    retry: {
+      retries: options.retryCount,
+      minTimeout: options.retryDelay ?? RETRY_CONSTANTS.RETRY_DELAY,
+      shouldRetry: (ctx) => HttpError.is(ctx.error) && ctx.error.code === 429,
```
rvagg: eh, retry on a 429 even though we're going to get that until the next proving period? probably not, we should just bork and report as an error

packages/synapse-core/src/utils/constants.ts
```diff
@@ -99,6 +99,14 @@ export const SIZE_CONSTANTS = {
    */
   MAX_ADD_PIECES_BATCH_SIZE: 40,
 
+  /**
+   * Maximum pieces per schedulePieceDeletions call accepted by the Curio PDP API.
+   *
+   * Curio also rejects requests (429) when the data set already has 200 or more
+   * removals queued on-chain; the queue only drains at the next proving period.
+   */
+  MAX_DELETE_PIECES_BATCH_SIZE: 200,
```
rvagg: I'm not following the discussion or progress on this but last I looked the real limit was more like 70: https://github.com/FilOzone/pdp/issues/283#issuecomment-5065516715
@LexLuthr is 200 correct here?

rvagg (approved):
approval pending confirmation of the right number of max removals

packages/synapse-core/src/utils/constants.ts
ZenGround0: Please move to 35.  We are trying to tighten this down a lot to prevent pausing more datasets before the pdp upgrade.

PR #909, opened 2026-08-07
feat: support custom client for reads

### LINKED ISSUE #895: Split viem clients into read and write clients
There's two very different use cases that we need to consider, Dapps and backend services.

- Dapps need public nodes and should use the [fallback](https://viem.sh/docs/clients/transports/fallback) transport
- Backend services can use authenticated private node and probably also the fallback transport or something like ERPC

We need to split these out and also take into considerations that Dapps use Wagmi and not viem directly.

The issue is how we handle the "clients" in the SDK we should not make reads go through the wallets, we can do that by default and provide a config if the developer wants to configure the "read client" manually.

Considerations
- We need to rethink how the SDK gets the "client" it can be from viem directly or Wagmi
- The SDK should have a clear split between a read client and a write (or wallet) client
- The Core should be specific which type of client it needs for each action
- In the SDK we can turn a wallet client into a read client with good defaults using fallback transport but also provide a config for manual configuration
- For Dapps, Detect if the wallet client transport is a custom EIP-1193 transport and default to a fallback transport for the read client using https://docs.filecoin.io/networks-and-tools/networks/mainnet/rpcs
- For backend service, when the wallet transport is not a custom EIP-1193 we can assume "backend" and keep the same transport for the read client or use default to the public fallback
- Add a new option to configure manually the read client, this can be just the transport and we use the wallet client to complete the config or just accept a full client from the user. 
- We can optimize by enabling multicall by default on any client in the SDK
- We should provide good documentation on how to configure client based on https://github.com/FilOzone/synapse-sdk/pull/873

References:

- https://github.com/FilOzone/synapse-sdk/issues/504
- https://github.com/FilOzone/synapse-sdk/pull/876
- https://github.com/FilOzone/synapse-sdk/pull/873

hugomrdias:
## Summary
- Add `@filoz/synapse-core/client` with `getTransport`, `asClient`, and `toReadClient` (ranked Filecoin RPC transports; normalize clients to `FilecoinChain`; swap custom transports to HTTP for non-JSON-RPC accounts).
- Introduce shared `ReadClient` / `AccountClient` / `SessionKeyClient` aliases and rename the FOC chain interface to `FilecoinChain` (avoids clashing with viem’s `Chain`).
- Keep synapse-core action signatures on plain viem clients (`Client<Transport, Chain>` for reads, `Client<Transport, Chain, Account>` for writes/signing).
- Update synapse-sdk so `Synapse` exposes `client` + `readClient` (both extended with `publicActions`) and threads them into payments / warm-storage / SP registry services.
- Update synapse-react and the CLI to use connector/wallet clients for writes and receipt waits, and drop obsolete account-stripping tests.

## Test plan
- [x] `pnpm --filter @filoz/synapse-core test`
- [x] `pnpm --filter @filoz/synapse-sdk test`
- [x] `pnpm --filter @filoz/synapse-react build`
- [x] Spot-check CLI deposit/list flows against calibration with a wallet client
- [x] Spot-check playground deposit + data-set list with a browser wallet


packages/synapse-core/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+import type { Extended } from './types.ts'
+
+/**
+ * Ranked fallback HTTP transport for Filecoin mainnet.
+ *
+ * Tries the default, Ankr, and DRPC endpoints in latency order.
+ */
+export const mainnetTransport = fallback(
+  [http(mainnet.rpcUrls.default.http[0]), http(mainnet.rpcUrls.ankr.http[0]), http(mainnet.rpcUrls.drpc.http[0])],
+  { rank: true }
```
rjan90, line 14: I think this will cause CLI commands and one-shot SDK scripts using the new defaults to hang after completing.

examples/cli/src/commands/datasets-create.ts
```diff
@@ -31,7 +31,6 @@ export const datasetsCreate: Command = command(
   },
   async (argv) => {
     const { client, chain } = privateKeyClient(argv.flags.chain)
-
```
SgtPooki, line 34: nit: unnecessary line removal?

docs/src/content/docs/developer-guides/synapse-core.mdx
```diff
@@ -47,26 +47,29 @@ All Synapse Core functions accept a viem `Client` as their first argument. Creat
...
 })
 
 // Wallet client for transactions
 const account = privateKeyToAccount("0x...")
 const walletClient = createWalletClient({
   account,
   chain: calibration, // or mainnet
-  transport: http(),
+  transport: getTransport(calibration),
 })
```
SgtPooki, line 67: we should probably have a tracking issue to follow up with `createPublicClient` and `createWalletClient` because we're passing chain twice here (once to the actual method, and once to derive the transport). feels odd.

docs/package.json
```diff
@@ -4,11 +4,43 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "dev": "astro dev",
-    "build": "astro build",
+    "dev": "wireit",
+    "build": "wireit",
```
SgtPooki, line 8: feels like this could have been a separate PR.

why the switch to wireit wrapped astro build commands?

examples/cli/src/commands/upload-dataset.ts
```diff
@@ -31,7 +31,9 @@ export const uploadDataset: Command = command(
 
     const filePath = argv._.path
     const provider = argv._.providerId
-      ? await getPDPProvider(client, { providerId: BigInt(argv._.providerId) })
+      ? await getPDPProvider(client, {
+          providerId: BigInt(argv._.providerId),
+        })
```
SgtPooki, line 36: literally only a style change.. did biome update again?

packages/synapse-core/test/register-provider.test.ts
```diff
@@ -201,13 +197,7 @@ describe('registerProvider', () => {
...
         account,
         chain: calibration,
-        transport: http(undefined, {
-          onFetchRequest(_request, init) {
-            if (typeof init.body === 'string') {
-              requestBodies.push(JSON.parse(init.body) as (typeof requestBodies)[number])
-            }
-          },
-        }),
+        transport: http(),
```
SgtPooki, line 200: it looks like this is removing the network stubbing, I just want to confirm that it's not.

packages/synapse-core/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+export function asClient<
+  account extends Account | undefined = Account | undefined,
+  rpcSchema extends RpcSchema | undefined = undefined,
+  extended extends Extended | undefined = Extended | undefined,
+  chain extends ViemChain | FilecoinChain = ViemChain | FilecoinChain,
+  transport extends Transport = Transport,
+>(client: Client<transport, chain, account, rpcSchema, extended>) {
+  asChain(client.chain)
+  return client as Client<transport, FilecoinChain, account, rpcSchema, extended>
```
SgtPooki, line 83:
```suggestion
export function asClient<
  account extends Account | undefined = Account | undefined,
  rpcSchema extends RpcSchema | undefined = undefined,
  extended extends Extended | undefined = Extended | undefined,
  chain extends ViemChain | FilecoinChain = ViemChain | FilecoinChain,
  transport extends Transport = Transport,
>(client: Client<transport, chain, account, rpcSchema, extended>): Client<transport, FilecoinChain, account, rpcSchema, extended> {
  asChain(client.chain)
  return client
```

why cast and not just set the return type?

packages/synapse-core/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+export function getTransport(chain: ViemChain): Transport {
+  if (chain.id === mainnet.id) {
+    return mainnetTransport
+  }
+
+  if (chain.id === calibration.id) {
+    return calibrationTransport
+  }
+
+  throw new Error(`Unsupported chain: ${chain.id}`)
```
SgtPooki:
```suggestion
 * @throws UnsupportedChainError if the chain is not supported.
 */
export function getTransport(chain: ViemChain): Transport {
  if (chain.id === mainnet.id) {
    return mainnetTransport
  }

  if (chain.id === calibration.id) {
    return calibrationTransport
  }

  throw new UnsupportedChainError(`Unsupported chain: ${chain.id}`)
```

packages/synapse-core/src/chains.ts
```diff
@@ -330,36 +344,26 @@ export function getChain(id?: number): Chain {
...
+
 /**
  * Convert a viem chain to a filecoin chain.
  *
  * @param chain - The viem chain.
  * @returns The filecoin chain.
  * @throws Errors {@link asChain.ErrorType}
  */
-export function asChain(chain: ViemChain): Chain {
+export function asChain(chain: ViemChain): FilecoinChain {
```
SgtPooki, line 363: we should probably normalize how we throw `UnsupportedChainError`

SgtPooki (approved):
lgtm.. a few nits that aren't blocking

packages/synapse-core/test/register-provider.test.ts
hugomrdias: RPC is still mocked; only the account-stripping assertion was removed, matching the rest of the PR.

docs/src/content/docs/developer-guides/synapse-core.mdx
hugomrdias: Opened a follow-up: https://github.com/FilOzone/synapse-sdk/issues/929

Agreed this is odd, but wrapping `createPublicClient` / `createWalletClient` is a new public API and not needed for the client split. Tracking it separately.

packages/synapse-core/src/client.ts
hugomrdias: Done — `getTransport` now throws `UnsupportedChainError` (same constructor as `asChain`, `chain.id` not a custom string). Also added a local `devnetTransport` so the three chains `UnsupportedChainError` lists all have a default transport.

packages/synapse-core/src/chains.ts
hugomrdias: Normalized: `getChain`, `asChain`, and `getTransport` all throw `UnsupportedChainError` and expose `*.ErrorType` for that class.

PR #912, opened 2026-08-10
feat: add support for Curios's confirmedTxHash

snadrus:
This is the minimal compatibility layer for Curio's Replace-by-fee feature:
- pre-existing retries and confirmation works as before

New:
confirmedTxHash will actually be queryable on chain if it exists, otherwise the previous hash will work.

Compat: 
This change will break consumers of txHash. 
Curio disabled Replace-by-fee until we decide it's time to do a rollout due to sufficient Synapse uptake that it's worth breaking the remaining txHash consumers.


hugomrdias:
can you fix the lint errors in ci please

packages/synapse-core/src/piece/internal/hasher.ts
hugomrdias, line 1: can revert this one pls its unrelated to the PR

hugomrdias (changes requested):
LGTM minus the hasher.ts changes

packages/synapse-core/src/piece/internal/hasher.ts
snadrus: done

hugomrdias (approved):

PR #913, opened 2026-08-12
Add pagination and expanded data set, piece, provider, and payment queries

### LINKED ISSUE #753: Normalize pagination interface
We have at least 5 different interfaces across the contracts for paginated methods
Filecoin pay
```
 * @param offset The offset to start from.
 * @param limit Maximum number of entries to return
 * @return results Array of RailInfo structs containing rail IDs and termination status.
 * @return nextOffset The next offset to use for pagination.
 * @return total The total number of rails.
```
https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1676-L1692

FWSS
```
* @param offset Starting index (0-based). Use 0 to start from beginning
* @param limit Maximum number of data set IDs to return. Use 0 to get all remaining
* @return dataSetIds Array of data set IDs
* @dev For large lists, use pagination to avoid gas limit issues. If limit=0,
* returns all remaining data sets starting from offset. Example:
* clientDataSets(service, payer, 0, 100) gets first 100 data set IDs.
```
https://github.com/FilOzone/filecoin-services/blob/3adbb283928eb13c5caa1e007444ee22178d47e1/service_contracts/src/lib/FilecoinWarmStorageServiceStateLibrary.sol#L157-L190

PDP Verifier
```
/**
 * @notice Returns active pieces (non-zero leaf count) for a data set with pagination
 * @dev WARNING: This function has O(offset) gas complexity because it always iterates from
 * piece ID 0. For large data sets with high offsets, consider using getActivePiecesByCursor()
 * @param setId The data set ID
 * @param offset Starting index for pagination (0-based)
 * @param limit Maximum number of pieces to return
 * @return pieces Array of active piece CIDs
 * @return pieceIds Array of corresponding piece IDs
 * @return hasMore True if there are more pieces beyond this page
 */
```
This one, the limit can NOT be 0 to get all
https://github.com/FilOzone/pdp/blob/main/src/PDPVerifier.sol#L383-L400

```
/**
 * @notice Returns active pieces using cursor-based pagination for O(limit) gas complexity
 * @dev This function is more gas-efficient than getActivePieces() for large data sets because
 * it starts iteration from startPieceId instead of always from 0. Use this for paginating
 * through large data sets.
 * @param setId The data set ID
 * @param startPieceId The piece ID to start from (use 0 for first page, then last returned pieceId + 1)
 * @param limit Maximum number of pieces to return
 * @return pieces Array of active piece CIDs
 * @return pieceIds Array of corresponding piece IDs
 * @return hasMore True if there are more pieces beyond this page
 */
```
Uses `startPieceId` not `offset`
https://github.com/FilOzone/pdp/blob/22df1d510002a9ff5dfa812155e84ab993840e72/src/PDPVerifier.sol#L446-L512

```
/**
 * @notice Finds piece IDs matching a given piece CID, with cursor-based pagination
 * @param setId The data set ID
 * @param pieceCid The piece CID to search for
 * @param startPieceId Piece ID to start scanning from (0 for first call)
 * @param limit Maximum number of matching piece IDs to return
 * @return pieceIds Array of matching piece IDs (up to limit)
 */
```
Same as the previous example but without `hasMore` in the output
https://github.com/FilOzone/pdp/blob/22df1d510002a9ff5dfa812155e84ab993840e72/src/PDPVerifier.sol#L514-L546

Maybe we could pick one and go with it?

### LINKED ISSUE #851: `getPdpDataSet` fails on large data sets because `getActivePieceCount` reverts inside  multicall
## Description

`getPdpDataSet()` can fail for large live datasets even when all of its underlying reads are valid. The issue is that `getActivePieceCount(setId)` iterates through every piece ID from `0` to `nextPieceId[setId]`, so the cost of the call grows with the total piece ID range. For large datasets, this subcall can exceed the limits and revert inside `Multicall3`, causing the entire `getPdpDataSet()` call to fail.

This can be reproduced on Filecoin mainnet with `setId = 1273`. All of the underlying reads succeed when called individually, including `getDataSet`, `dataSetLive`, `getDataSetListener`, `getAllDataSetMetadata`, `getPDPProvider`, and `getActivePieceCount`. However, `getPdpDataSet()` fails because it executes all of these reads through a single multicall. In that context, the `getActivePieceCount(setId)` subcall becomes too expensive and reverts, causing the entire `getPdpDataSet()` call to fail.

## Proposed solution

Make `getPdpDataSet()` reliable for large datasets by removing `getActivePieceCount` from the `readPdpDataSetInfo` multicall. `getPdpDataSet()` can fetch `activePieceCount` separately, derive it using `getActivePiecesByCursor` (which is guaranteed to scale to large datasets), or omit the piece count entirely.

Also add documentation recommending `getActivePiecesByCursor` as a fallback for callers that explicitly need a derived active piece count.

## Done criteria

- [ ] getPdpDataSet(setId) succeeds for large live data sets where the current implementation fails

## Notes
- Related issue FilOzone/pdp#265

hugomrdias:
## Summary

- Add reusable pagination utilities and paginated PDP, warm-storage, provider registry, and payment queries
- Return cursor-paginated dataset, piece, provider, and payment results from synapse-core, and update synapse-sdk consumers to traverse those pages explicitly
- Update CLI commands and flags to support the new query capabilities
- Add mocks and coverage for pagination and newly supported queries

## Pagination examples

Read one bounded page and pass the returned cursor back unchanged:

```ts
const page = await getClientDataSets(client, {
  address,
  limit: 100n,
})

if (page.nextCursor !== undefined) {
  const nextPage = await getClientDataSets(client, {
    address,
    cursor: page.nextCursor,
    limit: 100n,
  })
}
```

Iterate over every item with the shared `paginate()` generator:

```ts
import { paginate } from "@filoz/synapse-core"

for await (const dataSet of paginate(({ cursor }) =>
  getClientDataSets(client, { address, cursor })
)) {
  console.log(dataSet.dataSetId)
}
```

Accumulate all items when a complete array is required:

```ts
const dataSets = await Array.fromAsync(
  paginate(({ cursor }) => getClientDataSets(client, { address, cursor }))
)
```

Cursors are opaque continuation values. Each action uses a bounded default when `limit` is omitted, and rejects `limit: 0n`.

## Testing

- Added unit tests covering pagination, data sets, pieces, providers, payment rails, and storage queries
- Not run (not requested)


docs/src/content/docs/developer-guides/migration-guide.md
```diff
@@ -9,6 +9,102 @@ If you are coming from an earlier version of any of the Synapse packages, you wi
...
+```ts
+// before
+const result = await getActivePieces(client, {
+  dataSetId,
+  offset: 0n,
+  limit: 100n,
+})
+
+// after
+const page = await getActivePiecesByCursor(client, {
```
SgtPooki, line 90: the name here feels weird, when we're not passing a cursor. the rename seems unnecessary. can we leave it as `getActivePieces` ?

docs/src/content/docs/developer-guides/migration-guide.md
```diff
@@ -9,6 +9,102 @@ If you are coming from an earlier version of any of the Synapse packages, you wi
...
+for await (const piece of paginate(({ cursor }) =>
+  getActivePiecesByCursor(client, { dataSetId, cursor, limit: 100n })
+)) {
+  console.log(piece.id, piece.cid)
+}
+```
+
+`findPieceIdsByCid`, `getPieces`, and `getPiecesWithMetadata` also return pages and accept `cursor` rather than `startPieceId` or `offset` at the action level.
+
+Raw `*Call` helpers remain ABI-oriented: provide their required contract-facing `offset` or `startPieceId` and `limit` fields explicitly when constructing multicalls.
```
SgtPooki: and these raw call methods are just in synapse-core, yes? I think it would be nice to normalize synapse-sdk being the "polished" surface and synapse-core being the lego bricks.

docs/src/content/docs/developer-guides/synapse-core.mdx
```diff
@@ -68,6 +68,79 @@ const walletClient = createWalletClient({
...
+  const nextPage = await getClientDataSets(publicClient, {
+    address,
+    cursor: page.nextCursor,
+    limit: 25n,
+  })
+  console.log(nextPage.items)
+}
+```
+
+Use `paginate()` when you want to traverse every page. It follows `nextCursor`, yields individual items, and rejects a repeated or non-advancing cursor.
```
SgtPooki: is it possible to break out of a paginate inner loop?

packages/synapse-core/src/warm-storage/get-client-data-sets.ts
```diff
@@ -64,139 +59,47 @@ export namespace getClientDataSets {
...
+      offset: cursor,
+      limit: limit + 1n,
+      contractAddress: options.contractAddress,
+    })
+  )
+  const hasMore = BigInt(data.length) > limit
+  return {
+    items: Array.from(hasMore ? data.slice(0, -1) : data),
+    ...(hasMore ? { nextCursor: cursor + limit } : {}),
   }
```
SgtPooki: this is duplicated in three places, and `findPieceIdsByCid`, `getApprovedProviderIds`, and `getProvidersByProductType` are missing limit+1 boundary testing, which means any individual implementation can drift.

Can we pull this out into a shared helper and then test that shared helper so we are sure each implementation is doing what it's supposed to?

packages/synapse-core/src/sp-registry/get-pdp-providers.ts
```diff
@@ -184,32 +208,15 @@ export async function getApprovedPDPProviders(
...
-  }).providers
+  const [pdpProviders, approvedProviders] = await Promise.all([
+    Array.fromAsync(
+      paginate(({ cursor }) =>
+        getPDPProviders(client, { onlyActive: true, cursor, contractAddress: options.contractAddress })
+      )
+    ),
+    Array.fromAsync(paginate(({ cursor }) => getApprovedProviderIds(client, { cursor }))),
+  ])
+  return pdpProviders.filter((provider) => approvedProviders.includes(provider.id))
```
SgtPooki, line 219: suggestion (non-blocking): `getApprovedPDPProviders` is only exercised indirectly, via fetch-provider-selection-input.test.ts, and only against single-page responses.. the basic preset never returns `limit+1` items or `hasMore: true`, so `paginate()`'s multi-page path never runs against a real action anywhere in synapse-core (pagination.test.ts uses an in-memory stub). 

The old get-client-data-sets.test.ts covered real multi-call traversal (150 items across 100+50 RPC calls) and was deleted with the old API, so multi-page coverage is net lower after this PR.

Can we add a direct test here with a mock returning 2+ pages? That covers `getApprovedPDPProviders` itself (including the approved-ids intersection) and restores the traversal coverage in one shot.

packages/synapse-core/src/mocks/jsonrpc/index.ts
SgtPooki, line 731: this `getProvidersByProductType` mock ignores offset/limit and hardcodes hasMore:false.. we could extend this so we can write tests to help catch any cursor handling regression

SgtPooki (approved):
in general lgtm.. huge change, but everything seems legit.. a few callouts that aren't necessarily blocking but could make some things better.

docs/src/content/docs/developer-guides/migration-guide.md
```diff
@@ -9,6 +9,102 @@ If you are coming from an earlier version of any of the Synapse packages, you wi
...
+  limit: 100n,
+})
+
+// after
+const page = await getActivePiecesByCursor(client, {
+  dataSetId,
+  limit: 100n,
+})
+
+for await (const piece of paginate(({ cursor }) =>
```
rvagg, line 96:
```suggestion
// iterate
for await (const piece of paginate(({ cursor }) =>
```

currently reads like it's part of `// after`

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
```diff
@@ -1,29 +1,53 @@
...
+  parseAllDataSetMetadata,
+} from './get-all-data-set-metadata.ts'
+import { getClientDataSets } from './get-client-data-sets.ts'
+import type { DataSetInfo, PdpDataSet } from './types.ts'
 
-import { type getClientDataSets, getClientDataSetsIterable } from './get-client-data-sets.ts'
-import { readPdpDataSetInfo } from './get-pdp-data-set.ts'
-import type { PdpDataSet } from './types.ts'
+const ENRICHMENT_BATCH_SIZE = 20
+const DATA_SET_CALL_COUNT = 4
```
rvagg, line 21: I think this one deserves a comment, it's going to be very easy to get this out of syncwith the number of calls

rvagg (approved):
well I don't hate it, I think that's what you wanted to hear?

needs a `!` in a commit somewhere to bump major I guess

also utils/sp-tool.js will be broken after this, may as well fix it now, also core-concepts/storage-providers.mdx could do with a look, it has `getApprovedProviderIds` calls

docs/src/content/docs/developer-guides/migration-guide.md
hugomrdias: I’d prefer to keep this name. Synapse Core actions intentionally match contract function names, and the underlying endpoint is getActivePiecesByCursor. It also avoids conflating this endpoint with the older offset-based getActivePieces contract method.

docs/src/content/docs/developer-guides/migration-guide.md
hugomrdias: Yes, these raw ABI-oriented helpers are part of @filoz/synapse-core. The higher-level SDK remains the polished surface and either exposes normalized pages or fully traversed arrays depending on the method. I’ll make that package boundary explicit in the migration guide.

docs/src/content/docs/developer-guides/synapse-core.mdx
hugomrdias: Yes. paginate() is an async generator, so a normal break exits the loop and closes the generator; it won’t fetch another page.

PR #914, opened 2026-08-12
Add cursor-based pagination and safer data set piece detection

hugomrdias:
## Summary

- Replace exact active-piece counts in core data set enrichment with a bounded one-item `getActivePiecesByCursor` check
- Expose `hasActivePieces` on enriched core data sets and use it for provider selection
- Update SDK `hasActivePieces()` to use the same bounded cursor read
- Document the large-data-set risk of `getActivePieceCount` and update affected mocks and tests

## Testing

- Core and SDK type checks
- Core and SDK Node and browser test suites
- Workspace build, including `synapse-react`

### DIFF
--- a/docs/src/content/docs/developer-guides/migration-guide.md
+++ b/docs/src/content/docs/developer-guides/migration-guide.md
@@ -122,9 +122,15 @@ if (dataSet?.hasActivePieces) {
 }
 ```
 
-`WarmStorageService.hasActivePieces()` keeps the same public API, but now uses the same leaf-count proxy instead of calculating an exact count.
+`WarmStorageService.hasActivePieces()` keeps the same public API, but now uses the same leaf-count proxy instead of calculating an exact count. `EnhancedDataSetInfo` from `getClientDataSetsWithDetails()` / `findDataSets()` also exposes `hasActivePieces` instead of `activePieceCount`.
 
-The standalone `getActivePieceCount()` action remains available, but the underlying contract getter scans the data set's piece-ID range and can fail for large data sets. If you need an exact count, explicitly paginate the active pieces:
+The core `getActivePieceCount()` action remains available, but the underlying contract getter scans the data set's piece-ID range and can fail for large data sets. `WarmStorageService.getActivePieceCount()` now paginates `getActivePiecesByCursor` to derive an exact count:
+
+```ts
+const activePieceCount = await warmStorageService.getActivePieceCount({ dataSetId })
+```
+
+To paginate explicitly in core:
 
 ```ts
 let activePieceCount = 0n
@@ -135,8 +141,6 @@ for await (const _piece of paginate(({ cursor }) =>
   activePieceCount++
 }
 ```
-
-This field change is limited to the enriched data set types in `@filoz/synapse-core`; the SDK's legacy `EnhancedDataSetInfo.activePieceCount` field is unchanged.
 
 ---
 
--- a/docs/src/content/docs/developer-guides/migration-guide.md
+++ b/docs/src/content/docs/developer-guides/migration-guide.md
@@ -106,7 +106,7 @@ Raw `*Call` helpers in `@filoz/synapse-core` remain ABI-oriented: provide their 
 
 ### Action: Replace core `activePieceCount` with `hasActivePieces`
 
-The enriched `PdpDataSet` values returned by `getPdpDataSet()` and `getPdpDataSets()` no longer include an exact `activePieceCount`. They now expose `hasActivePieces`, which is derived from a bounded one-item `getActivePiecesByCursor` read:
+The enriched `PdpDataSet` values returned by `getPdpDataSet()` and `getPdpDataSets()` no longer include an exact `activePieceCount`. They now expose `hasActivePieces`, derived from a non-zero `getDataSetLeafCount` read. Leaf count is an O(1) storage lookup, unlike `getActivePiecesByCursor` and `getActivePieceCount`, which scan piece IDs and can run out of gas on large or fully drained data sets:
 
 ```ts
 // before
@@ -122,7 +122,7 @@ if (dataSet?.hasActivePieces) {
 }
 ```
 
-`WarmStorageService.hasActivePieces()` keeps the same public API, but now performs the same bounded cursor read instead of calculating an exact count.
+`WarmStorageService.hasActivePieces()` keeps the same public API, but now uses the same leaf-count proxy instead of calculating an exact count.
 
 The standalone `getActivePieceCount()` action remains available, but the underlying contract getter scans the data set's piece-ID range and can fail for large data sets. If you need an exact count, explicitly paginate the active pieces:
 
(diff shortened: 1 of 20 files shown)


packages/synapse-sdk/src/warm-storage/service.ts
SgtPooki: `getClientDataSetsWithDetails` still reads `getActivePieceCount` per live data set.. same OOG this PR says it's working around.

is another PR addressing this?

SgtPooki, line 296: do we want to add the OOG warning in this JSDoc as well?

packages/synapse-core/src/warm-storage/types.ts
```diff
@@ -44,8 +44,8 @@ export type PdpDataSetInfo = {
   metadata: MetadataObject
   /** PDP provider associated with the data set. */
   provider: PDPProvider
-  /** Number of active (non-zero) pieces in the data set. */
-  activePieceCount: bigint
+  /** Whether the data set contains at least one active piece. */
+  hasActivePieces: boolean
```
SgtPooki, line 48: this is a breaking change, is that going to be handled in a subsequent PR?

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
```diff
@@ -127,7 +134,7 @@ async function enrichDataSetBatch(
     dataSetLiveCall({ chain: client.chain, dataSetId }),
     getDataSetListenerCall({ chain: client.chain, dataSetId }),
     getAllDataSetMetadataCall({ chain: client.chain, dataSetId }),
-    getActivePieceCountCall({ chain: client.chain, dataSetId }),
+    getActivePiecesByCursorCall({ chain: client.chain, dataSetId, startPieceId: 0n, limit: 1n }),
```
SgtPooki: a fully drained data set makes this `limit: 1n` read scan to `nextPieceId` before returning empty.. same worst case as `getActivePieceCount`, and with `allowFailure: false` this can still OOG, yes?

SgtPooki (approved):

packages/synapse-core/src/warm-storage/get-pdp-data-set.ts
```diff
@@ -75,7 +84,7 @@ export async function getPdpDataSet(
 }
 
 /**
- * Read the PDP data set info.
+ * Read PDP data set information without calculating an active-piece count.
```
rvagg:
```suggestion
 * Read PDP data set information.
```

packages/synapse-core/src/warm-storage/location-types.ts
```diff
@@ -2,12 +2,12 @@ import type { PDPProvider } from '../sp-registry/types.ts'
 import type { MetadataObject } from '../utils/metadata.ts'
 
 /**
- * Dataset with piece count, for provider selection.
+ * Dataset with piece-presence information used for provider selection.
  *
  * Picks the fields that selectProviders() and findMatchingDataSets()
- * need, plus activePieceCount which is fetched separately via multicall.
+ * need, plus whether the data set contains an active piece.
```
rvagg:
```suggestion
 * need, plus whether the data set contains at least one active piece.
```

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
rvagg: how about just `getDataSetLeafCountCall()` and use non-zero leaf-count as a proxy here

packages/synapse-sdk/src/warm-storage/service.ts
```diff
@@ -290,13 +290,18 @@ export class WarmStorageService {
...
-   * Uses the contract's dedicated active-piece count rather than paginating pieces.
+   * Reads one cursor-paginated piece rather than calculating the exact active
+   * piece count, which can run out of gas for large data sets.
    * @param options - Options for the data set
    * @param options.dataSetId - The PDPVerifier data set ID
    * @returns True when the data set has at least one active piece
    */
   async hasActivePieces(options: { dataSetId: bigint }): Promise<boolean> {
-    return (await PDPVerifier.getActivePieceCount(this._client, options)) > 0n
+    const page = await PDPVerifier.getActivePiecesByCursor(this._client, {
```
rvagg: same problem as above I think, use `getDataSetLeafCount()` instead?

docs/src/content/docs/developer-guides/migration-guide.md
```diff
@@ -103,6 +103,40 @@ for await (const piece of paginate(({ cursor }) =>
 
 Raw `*Call` helpers remain ABI-oriented: provide their required contract-facing `offset` or `startPieceId` and `limit` fields explicitly when constructing multicalls.
 
+### Action: Replace core `activePieceCount` with `hasActivePieces`
+
+The enriched `PdpDataSet` values returned by `getPdpDataSet()` and `getPdpDataSets()` no longer include an exact `activePieceCount`. They now expose `hasActivePieces`, which is derived from a bounded one-item `getActivePiecesByCursor` read:
```
rvagg: as per conversations below, let's reframe `hasActivePieces` to use `getDataSetLeafCount` non-zero as a more efficient proxy, the scanning to find `nextPieceId` is a bit of a problem, see: https://github.com/FilOzone/pdp/blob/eeb764abde2c52277c0b166654f91ad34ec0a199/src/PDPVerifier.sol#L551-L560

packages/synapse-sdk/src/warm-storage/service.ts
hugomrdias: Yes — added the OOG warning on `getActivePieceCount` and pointed callers at `hasActivePieces` or cursor pagination for an exact count.

packages/synapse-core/src/warm-storage/get-pdp-data-set.ts
hugomrdias: Applied.

packages/synapse-core/src/warm-storage/types.ts
hugomrdias: Handled in this PR: the migration guide covers replacing enriched `activePieceCount` with `hasActivePieces`. We're still pre-1.0, so this ships with the pagination breaking-change batch rather than as a follow-up.

packages/synapse-sdk/src/warm-storage/service.ts
hugomrdias: Changed to hasActivePieces here too

packages/synapse-sdk/src/warm-storage/service.ts
hugomrdias: Done — `hasActivePieces()` now uses `getDataSetLeafCount()`.

packages/synapse-core/src/warm-storage/location-types.ts
hugomrdias: Applied.

packages/synapse-core/src/warm-storage/get-pdp-data-sets.ts
hugomrdias: Yes, drained sets still scanned to `nextPieceId`. Switched enrichment to `getDataSetLeafCountCall()` and treat a non-zero leaf count as `hasActivePieces`.

docs/src/content/docs/developer-guides/migration-guide.md
hugomrdias: Updated the migration notes to describe `hasActivePieces` as a non-zero `getDataSetLeafCount` proxy.

PR #933, opened 2026-08-21
feat(synapse-core): add piece batcher with message-size addPieces limiter

### LINKED ISSUE #922: Stateful batching
We need better primitives to support automatic batching in synapse, improve add pieces UX and cost.

Synapse-core should have a support for upload/pull addpieces batching, with configurable wait time and batch size. Wait time should default to 1 event loop and be configurable in milliseconds and batch size should default to "auto" based on message size limit and be configurable in number of pieces.

High level usege in sdk should have settings to configure batch and disable batching.

Notes:

```text
overhead (message encoding, addresses, msg signature, setid, listener, buffer as well) + 5 slots per piece + [nonce, keys, values, signature] < 64 kib
// overhead should be a constant, piece cids 5*32b, extra data should be abi encode hex > buffer.length 

```

hugomrdias:
## Summary
- Add `createPieceBatcher` to park/pull pieces immediately and coalesce on-chain `addPieces` / `createDataSetAndAddPieces` with a tumbling window.
- Replace the 40-piece count cap with a Filecoin message-size limiter (`addPiecesFits`) and Curio PieceCID size bounds (`MIN_UPLOAD_SIZE` / `MAX_UPLOAD_SIZE`).
- Wire the same message-size check into synapse-sdk `presignForCommit` / `pull` / `commit`. One-shot `upload()` is unchanged.

## How it works

Park and pull run **per piece, immediately**. The tumbling window only batches the on-chain call:

```
upload / pull / enqueue ──► piece on this SP
                              │
                              ▼
                    tumbling addPieces window
                              │
              delay elapsed, limiter overflow,
              flush(), or close()
                              │
                              ▼
         addPieces  or  createDataSetAndAddPieces
```

- **Existing data set:** each flush submits `addPieces` (tx hash + status URL; callers poll if they need confirmation).
- **No data set:** first flush uses `createDataSetAndAddPieces` (waits until the set exists, then caches it). Later windows use `addPieces`.
- **Wait:** default `{ kind: 'delay', ms: 0 }` starts the timer only once the window has a piece and no `upload` / `pull` is still parking. New parking restarts the delay, so concurrent operations coalesce despite size or provider-latency differences. `ms: 0` flushes on the next macrotask after parking settles. `{ kind: 'limiter' }` sits until the next piece does not fit, or `flush` / `close`.
- **Overflow:** if `pending + incoming` does not fit, the current window flushes and the incoming piece starts the next one. A piece that cannot sit in a batch alone is rejected.
- **Pull auth:** `pull` signs extraData for **that one piece** so Curio can `estimateGas`. Flush signs a new extraData for the whole window.
- **Failures:** a failed park/pull never enters the window (siblings continue; retry `upload` / `pull`). A failed flush rejects every slot with `AddPiecesFlushError`; retry with `enqueue`.
- **`onParked`:** runs after the piece is on this SP and before it joins the window. Throw to keep it out of the batch.

## Public interfaces

Import from `@filoz/synapse-core/sp` (and errors from `@filoz/synapse-core/errors`).

```ts
function createPieceBatcher(
  client: Client<Transport, Chain, Account>,
  options: createPieceBatcher.OptionsType
): PieceBatcher

type createPieceBatcher.OptionsType = {
  dataSet: PdpDataSet | undefined
  wait?: PieceBatcherWait              // default { kind: 'delay', ms: 0 }
  limiter?: Limiter                    // default addPiecesFits
  serviceURL?: string                  // required when dataSet is undefined
  payee?: Address                      // required when dataSet is undefined
  payer?: Address
  metadata?: MetadataObject
  cdn?: boolean
}

type PieceBatcherWait =
  | { kind: 'delay'; ms: number }
  | { kind: 'limiter' }

type PieceBatcher = {
  upload: (input: UploadInput) => Promise<PieceResult>
  pull: (input: PullInput) => Promise<PieceResult>
  enqueue: (piece: EnqueuePiece) => Promise<PieceResult>
  flush: () => Promise<FlushResult | undefined>
  close: () => Promise<void>
  readonly pending: readonly EnqueuePiece[]
  readonly dataSet: PdpDataSet | undefined
}

type UploadInput = {
  data: File | Uint8Array | ReadableStream<Uint8Array>
  size?: number                        // for streams; File uses .size
  metadata?: MetadataObject
  pieceCid?: PieceCID
  onParked?: OnParked
  onProgress?: (bytesUploaded: number) => void
  signal?: AbortSignal
}

type PullInput = {
  pieceCid: PieceCID
  sourceUrl: string
  metadata?: MetadataObject
  onParked?: OnParked
}

type EnqueuePiece = { pieceCid: PieceCID; metadata?: MetadataObject }
type FlushResult = { txHash: Hex; statusUrl: string; pieces: EnqueuePiece[] }
type PieceResult = FlushResult & { pieceCid: PieceCID; batchIndex: number }
type OnParked = (piece: EnqueuePiece) => void | Promise<void>
```

Limiter (also used by synapse-sdk `presignForCommit` / `pull` / `commit`):

```ts
function addPiecesFits(options: LimiterOptions): boolean
function assertAddPiecesFit(options: LimiterOptions): void
function assertPieceCidSize(pieceCid: PieceCID): void
function estimateAddPiecesCalldataSize(options: LimiterOptions): number

type Limiter = (options: LimiterOptions) => boolean
type LimiterOptions =
  | { kind: 'addPieces'; dataSet?: PdpDataSet; pieces: LimiterPiece[] }
  | { kind: 'createDataSetAndAddPieces'; metadata?: MetadataObject; cdn?: boolean; pieces: LimiterPiece[] }
```

Budget is `SIZE_CONSTANTS.MAX_ADD_PIECES_MESSAGE_SIZE` (64 KiB minus overhead), estimated from encoded `addPieces` calldata with dummy extraData. PieceCID raw size must be within Curio's `MIN_UPLOAD_SIZE`–`MAX_UPLOAD_SIZE`. `MAX_ADD_PIECES_BATCH_SIZE` (40) remains a fee-preview heuristic only.

## Examples

**Existing data set — coalesced uploads**

```ts
import { createPieceBatcher } from '@filoz/synapse-core/sp'

const batcher = createPieceBatcher(client, { dataSet })

const [a, b] = await Promise.all([
  batcher.upload({ data: fileA }),
  batcher.upload({ data: fileB }),
])
// a.txHash === b.txHash when they landed in the same window

await batcher.close()
```

**Create a data set on first flush**

```ts
const batcher = createPieceBatcher(client, {
  dataSet: undefined,
  serviceURL: provider.pdp.serviceURL,
  payee: provider.serviceProvider,
  payer: client.account.address,
  cdn: false,
})

await batcher.upload({ data: firstFile })
await batcher.close()
// batcher.dataSet is now the created set; later windows use addPieces
```

**Mix upload and SP-to-SP pull**

```ts
await Promise.all([
  batcher.upload({ data: localFile }),
  batcher.pull({
    pieceCid,
    sourceUrl: `https://primary.example/pdp/piece/${pieceCid}`,
  }),
])
await batcher.close()
```

**`onParked`, then retry a failed flush with `enqueue`**

```ts
import { AddPiecesFlushError } from '@filoz/synapse-core/errors'

try {
  await batcher.upload({
    data: file,
    onParked: ({ pieceCid }) => {
      // piece is on this SP; not yet in an addPieces window
    },
  })
} catch (error) {
  if (AddPiecesFlushError.is(error)) {
    await batcher.enqueue({ pieceCid: error.pieceCid, metadata: error.metadata })
  } else {
    throw error
  }
}
```

**Limiter-only wait (tests / explicit flush)**

```ts
const batcher = createPieceBatcher(client, {
  dataSet,
  wait: { kind: 'limiter' },
})

const p1 = batcher.upload({ data: fileA })
const p2 = batcher.upload({ data: fileB })
await batcher.flush() // or close()
await Promise.all([p1, p2])
```

## Test plan
- [x] `pnpm run lint:fix` from `packages/synapse-core`
- [x] `pnpm test` from `packages/synapse-core` (857 passing)
- [x] Confirm `delay: 0` waits for all in-flight parking and coalesces different-latency uploads
- [x] `pnpm test` in `packages/synapse-sdk` (storage message-size tests)
- [x] Confirm `upload()` still accepts `File` / `Uint8Array` / `ReadableStream` and flushes on close
- [x] Confirm a PieceCID below min or above max upload size is rejected before park/pull

### DIFF
--- a/packages/synapse-core/src/sp/create-piece-batcher.ts
+++ b/packages/synapse-core/src/sp/create-piece-batcher.ts
@@ -1,0 +1,460 @@
+import type { Account, Address, Chain, Client, Hex, Transport } from 'viem'
+import { ValidationError } from '../errors/base.ts'
+import { AddPiecesFlushError } from '../errors/pdp.ts'
+import { PullError } from '../errors/pull.ts'
+import { DataSetNotFoundError } from '../errors/warm-storage.ts'
+import type { PieceCID } from '../piece/piece-cid.ts'
+import { signAddPieces } from '../typed-data/sign-add-pieces.ts'
+import { type MetadataObject, pieceMetadataObjectToEntry } from '../utils/metadata.ts'
+import { randU256 } from '../utils/rand.ts'
+import { isUint8Array } from '../utils/streams.ts'
+import { getPdpDataSet } from '../warm-storage/get-pdp-data-set.ts'
+import type { PdpDataSet } from '../warm-storage/types.ts'
+import { addPieces } from './add-pieces.ts'
+import {
+  addPiecesFits,
+  assertPieceCidSize,
+  type Limiter,
+  type LimiterOptions,
+  type LimiterPiece,
+} from './add-pieces-fits.ts'
+import { waitForCreateDataSet } from './create-dataset.ts'
+import { createDataSetAndAddPieces } from './create-dataset-add-pieces.ts'
+import { findPiece } from './find-piece.ts'
+import { waitForPullPieces } from './pull-pieces.ts'
+import { type UploadPieceStreamingData, uploadPieceStreaming } from './upload-streaming.ts'
+
+export type EnqueuePiece = LimiterPiece
+
+export type FlushResult = {
+  txHash: Hex
+  statusUrl: string
+  pieces: EnqueuePiece[]
+}
+
+export type PieceResult = FlushResult & {
+  pieceCid: PieceCID
+  batchIndex: number
+}
+
+/**
+ * When to flush a tumbling addPieces window (limiter overflow, `flush()`, and
+ * `close()` always flush regardless).
+ *
+ * - `delay`: wait `ms` after the first piece in the window (`ms: 0` is one macrotask).
+ * - `limiter`: no timer; sit until the next piece does not fit, or `flush`/`close`.
+ */
+export type PieceBatcherWait = { kind: 'delay'; ms: number } | { kind: 'limiter' }
+
+/**
+ * Called after the piece is on this SP, before it joins the addPieces window.
+ * The callback is awaited, so a thrown error keeps the piece out of the batch
+ * (retry with {@link PieceBatcher.enqueue}). {@link PieceBatcher.close} waits
+ * for it the same way it waits for park/pull I/O.
+ */
+export type OnParked = (piece: EnqueuePiece) => void | Promise<void>
+
+export type UploadInput = {
+  data: File | Uint8Array | ReadableStream<Uint8Array>
+  /** Known length for a stream (`File` uses `.size`). */
+  size?: number
+  metadata?: MetadataObject
+  pieceCid?: PieceCID
+  onParked?: OnParked
+  onProgress?: (bytesUploaded: number) => void
+  signal?: AbortSignal
+}
+
+export type PullInput = {
+  pieceCid: PieceCID
+  sourceUrl: string
+  metadata?: MetadataObject
+  onParked?: OnParked
+}
+
+type Slot = {
+  piece: EnqueuePiece
+  resolve: (result: PieceResult) => void
+  reject: (error: unknown) => void
+}
+
+export type PieceBatcher = {
+  /** Stream onto this SP, then join the addPieces window. */
+  upload: (input: UploadInput) => Promise<PieceResult>
+  /** Pull onto this SP (own extraData), then join the addPieces window. */
+  pull: (input: PullInput) => Promise<PieceResult>
+  /** Already on this SP. Join the addPieces window only. */
+  enqueue: (piece: EnqueuePiece) => Promise<PieceResult>
+  flush: () => Promise<FlushResult | undefined>
+  close: () => Promise<void>
+  readonly pending: readonly EnqueuePiece[]
+  readonly dataSet: PdpDataSet | undefined
+}
+
+export namespace createPieceBatcher {
+  export type OptionsType = {
+    dataSet: PdpDataSet | undefined
+    /** When to flush a window. Defaults to `{ kind: 'delay', ms: 0 }`. */
+    wait?: PieceBatcherWait
+    /** Defaults to {@link addPiecesFits}. */
+    limiter?: Limiter
+    /** Required when `dataSet` is undefined. */
+    serviceURL?: string
+    /** Required when `dataSet` is undefined. */
+    payee?: Address
+    payer?: Address
+    metadata?: MetadataObject
+    cdn?: boolean
+  }
+  export type ReturnType = PieceBatcher
+}
+
+/**
+ * Create a stateful piece batcher that parks/pulls immediately and coalesces addPieces.
+ *
+ * `upload` and `pull` run per-piece I/O right away (`upload` uses the
+ * streaming CommP-last protocol for bytes and streams). The tumbling window
+ * only batches the on-chain addPieces (or createDataSetAndAddPieces) call.
+ * Pull authorization uses `signAddPieces` for that one piece; flush signs a
+ * new extraData for the whole window.
+ *
+ * @param client - Wallet client used to sign and submit.
+ * @param options - {@link createPieceBatcher.OptionsType}
+ * @returns Batcher {@link createPieceBatcher.ReturnType}
+ *
+ * @example
+ * ```ts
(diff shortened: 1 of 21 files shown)


packages/synapse-core/src/sp/create-piece-batcher.ts
```diff
@@ -0,0 +1,480 @@
...
+  }
+
+  function limiterOptions(pieces: LimiterPiece[]): LimiterOptions {
+    if (dataSet != null) {
+      return { kind: 'addPieces', dataSet, pieces }
+    }
+    return { kind: 'createDataSetAndAddPieces', metadata: datasetMetadata, cdn, pieces }
+  }
+
+  function fits(pieces: LimiterPiece[]): boolean {
```
Kubuxu, line 194: I'm assuming for legacy datasets we are going to limit here.

Kubuxu (approved):
SGTM, although the TS async flows are not my strong suit

packages/synapse-core/src/sp/create-piece-batcher.ts
```diff
@@ -0,0 +1,480 @@
...
+}
+
+export type PieceBatcher = {
+  /** Stream onto this SP, then join the addPieces window. */
+  upload: (input: UploadInput) => Promise<PieceResult>
+  /** Pull onto this SP (own extraData), then join the addPieces window. */
+  pull: (input: PullInput) => Promise<PieceResult>
+  /** Already on this SP. Join the addPieces window only. */
+  enqueue: (piece: EnqueuePiece) => Promise<PieceResult>
+  flush: () => Promise<FlushResult | undefined>
```
BravoNatalie, line 102: Could you add comments for `flush` and `close` to make their expected behavior a clearer?

The current comments say that both trigger a flush and that `close()` waits for parking, but they don't make it clear that `flush()` only submits pieces that are already in `windowSlots`.

packages/synapse-core/src/sp/create-piece-batcher.ts
```diff
@@ -0,0 +1,480 @@
...
+  }
+
+  function requirePayee(): Address {
+    if (payee == null) {
+      throw new ValidationError('`payee` is required when dataSet is undefined.')
+    }
+    return payee
+  }
+
+  async function flush(): Promise<FlushResult | undefined> {
```
BravoNatalie, line 510: Should `flush()` mean “drain all work already accepted and report whether submission succeeded,” or only “submit the current internal window”?

The fush API does not let this clear for the user, but it currently flushes only pieces already in `windowSlots`. If an `upload` or `pull` is still parking, `flush()` can return `undefined` before that piece
 reaches the window. 

The examples in the PR description make `flush()` and `close()` seem interchangeable. Could you make the difference between them a bit clearer?

packages/synapse-core/src/sp/create-piece-batcher.ts
```diff
@@ -0,0 +1,480 @@
...
+        slot.reject(
+          new AddPiecesFlushError({
+            pieceCid: slot.piece.pieceCid,
+            metadata: slot.piece.metadata,
+            pieces,
+            cause,
+          })
+        )
+      }
+      return undefined
```
BravoNatalie, line 304: An empty window and a failure both make flush resolve with `undefined`. Should we distinguish between the two cases?

packages/synapse-core/src/sp/create-piece-batcher.ts
hugomrdias: yes

packages/synapse-core/src/sp/create-piece-batcher.ts
hugomrdias: the logic changed a little in https://github.com/FilOzone/synapse-sdk/pull/935 they almost the same in the changes in #935 `close` calls flush and makes the batcher non-reusable

packages/synapse-core/src/sp/create-piece-batcher.ts
hugomrdias: do you have a use case ? errors should reject with the related promises (upload, pull, enqueue) not in the flush

BravoNatalie: ah ok, that makes sense! if flush also rejects the caller would have the same failure twice.

BravoNatalie (approved):

PR #935, opened 2026-08-24
feat(synapse-sdk): batch storage uploads by default

hugomrdias:
## Summary

- add a core piece batcher that groups `addPieces` work while respecting transaction message-size limits
- enable batching by default for `StorageManager.upload()` and `StorageContext.upload()`
- preserve the SDK's default two-copy flow: upload to a primary provider, pull to secondary providers, then batch each provider's on-chain commit independently
- expose delay and limiter wait strategies through `Synapse`, plus `synapse.storage.flush()` for explicitly releasing limiter-held batches
- allow callers to opt out with `pieceBatching: false`

## Usage

Batching is enabled by default, using a zero-delay window:

```ts
const synapse = await Synapse.create({
  privateKey,
  rpcUrl,
})
```

The delay can be configured when creating the client:

```ts
const synapse = await Synapse.create({
  privateKey,
  rpcUrl,
  pieceBatching: {
    wait: { kind: 'delay', ms: 100 },
  },
})
```

The limiter strategy holds a batch until it reaches the transaction size boundary or the application flushes it:

```ts
const synapse = await Synapse.create({
  privateKey,
  rpcUrl,
  pieceBatching: {
    wait: { kind: 'limiter' },
  },
})

const uploads = files.map((file) => synapse.storage.upload(file))
await synapse.storage.flush()
const results = await Promise.all(uploads)
```

To retain the previous immediate, per-upload behavior:

```ts
const synapse = await Synapse.create({
  privateKey,
  rpcUrl,
  pieceBatching: false,
})
```

## How it works

1. `StorageManager.upload()` resolves the primary and secondary contexts exactly as before, including the default request for at least two copies.
2. The primary upload is parked at its provider. Once parked, each secondary starts its provider-to-provider pull from the primary.
3. A Synapse-scoped coordinator assigns the upload or pull to a batcher keyed by provider and data set. Providers therefore submit separate batches, while compatible pieces for the same data set share one transaction.
4. The core batcher waits according to the configured delay or limiter and automatically splits batches that would exceed the encoded `addPieces` message-size limit.
5. All pieces in a submitted batch share the confirmation request. Each upload resolves to its own piece ID using its batch index.
6. Newly created data-set IDs are synchronized back into their storage contexts so later uploads reuse the established data set.

The existing result and failure semantics remain intact: successful copies are returned in `copies`, intermediate provider failures remain in `failedAttempts`, and the manager still throws `StoreError` or `CommitError` when the corresponding operation cannot succeed.

The split `store()`, `pull()`, and `commit()` APIs remain unchanged and unbatched. `flush()` waits for uploads and pulls to finish parking before releasing all current windows; the core batcher's `close()` prevents new work and delegates to that same flush behavior.

## Compatibility

- batching can be disabled with `pieceBatching: false`
- the public `PdpDataSet` contract remains unchanged
- the core custom limiter contract still receives the data set and pending pieces
- session-key signing uses the delegated client while preserving the root payer

## Tests

- `packages/synapse-core` build
- `packages/synapse-sdk` build
- package lint
- core Node tests: 858 passing
- core browser tests: 858 passing
- SDK Node tests: 280 passing, 7 pending
- SDK browser tests: 280 passing, 7 pending

### DIFF
--- a/packages/synapse-sdk/src/storage/manager.ts
+++ b/packages/synapse-sdk/src/storage/manager.ts
@@ -70,6 +70,12 @@ import type {
 import { combineMetadata, createError, SIZE_CONSTANTS, TIME_CONSTANTS } from '../utils/index.ts'
 import type { WarmStorageService } from '../warm-storage/index.ts'
 import { StorageContext } from './context.ts'
+import {
+  type BatchedCommitResult,
+  type BatchingHold,
+  getPieceBatchingService,
+  type PieceBatchingService,
+} from './piece-batching.ts'
 import { terminateServiceFlow } from './terminate.ts'
 
 // Multi-copy upload constants
@@ -204,6 +210,11 @@ export class StorageManager {
    * @throws CommitError if all commit attempts fail (data stored but not on-chain)
    */
   async upload(data: UploadPieceStreamingData, options?: StorageManagerUploadOptions): Promise<UploadResult> {
+    const batching = getPieceBatchingService(this._synapse)
+    if (batching != null) {
+      return this._uploadBatched(data, options, batching, batching.hold())
+    }
+
     const { contexts, explicitProviders } = await this._resolveUploadContexts(options)
     const [primary, ...secondaries] = contexts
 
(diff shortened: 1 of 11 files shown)


packages/synapse-sdk/src/storage/piece-batching.ts
```diff
@@ -0,0 +1,251 @@
...
+    void ready.promise.finally(() => {
+      this.pendingReady.delete(ready.promise)
+    })
+  }
+
+  private entry(context: StorageContext): Promise<Entry> {
+    const key = this.key(context)
+    const existing = this.entries.get(key)
+    if (existing != null) {
+      void existing.then((entry) => {
```
BravoNatalie: two concurrent uploads sharing an entry can produce an unhandled rejection if `createEntry()` fails, even
though the original entry and upload promises have their own catches.

Returning `existing.then(...)` and returning `entry` from the callback would keep the synchronization and its rejection within the upload’s handled promise chain.

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -381,6 +392,251 @@ export class StorageManager {
     }
   }
 
+  /** Flush all piece windows currently accepted by this Synapse instance. */
```
BravoNatalie: Could you add more details to the jsdoc comment clarify what `flush()` means from the caller's perspective?

On a first read, it's easy to assume that once `await storage.flush()` resolves, all batches were submitted successfully. But submission errors are actually reported through the individual upload promises, and `flush()` can still resolve when a batch fails.

packages/synapse-sdk/src/storage/manager.ts
BravoNatalie: Could we update this jsdoc and the upload pipeline guide (docs/) to explain the new concurrent `upload()` behavior?

It would also be helpful to mention that sequential `await upload(...)` calls won't be grouped into the same batch. An example could make the difference clearer:

```ts
// Concurrent uploads can be grouped into the same batch (default batching kind)
const results = await Promise.all(
  files.map((file) => synapse.storage.upload(file))
)
```

vs.

```ts
// Sequential uploads are handled one at a time (default batching kind)
for (const file of files) {
  await synapse.storage.upload(file)
}
```

packages/synapse-sdk/src/storage/piece-batching.ts
hugomrdias: Good catch. Updated `entry()` to return the `existing.then(...)` synchronization chain (and return the entry from the callback), so a shared `createEntry()` rejection stays in each upload's handled promise chain. I also added a concurrent-entry-failure regression test that runs in Node and Chromium.

packages/synapse-sdk/src/storage/manager.ts
hugomrdias: Updated the JSDoc to clarify the caller contract: `flush()` waits for accepted uploads/pulls to finish parking and submits pending windows, but upload submission/confirmation failures are surfaced through the individual upload promises. It also now states that callers must await those promises and that flushing is a no-op when batching is disabled.

packages/synapse-sdk/src/storage/manager.ts
hugomrdias: Updated both the `upload()` JSDoc and the upload pipeline guide. They now show concurrent versus sequential calls, explain that only compatible concurrent uploads can share a batch, document limiter-mode `flush()`, and retain split operations as the manual-control path.

packages/synapse-sdk/src/test/storage-upload.test.ts
```diff
@@ -182,6 +182,129 @@ describe('Storage Upload', () => {
     )
   })
 
+  it('should reject concurrent uploads cleanly when shared batch entry creation fails', async () => {
```
BravoNatalie: Before the fix, both promises were already rejecting with `expected`. The issue was the extra unhandled rejection from `void existing.then(...)`.

Could we also check that no `unhandledrejection` happens? Maybe with a Node subprocess using `--unhandled-rejections=strict`? Since Chromium only logs page errors without failing the test, I think reverting the fix could still leave this test green.

BravoNatalie (approved):
Things look good to me! The only thing my agent found was some testing hardening, which I left as a comment.

packages/synapse-sdk/src/test/storage-upload.test.ts
hugomrdias: Agreed. I removed this test because its assertions covered the existing upload rejection behavior rather than the detached unhandled rejection itself. The production fix remains in place; the surrounding storage upload tests continue to cover the current public batching behavior.

BravoNatalie (approved):
👍

PR #936, opened 2026-08-24
feat!: remove piece metadata getters

hugomrdias:
## Summary

- Remove FWSS piece metadata getter APIs and related types
- Update SDK, core, React hooks, CLI, mocks, and generated ABI handling
- Document reading piece metadata from `PieceAdded` events or an indexer
- Update the Synapse React example for pieces without on-chain metadata

## Follow-up required

- [x] Once the new `filecoin-services` commit is available, update `FILECOIN_SERVICES_GIT_REF` in `packages/synapse-core/wagmi.config.ts` and run `pnpm --dir packages/synapse-core run generate-abi`.

## Testing

- `pnpm test`
- `pnpm run build`
- `pnpm --dir examples/cli run lint`
- `pnpm run check:markdown`
- `pnpm --dir docs exec astro build`

PR #938, opened 2026-08-25
fix: deprecate piece metadata getters

Kubuxu:
## Summary

- mark the public FWSS piece metadata getter APIs and related types as deprecated before their removal in #936
- direct consumers to `PieceAdded` events or an indexer while keeping metadata writes supported
- add matching warnings to the storage and React documentation

## Testing

- `pnpm exec playwright-test "test/get-all-piece-metadata.test.ts" "test/get-pieces.test.ts" --mode node` (packages/synapse-core)
- `pnpm exec playwright-test "src/test/warm-storage-metadata.test.ts" --mode node` (packages/synapse-sdk)
- `pnpm run build`
- `pnpm run check:markdown`
- `pnpm exec astro build` (docs)

### DIFF
--- a/packages/synapse-core/tsconfig.json
+++ b/packages/synapse-core/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "./dist",
+    "removeComments": false,
     "types": ["mocha", "node"]
   },
   "include": ["src", "test"],
--- a/packages/synapse-core/tsconfig.json
+++ b/packages/synapse-core/tsconfig.json
@@ -2,7 +2,6 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "removeComments": false,
     "types": ["mocha", "node"]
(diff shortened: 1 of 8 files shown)


packages/synapse-sdk/tsconfig.json
hugomrdias, line 1: revert this, this option removes comments from the compiled js code not the types, types always keep the comments and that what we need

packages/synapse-core/tsconfig.json
hugomrdias, line 1: revert this, this option removes comments from the compiled js code not the types, types always keep the comments and thats what we need

hugomrdias (changes requested):

packages/synapse-sdk/tsconfig.json
Kubuxu: Reverted in 666a691.

packages/synapse-core/tsconfig.json
Kubuxu: Reverted in 666a691.

hugomrdias (approved):

PR #946, opened 2026-08-27
feat!: correct storage cost and fee calculations

### LINKED ISSUE #859: `getDataSetSizes` should adjust for fr32 expansion and clarify that it's an approximation
Same as https://github.com/FilOzone/filecoin-services/issues/451; we overestimate the actual size of the data by nearly 0.8% by taking leaf count as it is. Leaf count operates over the virtual fr32 expanded data which is not stored on disk and not charged. We just need to do `leafCountToRawSize` as per https://github.com/FilOzone/pdp/pull/266 in here. But we also should document in the API that we're overestimating by _up to 32 bytes_. This is the value we use to charge for the data set, it's just not the precise value of how much the client is storing. An accurate amount would be obtained by iterating over piece CIDs and decoding raw size from the CID but that's costly. It could be documented at least as an accurate alternative for those that want it.

### LINKED ISSUE #899: Move internal calculateMultiContextCosts logic into stateless synapse-core method(s)
Seeing prepare & costs used in practice here: https://github.com/FilOzone/foc-devnet/pull/153 where the `context` is only made just to be able to call this nice method that gets all the pricing right, and then it goes on to use synapse-core directly. Currently we hide a lot of fancy logic behind `StorageManager#calculateMultiContextCosts` which requires contexts _but_ it only uses those for 3 things (each) which could become some variation of arguments: (1) whether or not there's a createDataSet needed, (2) whether CDN is enabled or not, and a bit more complicated (3) how large the data set already is. But this could just as easily be a `(dataSetId: bigint | undefined, withCDN: bool)[]` argument that could do the same thing without needing the SDK or contexts. Possibly the caller could be made responsible for the existing size and it becomes `(existingSize: number | undefined, withCDN: bool)[]` but it is nice to do the parallel size lookup inline with the other 3 calls it does up front.

### LINKED ISSUE #944: Update to pricing changes in FWSS 1.3.0
1.3.0 introduced pricing changes that weren't integrated into Synapse.
https://github.com/FilOzone/filecoin-services/releases/v1.3.0#:~:text=Action%20Required%20For%20Integrators

hugomrdias:
## Summary

- Audit and correct warm-storage cost, deposit, lockup, reserve, and upload-fee calculations
- Add PDP dataset size and leaf-count helpers plus related exports and mocks
- Update multi-context storage cost handling and refresh storage documentation
- Expand unit test coverage across core calculations and storage services

## Testing

- Added and updated unit tests for cost calculations, dataset sizing, PDP helpers, and storage workflows


packages/synapse-core/src/warm-storage/calculate-additional-lockup-required.ts
```diff
@@ -37,27 +39,37 @@ export namespace calculateAdditionalLockupRequired {
...
  * Existing datasets pay only the incremental rate lockup. New datasets also
- * include lifecycle and optional CDN/cache-miss lockups.
+ * include lifecycle and optional CDN/cache-miss lockups. Storage rates are
+ * calculated from aggregate PDP leaf counts, matching FWSS's rounding order.
  *
  * @param params - {@link calculateAdditionalLockupRequired.ParamsType}
  * @returns {@link calculateAdditionalLockupRequired.OutputType}
+ * @throws {@link ValidationError} when the leaf count or piece sizes are invalid
  */
 export function calculateAdditionalLockupRequired(
```
BravoNatalie, line 52: Existing datasets can still be paying the old `0.024/month `proving rate until their next rate-touching operation (e.g. AddPieces). 

Recomputing both rates with `0.12` makes the proving fee cancel, so `prepare() `can miss roughly `0.096 USDFC` of 30-day lockup. 

Maybe the resolved dataset state should use the current PDP rail rate from `getRail(pdpRailId)`, then we can compare the new target rate against that actual value.

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -802,59 +745,40 @@ export class StorageManager {
...
-        })
+    const resolvedContexts = contexts.map((context): calculateUploadCosts.ContextType => {
+      if (context.dataSetId == null) {
+        return { pieceSizes: options.pieceSizes, withCDN: context.withCDN, dataSet: null }
+      }
 
-    const clamped = rawDepositNeeded > 0n ? rawDepositNeeded : 0n
-    const depositNeeded = clamped + buffer
-    const needsFwssMaxApproval = !approved
+      const dataSetState = dataSetsById.get(context.dataSetId)
```
BravoNatalie, line 1027: Should we reject the context here when` dataSetState.pdpEndEpoch !== 0n`? 

A terminated dataset can still be live in PDP during the wind-down period, so the normal `validateDataSet()` checks pass, but FWSS will always reject `piecesAdded()` with `DataSetPaymentAlreadyTerminated`. 

Without this check, `prepare() `can ask the user to fund an upload that cannot be committed.

packages/synapse-sdk/src/storage/manager.ts
BravoNatalie: This is an old detail I ended up stumbling on when I first started using synapse-sdk.

The JSDoc mentions that `prepare()` can accept precomputed `costs` from a previous `getUploadCosts()` call, but `StorageManager.getUploadCosts()` always calculates costs for a single storage context. Meanwhile, `prepare()` without an explicit context normally prepares the default two copies.

Since passing `costs` skips context resolution, someone following this example could fund one rail and then have the default upload try to create two.

 Could we require the matching contexts with precomputed costs, include enough context information to validate them, or avoid recommending `getUploadCosts()` output in this example?

packages/synapse-sdk/src/types.ts
```diff
@@ -47,8 +47,8 @@ export type FundOptions = import('@filoz/synapse-core/pay').fund.OptionsType
 export interface PrepareOptions {
   /** StorageContext(s) to prepare for upload. */
   context?: import('./storage/context.ts').StorageContext | import('./storage/context.ts').StorageContext[]
-  /** Size of new data to upload, in bytes. */
-  dataSize: bigint
+  /** Exact raw payload size of every piece committed per storage context, in bytes. */
+  pieceSizes: readonly bigint[]
```
BravoNatalie, line 51: Just flagging this to make sure we're releasing the package as a breaking change

docs/src/content/docs/developer-guides/storage/storage-costs.mdx
BravoNatalie: Shouldn't this say “Total monthly rate.” ?
`rates.perMonth` is summed across all contexts, so with the three contexts above this prints the combined rate for all three copies, not the rate for one copy.

docs/src/content/docs/developer-guides/storage/storage-costs.mdx
```diff
@@ -63,20 +63,20 @@ These examples use plain arithmetic for clarity. For programmatic cost calculati
...
 ```
 
 | Duration | Per copy |
 | --------- | ---------- |
-| 1 month | ≈ 0.024 USDFC |
-| 24 months | ≈ 0.58 USDFC |
+| 1 month | ≈ 0.120 USDFC |
+| 24 months | ≈ 2.88 USDFC |
 
 The default 2-copy configuration roughly doubles the recurring cost. One-time fees apply once: $0.025 to create the data set plus the per-batch add-pieces fee.
```
BravoNatalie: Could we clarify that these fees apply once per dataset/copy?

packages/synapse-core/test/get-account-total-storage-size.test.ts
```diff
@@ -88,8 +88,8 @@ describe('getAccountTotalStorageSize', () => {
       address: ADDRESSES.client1,
     })
 
-    // Only datasets 1 and 3 are live: (100 + 300) * 32
-    assert.equal(result.totalSizeBytes, (100n + 300n) * SIZE_CONSTANTS.BYTES_PER_LEAF)
+    // Only datasets 1 and 3 are live, converted from padded leaves to raw bytes.
+    assert.equal(result.totalSizeBytes, leafCountToRawSize(100n) + leafCountToRawSize(300n))
```
BravoNatalie, line 95: Real PDP does not permit that combination.

`getDataSetLeafCount()` reverts when` dataSetLive() ` is `false`, and production uses `allowFailure: false`, so dataset 2 would reject the whole multicall instead of being skipped

BravoNatalie (changes requested):

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -648,12 +646,15 @@ export class StorageManager {
...
-   * - Fetches each existing dataset's size from chain
+   * - Fetches each existing dataset's aggregate PDP leaf count from chain
+   * - Fetches each existing dataset's lifecycle reserve and pending fees
    * - Sums lockup across all contexts
    * - Computes debt, runway, and buffer once at the account level
+   * - Prices the known `pieceSizes` committed to each context
    *
    * @param options - {@link PrepareOptions}
    * @returns {@link PrepareResult} with costs and an optional transaction
+   * @throws When `pieceSizes` is empty or contains a non-positive size
```
BravoNatalie, line 930: Should `prepare()` validate the same execution limits as `store()` and `commit()`? 
Right now 41 piece sizes are accepted and priced as one add-pieces call, but `commit()` rejects anything above 40 at the moment.

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -802,59 +745,40 @@ export class StorageManager {
...
+      if (context.dataSetId == null) {
+        return { pieceSizes: options.pieceSizes, withCDN: context.withCDN, dataSet: null }
+      }
 
-    const clamped = rawDepositNeeded > 0n ? rawDepositNeeded : 0n
-    const depositNeeded = clamped + buffer
-    const needsFwssMaxApproval = !approved
+      const dataSetState = dataSetsById.get(context.dataSetId)
+      if (dataSetState == null) {
+        throw new Error(`Data set ${context.dataSetId} does not exist`)
```
BravoNatalie: There's an existing error type for this: `DataSetNotFoundError`

packages/synapse-core/src/warm-storage/calculate-additional-lockup-required.ts
hugomrdias: i decided not to account for old datasets as that would blow up the complexity, each dataset will get updated as they are touched by operations.
seem acceptable given the current usage vs the complexity it would add.

packages/synapse-sdk/src/storage/manager.ts
hugomrdias: the batching PRs changed this theres no more 40 limit

packages/synapse-core/src/warm-storage/get-upload-costs.ts
```diff
@@ -19,65 +17,66 @@ export namespace getUploadCosts {
...
-  const pieceCount = options.pieceCount ?? 1n
-  const extraRunwayEpochs = options.extraRunwayEpochs ?? DEFAULT_RUNWAY_EPOCHS
-  const bufferEpochs = options.bufferEpochs ?? DEFAULT_BUFFER_EPOCHS
+
+  let dataSet: calculateUploadCosts.ExistingDataSetType | null = null
+  if (!isNewDataSet) {
+    const leafCount = options.dataSetLeafCount
+    const lifecycleReserveBalance = options.currentLifecycleReserveBalance
+    if (leafCount == null) {
+      throw new ValidationError('dataSetLeafCount is required for an existing data set')
```
SgtPooki: suggestion: validate pdpEndEpoch here too. calculateMultiContextCosts throws ServiceAlreadyTerminatedError on pdpEndEpoch != 0, but this entry point and calculateDepositNeeded accept terminated state and simulate a replenishment Rails.sol never performs on terminated rails, so a caller can be told to fund an upload FWSS will reject with DataSetPaymentAlreadyTerminated. The single-context example in storage-costs.mdx reads getDataSet without checking pdpEndEpoch, so it hits this path. Adding pdpEndEpoch (or a terminated flag) to the required existing-data-set inputs closes all three.

packages/synapse-core/src/warm-storage/calculate-deposit-needed.ts
```diff
@@ -103,63 +42,78 @@ export namespace calculateDepositNeeded {
...
     lockupEpochs: params.lockupEpochs,
     isNewDataSet: params.isNewDataSet,
     withCDN: params.withCDN,
   })
-  const fees = calculateUploadFees({
+  let dataSet: calculateUploadCosts.ExistingDataSetType | null = null
+  if (!params.isNewDataSet) {
+    const lifecycleReserveBalance = params.currentLifecycleReserveBalance
+    if (lifecycleReserveBalance == null) {
+      throw new ValidationError('currentLifecycleReserveBalance is required for an existing data set')
```
SgtPooki: non-blocking: this existing-data-set validation block is nearly verbatim the one in get-upload-costs.ts. One shared resolver keeps the messages and defaults from drifting. Related: this function calls calculateAdditionalLockupRequired only to rebuild a breakdown calculateUploadCosts already computes internally. Fine if this helper is headed for deprecation, otherwise expose the per-context breakdown from the utility.

packages/synapse-core/src/utils/calculate-upload-costs.ts
```diff
@@ -0,0 +1,292 @@
+import { ValidationError } from '../errors/base.ts'
+import { calculateAdditionalLockupRequired } from '../warm-storage/calculate-additional-lockup-required.ts'
+import { calculateEffectiveRate } from '../warm-storage/calculate-effective-rate.ts'
+import { calculateLifecycleReserveFunding } from '../warm-storage/calculate-lifecycle-reserve-funding.ts'
+import { calculateUploadFees } from '../warm-storage/calculate-upload-fees.ts'
+import type { getPriceList } from '../warm-storage/price-list.ts'
```
SgtPooki: nit: utils/ now imports five things from warm-storage/ while warm-storage/ imports back from utils/ (pdp-size, constants). utils/ was leaf-level before this. AGENTS.md lists storage costs under warm-storage, so warm-storage/calculate-upload-costs.ts keeps the dependency direction one-way.

packages/synapse-sdk/src/types.ts
```diff
@@ -47,13 +47,17 @@ export type FundOptions = import('@filoz/synapse-core/pay').fund.OptionsType
 export interface PrepareOptions {
   /** StorageContext(s) to prepare for upload. */
   context?: import('./storage/context.ts').StorageContext | import('./storage/context.ts').StorageContext[]
-  /** Size of new data to upload, in bytes. */
-  dataSize: bigint
+  /** Exact raw payload size of every piece planned per context; multi-piece fee estimates are conservative. */
+  pieceSizes: readonly bigint[]
```
SgtPooki: non-blocking: "planned per context" reads like pieces can differ per context, but calculateMultiContextCosts applies the same array to every context. Something like "raw size of every piece in the upload; each context stores the same pieces" avoids that reading.

packages/synapse-core/src/warm-storage/get-upload-costs.ts
SgtPooki: Could we mirror the pdpEndEpoch check from calculateMultiContextCosts here? Without it, this and calculateDepositNeeded quote deposits for uploads that will revert with DataSetPaymentAlreadyTerminated.. the storage-costs.mdx single-context example doesn't check it either.

packages/synapse-core/src/warm-storage/calculate-deposit-needed.ts
SgtPooki: non-blocking: this existing-data-set validation block is nearly verbatim the one in get-upload-costs.ts. One shared resolver keeps the messages and defaults from drifting. Related: this function calls calculateAdditionalLockupRequired only to rebuild a breakdown calculateUploadCosts already computes internally. Fine if this helper is headed for deprecation, otherwise expose the per-context breakdown from the utility.

packages/synapse-sdk/src/storage/manager.ts
```diff
@@ -1071,59 +1019,43 @@ export class StorageManager {
...
 
-    const rates = {
-      perEpoch: totalRatePerEpoch,
-      perMonth: totalRatePerMonth,
-    }
+      return {
+        pieceSizes: options.pieceSizes,
+        withCDN: context.withCDN,
+        dataSet: {
+          leafCount: leafCounts.get(context.dataSetId) ?? 0n,
```
SgtPooki, line 1035: question: should a missing leaf count throw instead of defaulting to 0n? getDataSetLeafCounts was called with exactly these IDs, so a miss here is a bug, and 0n underprices the rate delta for an existing data set. A missing data set two lines up throws DataSetNotFoundError.

SgtPooki (commented):
I checked the math against filecoin-services main.. the fee/lockup constants match PriceListUSDFC.sol, the reserve replenishment logic matches Rails.sol replenishReserveIfNeeded, rawSizeToLeafCount is the inverse of Cids.sol, and calculateEffectiveRate matches calculateStorageRate (including truncation order). Removing the fees from the deposit also fixes a double count we had before.

Natalie's comments should be addressed before this merges.

One small thing: getAccountTotalStorageSize does two multicalls per page now instead of one. It's correct (leaf count reverts on non-live sets with allowFailure: false), just an extra roundtrip in case you'd rather keep the single call and use allowFailure: true.

packages/synapse-sdk/src/storage/manager.ts
SgtPooki: looked closer.. getDataSetLeafCounts returns an entry for every requested id (leaf count, or 0n when the data set isn't live), so the ?? 0n never actually fires, it's just narrowing Map.get's bigint | undefined for TS. ignore this one.

SgtPooki (approved):

PR #960, opened 2026-09-10
fix(storage): prefer compact data sets over legacy ones

### LINKED ISSUE #925: Select compact datasets with priority
If a client has both legacy and compact datasets that match the selection criteria, we should prioritise the compact dataset.
Datasets that have ID strictly lower than `legacyPieceStorageIdLimit` are considered legacy.

https://github.com/FilOzone/pdp/blob/f2783319c4870448fa9853ba3916a027ecaddf62/src/interfaces/IPDPVerifier.sol#L14

silent-cipher:
Prefer matching compact data sets during automatic and explicit provider selection, falling back to legacy data sets when no compact match exists. Read migration cutoffs from chain configuration without additional RPC calls.

closes #925


packages/synapse-core/src/devnet/index.ts
```diff
@@ -104,6 +104,8 @@ export function toChain(devnetInfo: VersionedDevnetInfo): FilecoinChain {
      * are unaffected as they use epochs directly.
      */
     genesisTimestamp: 0,
+    // A fresh devnet has no legacy history, so every data set is compact.
+    legacyPieceStorageIdLimit: 1n,
```
silent-cipher, line 108: For review: `legacyPieceStorageIdLimit` is hardcoded per-chain rather than read live from PDPVerifier.legacyPieceStorageIdLimit(). Hardcoding avoids an extra RPC round-trip on every dataset-selection call, but means a future on-chain change requires a manual SDK patch release rather than propagating automatically. Flagging for a second opinion.

packages/synapse-core/src/chains.ts
```diff
@@ -21,6 +21,10 @@ export interface FilecoinChain extends ViemChain {
    * The genesis timestamp of the chain in seconds (Unix timestamp)
    */
   genesisTimestamp: number
+  /**
+   * First data set ID using compact piece storage. Lower IDs use legacy storage.
+   */
+  legacyPieceStorageIdLimit: bigint
   /**
```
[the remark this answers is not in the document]
silent-cipher: Fixed in [234c881](https://github.com/FilOzone/synapse-sdk/pull/960/commits/234c88138c538ba9541243bf741aec6a41d402f0)

packages/synapse-core/src/devnet/index.ts
Kubuxu: that was our agreement with Hugo

Kubuxu (approved):
SGTM

hugomrdias (approved):

